Issue #71, opened 2019-06-10
too-general friend declaration in vector class


grahamlopez:
ASGarD
    asgardasgard
    Issues
    #71

Open
Opened 3 weeks ago by McDaniel, Benjamin T
Close issue
New issue
too-general friend declaration in vector class

for batching, we added the ability to create a matrix view from a vector - essentially, to interpret some section of a vector as a matrix whose modification affects the underlying vector. this is implemented as a matrix view constructor that takes the underlying vector as an argument.

however, this requires that the matrix view to access the vector's ref_count_ - which is private.

I attempted to make just this constructor a friend of the vector class to allow this access. However, I descended swiftly into syntax hell - I couldn't find a friend declaration that compiled AND allowed access in this function to the vector's private member.

So, I made matrices a friend of vectors in general. This is overly broad and we should find the right way to do the specific friending.
Related issues
0

McDaniel, Benjamin T @3bm added help_wanted hygeine labels 3 weeks ago
Graham Lopez
Graham Lopez @bgl · 2 weeks ago
Owner

compile with clang and get a different, more useful message:

    In file included from /home/bgl/code-projects/asgard/asgard-git/src/tensors.cpp:1:
    /home/bgl/code-projects/asgard/asgard-git/src/tensors.hpp:214:21: error: default template argument not permitted on a friend template
      template<mem_type m_ = mem, typename = enable_for_view<m_>, mem_type omem>
                        ^    ~~~
    /home/bgl/code-projects/asgard/asgard-git/src/tensors.hpp:214:31: error: default template argument not permitted on a friend template
      template<mem_type m_ = mem, typename = enable_for_view<m_>, mem_type omem>
                                  ^          ~~~~~~~~~~~~~~~~~~~
    2 errors generated.

and also this was interesting: https://stackoverflow.com/questions/35150772/friend-function-template-with-default-template-argument

grahamlopez:
closed by #98

PR #98, opened 2019-06-12
remove too-friendly matrix<> from vector<>

grahamlopez:
@bmcdanie, @adam-mcdaniel !!!

what do you think? :-D


grahamlopez:
closes #71

PR #103, opened 2019-06-17
Added assertions to prevent level < 2

adam-mcdaniel:
Changed the assertions and if statements in pde base and options for fixing level >= 2


grahamlopez:
okay to test

grahamlopez:
I think there are more occurrences that need updating. The following is not necessarily exhaustive:

```
basis.cpp:367:  assert(num_levels > 0);
connectivity.cpp:13:  assert(level >= 0);
connectivity.cpp:26:  assert(num_levels > 0);
element_table.cpp:19:  assert(num_levels > 0);
element_table.cpp:88:  assert(level_tuple.size() > 0);
element_table.cpp:91:    assert(level >= 0);
pde/pde_base.hpp:98:    assert(level > 0);
pde/pde_base.hpp:353:      assert(d.get_level() > 0);
transformations.hpp:25:  assert(num_levels > 0);
```

adam-mcdaniel:
Changing the following assertions break the tests. Do the tests still expect asgard to support level=1?

```
connectivity.cpp:13:  assert(level >= 0);
connectivity.cpp:26:  assert(num_levels > 0);
element_table.cpp:19:  assert(num_levels > 0);
element_table.cpp:88:  assert(level_tuple.size() > 0);
element_table.cpp:91:    assert(level >= 0);
```

grahamlopez:
They might currently, but this is incorrect and they should be fixed.

grahamlopez:
`get_1d_index()` should keep its current assert because it needs to handle levels 0 and 1, which do exist.

grahamlopez:
`connectivity_tests.cpp:33` will need to be updated (along with updates to the matlab generator)

bmcdanie:
fixed by #176

PR #106, opened 2019-06-19
WIP: Fix/test lambdas bad

adam-mcdaniel:
Get rid of / generalize `relaxed_comparison` as much as possible.
I'll clean up my changes to `testing/tests_general.hpp`, its pretty ugly right now.

I moved the `REQUIRE`s from inside the lambdas to where the lambdas are called, so you should be able to tell where the test actually failed.


grahamlopez:
okay to test

bmcdanie:
addressed by #172

Issue #74, opened 2019-06-10
use 64-bit integers for all indexing functions in tensors


grahamlopez:
This is partially done/WIP at #110 
ASGarD
    asgardasgard
    Issues
    #74

Open
Opened 2 weeks ago by McDaniel, Benjamin T
Close issue
New issue
use 64-bit integers for all indexing functions in tensors

the vectors for batch gemm output/workspace are very large and much be indexed with >32bit numbers.

use 64-bit numbers for the index calculations in build_batches and modify the tensor functions to support those types.
Related issues
0

Related merge requests
1

    Feature/setup and timeloop
    asgard/asgard !69

McDaniel, Benjamin T @3bm added bug label 2 weeks ago
McDaniel, Benjamin T @3bm mentioned in merge request !69 (merged) 2 weeks ago
McDaniel, Benjamin T
McDaniel, Benjamin T @3bm · 2 weeks ago
Developer

I'm going to suggest we use 64 bit integers for the size, rows, and columns of tensors. this, to me, is less error prone than deciding everywhere based on the problem instance whether we need 32 or 64 bits, and the space difference is probably negligible compared to what we need to solve the pde.

additionally, this doesn't prevent us from still using int32 as a type for the tensor elements, so it won't affect much the size of the element table.
McDaniel, Benjamin T
McDaniel, Benjamin T @3bm · 2 weeks ago
Developer

on hold pending solution of #75 (closed)
McDaniel, Benjamin T @3bm added someday label 2 weeks ago

bmcdanie:
suggesting we close this as we've decided not to move in this direction @grahamlopez

grahamlopez:
On Thu, Dec 12, 2019 at 11:20:57AM -0800, Tyler McDaniel wrote:
>suggesting we close this as we've decided not to move in this direction @grahamlopez

Makes sense to me.

PR #110, opened 2019-06-24
WIP: Implement 64-bit indexing for tensors

bmcdanie:
addresses #74 

I did not retroactively change all indexing into tensors to 64-bits everywhere in the code; the use-case for 64-bit indexing should be relatively limited (e.g., workspace indexing in `batch.cpp`). but, I did check in `fast_math` and `tensors` that we don't attempt to invoke blas (with 32-bit api) with invalid values.

I did change the indexing for workspaces in batch.cpp, as well.

### DIFF
--- a/src/batch.cpp
+++ b/src/batch.cpp
@@ -634,8 +634,9 @@ build_batches(PDE<P> const &pde, element_table const &elem_table,
         // * elem_table.size() * terms must be replaced by sum of lower indexed
         // connected items (scan) * terms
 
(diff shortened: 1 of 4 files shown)


grahamlopez:
okay to test

grahamlopez:
@bmcdanie Let me know when you're ready for me to start looking at this.

bmcdanie:
we've moved away from this direction for now - asserting on the 64 bit overflow problem

PR #112, opened 2019-06-25
Removed unused mem_usage component

adam-mcdaniel:
Removed unused component that was used to compound total memory usage


grahamlopez:
nice catch @adam-mcdaniel thanks

grahamlopez:
okay to test

PR #116, opened 2019-07-03
Feature/tasking

bmcdanie:
break problems across both dimensions (elements and connections) to provide fine-grained discretization. 

specify a task size and the problem will be broken into subproblems requiring between 50% and 100% of the provided size (in memory). will enable upcoming gpu/mpi work.


grahamlopez:
okay to test

src/grouping_tests.cpp
```diff
@@ -0,0 +1,805 @@
+
+#include "grouping.hpp"
+#include "tests_general.hpp"
+
+// check for complete, non-overlapping element assignment
+auto const validity_check = [](std::vector<element_group> const &groups,
```
grahamlopez: why is this a lambda?

src/grouping_tests.cpp
```diff
@@ -0,0 +1,805 @@
...
+  {
+    for (int j = 0; j < coverage.ncols(); ++j)
+    {
+      REQUIRE(coverage(i, j) == element_status::assigned);
+    }
+  }
+};
+
+// check that a given task vector occupies between 49% and 101% of the limiit
+auto const size_check = [](std::vector<element_group> const &groups,
```
grahamlopez: why is this a lambda?

src/grouping_tests.cpp
```diff
@@ -0,0 +1,805 @@
...
+      assert(groups.size() % ranks == 0);
+      assert(static_cast<int>(groups.size()) == num_groups);
+      validity_check(groups, table);
+      size_check(groups, *pde, limit_MB, large_problem);
+    }
+  }
+}
+
+// FIXME we should eventually put this in the pde class
+auto const element_segment_size = [](auto const &pde) {
```
grahamlopez: this is repeated from `grouping.cpp:49`

src/grouping_tests.cpp
grahamlopez: and others as well

src/time_advance.cpp
```diff
@@ -4,32 +4,18 @@
...
-
-  for (auto const &ops_list : batches)
-  {
-    assert(static_cast<int>(ops_list.size()) == pde.num_dims);
-    for (batch_operands_set<P> ops : ops_list)
-    {
-      assert(ops.size() == 3);
-    }
-  }
+  assert(dt >= 0);
```
grahamlopez: @dlg0 are we allowing breaking of time-reversal symmetry? :-D
or is this merely for thermodynamic reasons :-(

src/time_advance.hpp
```diff
@@ -1,26 +1,31 @@
...
 #include "batch.hpp"
+#include "grouping.hpp"
 #include "program_options.hpp"
 #include "tensors.hpp"
 
 // this function executes a time step using the current solution
 // vector x. on exit, the next solution vector is stored in fx.
 template<typename P>
-void explicit_time_advance(PDE<P> const &pde,
+void explicit_time_advance(PDE<P> const &pde, element_table const &table,
```
grahamlopez: should there be an `element_table.hpp` include?

Is there any way to avoid adding this dependency?

src/time_advance.cpp
```diff
@@ -41,40 +27,41 @@ void explicit_time_advance(PDE<P> const &pde,
   P const c2  = 1.0 / 2.0;
   P const c3  = 1.0;
 
-  apply_explicit(batches, system);
-  scale_sources(pde, unscaled_sources, system.scaled_source, time);
-  fm::axpy(system.scaled_source, system.batch_output);
-  fm::copy(system.batch_output, system.result_1);
+  apply_explicit(pde, table, groups, host_space, rank_space);
```
grahamlopez: should there be an `element_table.hpp` include?

Is there any way to avoid adding this component dependency?

CMakeLists.txt
```diff
@@ -64,13 +64,14 @@ set (components
   coefficients
   connectivity
   element_table
-  fast_math
+  fast_math 
+  grouping
```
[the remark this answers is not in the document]
grahamlopez: also, "group" is fine. just curious about the change of heart.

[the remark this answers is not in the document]
bmcdanie: I think I'd changed from tasking to grouping before the big "chunk" craze...

src/fast_math.hpp
```diff
@@ -28,7 +28,7 @@ axpy(fk::vector<P, omem> const &x, fk::vector<P, mem> &y, P const alpha = 1.0)
 template<typename P, mem_type mem, mem_type omem>
 fk::vector<P, mem> &copy(fk::vector<P, omem> const &x, fk::vector<P, mem> &y)
 {
-  assert(x.size() == y.size());
+  assert(y.size() >= x.size());
```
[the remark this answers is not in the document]
bmcdanie: yes - blas only takes pointers x and y, their increments between elements, and the size of the copy. so, I don't think it is semantically incorrect to copy a smaller vector into a segment of a larger one.

src/grouping.cpp
```diff
@@ -0,0 +1,361 @@
...
+  // calc in and out vector sizes for each elem
+  // since we will scheme to have most elems require overlapping pieces of
+  // x and y, we will never need 2 addtl xy space per elem
+  double const elem_xy_space_MB = get_MB(elem_size * 1.2);
+  return (elem_reduction_space_MB + elem_intermediate_space_MB +
+          elem_xy_space_MB);
+}
+
+// determine how many groups will be required to solve the problem
+// a task is a subset of all elements whose total workspace requirement
```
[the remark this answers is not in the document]
bmcdanie: yes, fixed

src/grouping.hpp
```diff
@@ -0,0 +1,130 @@
...
+    int64_t num_elems = batch_input.size() + reduction_space.size() +
+                        batch_intermediate.size() + batch_output.size() +
+                        unit_vector_.size();
+    double const bytes     = static_cast<double>(num_elems) * sizeof(P);
+    double const megabytes = bytes * 1e-6;
+    return megabytes;
+  };
+
+private:
+  fk::vector<P> unit_vector_;
```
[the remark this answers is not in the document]
bmcdanie: it isn't changed throughout the simulation - it should always be all ones. wanted to ensure it isn't mistakenly mis-set to something else. could do const, except that I don't have the size computed when it's initially constructed.

CMakeLists.txt
```diff
@@ -64,13 +64,14 @@ set (components
...
   lib_dispatch
   matlab_utilities
   pde
   permutations
   program_options
-  quadrature
+  quadrature
```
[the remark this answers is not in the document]
bmcdanie: fixed

src/grouping.cpp
```diff
@@ -0,0 +1,361 @@
...
+    int const groups_per_rank =
+        static_cast<int>(problem_size_per_rank / num_ranks + 1);
+    return groups_per_rank * num_ranks;
+  }();
+
+  return num_groups;
+}
+
+// divide the problem given the previously computed number of groups
+// this function divides via a greedy, row-major split.
```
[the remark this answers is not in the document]
bmcdanie: yes - comments added

src/main.cpp
```diff
@@ -147,46 +150,40 @@ int main(int argc, char **argv)
...
+            << get_MB(rank_space.reduction_space.size()) << '\n';
   std::cout << "kronmult working space size (MB): "
-            << get_MB(system.batch_intermediate.size()) << '\n';
-  std::cout << "output vector size (MB): " << get_MB(system.batch_output.size())
-            << '\n';
-  auto const &unit_vect = system.get_unit_vector();
+            << get_MB(rank_space.batch_intermediate.size()) << '\n';
+  std::cout << "output vector size (MB): "
+            << get_MB(rank_space.batch_output.size()) << '\n';
+  auto const &unit_vect = rank_space.get_unit_vector();
```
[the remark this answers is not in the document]
bmcdanie: I felt rank space was more descriptive than explicit system - open to other views, though

src/grouping.cpp
```diff
@@ -0,0 +1,361 @@
...
+          .second.second;
+  return std::make_pair(min_col, max_col);
+}
+std::pair<int, int> rows_in_group(element_group const &g)
+{
+  assert(g.size() > 0);
+  return std::make_pair(g.begin()->first, g.rbegin()->first);
+}
+
+auto const element_segment_size = [](auto const &pde) {
```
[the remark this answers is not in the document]
bmcdanie: it does not have to be - it's just a small snippet that is used multiple places here and elsewhere. it may be helpful to have this as a pde class member.

src/grouping_tests.cpp
bmcdanie: yes - that lambda isn't available in this compilation unit. maybe it should be a function?

src/grouping_tests.cpp
bmcdanie: could be a function - just a test I want to apply multiple times. could write it as a function - I think the catch macros make it weird to define them in line here.

src/grouping_tests.cpp
bmcdanie: reused snippet within this scope, and catch makes defining functions inline here weird

src/grouping_tests.cpp
bmcdanie: defined in header

src/time_advance.cpp
bmcdanie: fixed

src/time_advance.hpp
bmcdanie: introduced by building batches here, and unfortunately, can't think of a way to get around it. header added.

src/time_advance.cpp
bmcdanie: dup?

CMakeLists.txt
```diff
@@ -93,14 +94,16 @@ endif ()
 
 target_link_libraries (basis PRIVATE matlab_utilities quadrature tensors)
 
-target_link_libraries (batch PRIVATE lib_dispatch coefficients connectivity element_table pde tensors)
+target_link_libraries (batch PRIVATE lib_dispatch coefficients connectivity grouping element_table pde tensors)
```
grahamlopez: I think we want `INTERFACE chunk` for `time_advance` as well

src/grouping.hpp
```diff
@@ -0,0 +1,130 @@
...
+#include "pde.hpp"
+#include "tensors.hpp"
+
+using element_group = std::map<int, std::pair<int, int>>;
+
+// convenience functions when working with element groups
+int num_elements_in_group(element_group const &g);
+int max_connected_in_group(element_group const &g);
+
+std::pair<int, int> columns_in_group(element_group const &g);
```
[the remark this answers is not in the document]
bmcdanie: yes a small two-member struct may be more appropriate here, or rename. the pair is a start and stop index in both cases.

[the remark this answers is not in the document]
bmcdanie: I'd suggest `range` struct?

CMakeLists.txt
```diff
@@ -93,14 +94,16 @@ endif ()
...
-target_link_libraries (batch PRIVATE lib_dispatch coefficients connectivity element_table pde tensors)
+target_link_libraries (batch PRIVATE lib_dispatch coefficients connectivity chunk element_table pde tensors)
 
 target_link_libraries (coefficients
   PRIVATE pde matlab_utilities quadrature tensors transformations)
 
 target_link_libraries (connectivity
   PRIVATE element_table matlab_utilities permutations tensors)
 
+target_link_libraries(chunk PRIVATE element_table fast_math pde)
```
grahamlopez: I think we need `INTERFACE element_table` on `time_advance` as well

src/element_table.hpp
```diff
@@ -16,7 +16,7 @@
 // - coordinates are composed of a set of dimension-many pairs (level, cell).
 //   (technically, an element's coordinates will also have a degree component,
 //   but we omit this as we assume uniform degree).
-// - in our table, coordinates are stored with all level components grouped,
+// - in our table, coordinates are stored with all level components chunked,
```
grahamlopez: probably revert this change?

src/io.hpp
```diff
@@ -25,7 +25,7 @@ HighFive::DataSet initialize_output_file(fk::vector<P> const &vec)
   HighFive::DataSpace dataspace = HighFive::DataSpace(
       {1, vec_size}, {HighFive::DataSpace::UNLIMITED, vec_size});
 
-  // Use chunking
+  // Use chunk
```
grahamlopez: probably revert this change?

src/main.cpp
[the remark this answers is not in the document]
grahamlopez: Sorry, I meant `unit_vector` -> `unit_vect`

src/matlab_utilities_tests.cpp
```diff
@@ -190,7 +190,7 @@ TEMPLATE_TEST_CASE("find function", "[matlab]", float, double, int)
     fk::vector<int> gold;
     REQUIRE(find(haystack, greater_eq) == gold);
   }
-  SECTION("find a group -- vector")
+  SECTION("find a chunk -- vector")
```
grahamlopez: revert this change?

grahamlopez: (2 in this file)

src/time_advance.cpp
grahamlopez: @dlg0 says no reason for this other than "sanity"

PR #127, opened 2019-07-25
Adding fokkerplanck_1d_4p1a PDE (merge after diffusion PR)

dlg0:


src/pde.hpp
```diff
@@ -33,6 +34,7 @@ enum class PDE_opts
   continuity_2,
   continuity_3,
   continuity_6,
+  fokkerplanck_1d_4p1a,
```
bmcdanie: this name refers to a paper figure, right? pretty obscure, but I guess we want to tie it to the use case

src/pde/pde_fokkerplanck1_4p1a.hpp
```diff
@@ -0,0 +1,201 @@
...
+  initial_condition_dim0(fk::vector<P> const x, P const t = 0)
+  {
+    ignore(t);
+    auto f = analytic_solution_dim0(x, 0);
+    return f;
+  }
+
+  // analytic solution
+
+  // static fk::vector<P>
```
bmcdanie: trim commented code

src/time_advance_tests.cpp
```diff
@@ -1011,12 +1011,12 @@ TEMPLATE_TEST_CASE("time advance - fokkerplanck_1d_4p2",
     }
   }
 
-  SECTION("continuity1, level 2, degree 2, full grid")
+  SECTION("fokkerplanck_1_4p2, level 2, degree 2, full grid")
```
bmcdanie: my fix

dlg0:
> just need to clean some commented code. no other major issues*.
> 
> a couple of thoughts, though:
> 
> 1. do we need to establish a naming convention for these pdes? or something similar? is that even possible to do in a consistent way?

We already do :) It is something like $name$nD_$subname where here the subname refers to the problem number in the draft paper on this PDE. But yes, it could be better.

> 2. minor note - in future, we should only add the files to the generated-inputs directory that we are actually modifying - that way I don't have to click through 75 files that only changed in the git hash. unless we want to keep those consistent with one another, in which case I'll suck it up.

I've done this already. But it was done after I wrote this originally. 

> 3. for this problem, is sparse grid, lev 2 deg 2 sufficient to test the coeff generation? we have more robust tests for the others.

We do. My thought here was that we already test all the pieces this uses. This test is sufficient to see if I coded the PDE up right.

src/pde.hpp
dlg0: Yes, correct. We can update them once the paper stabilizes.

src/pde/pde_fokkerplanck1_4p1a.hpp
dlg0: Done.

dlg0:
Should be ready. Pending CI

Issue #132, opened 2019-08-02
apply_fmwt function tests sporadically failing


tyounkin:
There is an intermittent failure in apply_fmwt() here. To reproduce:

cmake -DCMAKE_CXX_COMPILER=clang++
make -j
for bob in $(seq 1 20) ; do ./basis-tests ; done

In the testing of this routine, random input is generated and then tested. With the test usually passing, but close to the tolerable error, the test occasionally fails.

The testing needs to be updated to use Catch::Matchers::WithinAbs and use matrix size and values to be more accurate.

grahamlopez:
I cannot reproduce this anymore, with either clang or gcc. It looks like David updated the relaxed compare lambda with `Catch::Matchers::WithinAbs` in e9aa6af47 (merged as part of #118).

PR #133, opened 2019-08-02
Fixup/apply fmwt testing

tyounkin:
Testing for apply_fmwt has been updated to compare with full multiplication.  This pull request closes issue #132 The tolerance of the difference between the two matrix multiplication methods is set to epsilon*n*n*max_value/10. Where the square matrix dimension is n and max_value is the maximum value of the coefficient matrix which fmwt is being multiplied against.

### DIFF
--- a/src/basis_tests.cpp
+++ b/src/basis_tests.cpp
@@ -101,15 +101,11 @@ TEMPLATE_TEST_CASE("operator_two_scale function working appropriately",
                    "[transformations]", double)
 {
   auto const relaxed_comparison = [](auto const first, auto const second) {
-    Catch::StringMaker<TestType>::precision = 20;
-
     auto first_it = first.begin();
-    auto tol      = std::numeric_limits<TestType>::epsilon() * 1e3;
-    std::for_each(second.begin(), second.end(),
-                  [&first_it, tol](auto &second_elem) {
-                    REQUIRE_THAT(*first_it++,
-                                 Catch::Matchers::WithinAbs(second_elem, tol));
-                  });
+    std::for_each(second.begin(), second.end(), [&first_it](auto &second_elem) {
+      auto difference = std::abs(*first_it++ - second_elem);
+      REQUIRE(difference <= std::numeric_limits<TestType>::epsilon() * 1e4);
+    });
   };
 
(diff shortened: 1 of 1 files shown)


grahamlopez:
okay to test

src/basis_tests.cpp
```diff
@@ -101,15 +101,11 @@ TEMPLATE_TEST_CASE("operator_two_scale function working appropriately",
                    "[transformations]", double)
 {
   auto const relaxed_comparison = [](auto const first, auto const second) {
```
grahamlopez: did you purposely revert this implementation of `relaxed_comparison`, or did it just fall through the rebase?

src/basis_tests.cpp
```diff
@@ -196,13 +192,17 @@ TEMPLATE_TEST_CASE("operator_two_scale function working appropriately",
...
-    });
+  auto const relaxed_comparison = [](auto const first, auto const second,
+                                     auto const n, auto const max) {
+    auto first_it                           = first.begin();
+    Catch::StringMaker<TestType>::precision = 15;
+    std::for_each(
+        second.begin(), second.end(), [&first_it, &n, &max](auto &second_elem) {
+          auto const f1 = *first_it++;
+          auto tol =
+              std::numeric_limits<TestType>::epsilon() * n * n / 10.0 * max;
```
grahamlopez: we're going to need a comment about why tolerance is chosen here

src/basis_tests.cpp
grahamlopez: after discussion, we need to revert this back.

src/basis_tests.cpp
grahamlopez: so if the largest matrix we are testing is either 80x80 or 128x128, with a max value of 200, then this gives us an `epsilon() * 20 * 128` ~~ `epsilon() * 3e3`. so not terribly worse than other places in the code.

grahamlopez:
I cannot reproduce this anymore, with either clang or gcc. It looks like David updated the relaxed compare lambda with Catch::Matchers::WithinAbs in e9aa6af (merged as part of #118).

I am closing until it shows up again.

PR #134, opened 2019-08-02
Fix/perf predict level 1

adam-mcdaniel:
This PR:
1. Makes the minimum level to test default to 2 in perf_predict.
2. fixes an error where spreadsheets would be improperly sized when the lowest level to profile was greater than 1.


grahamlopez:
closing this for now to clean up a little bit. we've gotten a lot of use out of perfpredict and would be interested to discuss some updated methods as we move forward.

PR #136, opened 2019-08-06
Feature/parallel batch (depends on feature/gpus)

bmcdanie:
parallel batch building with optional openmp. time advance 2x+ speedup for large problems.

### DIFF
--- a/src/batch.cpp
+++ b/src/batch.cpp
@@ -1,5 +1,5 @@
 #include "batch.hpp"
-
+#include "build_info.hpp"
 #ifdef ASGARD_USE_OPENMP
 #include <omp.h>
 #endif
--- a/src/batch.cpp
+++ b/src/batch.cpp
@@ -614,9 +614,9 @@ build_batches(PDE<P> const &pde, element_table const &elem_table,
     return builder;
   }();
 // loop over elements
-// FIXME eventually want to do this in parallel
 #pragma omp parallel for
-  for (int chunk_num = 0; chunk_num < chunk.size(); ++chunk_num)
+  for (int chunk_num = 0; chunk_num < static_cast<int>(chunk.size());
+       ++chunk_num)
   {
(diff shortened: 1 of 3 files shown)


grahamlopez:
okay to test

src/batch.cpp
```diff
@@ -1,4 +1,9 @@
 #include "batch.hpp"
+
+#ifdef ASGARD_USE_OPENMP
```
grahamlopez: follow the pattern for `ASGARD_IO_HIGHFIVE` here to get the conditional build info

src/batch.cpp
```diff
@@ -577,10 +582,28 @@ build_batches(PDE<P> const &pde, element_table const &elem_table,
   std::vector<batch_operands_set<P>> batches =
       allocate_batches<P>(pde, elements_in_chunk);
 
-  // loop over elements
-  // FIXME eventually want to do this in parallel
-  for (const auto &[i, connected] : chunk)
+  std::vector<int> const index_to_key = [&chunk]() {
```
grahamlopez: the need for this indirection is not yet obvious to me.

bmcdanie: I really want to loop over `[i, connected]` as I did before, but I could not find a way to make the structured binding work with openmp. `i` may not start at zero, and although our current chunking implementation assures that the range of `i` is contiguous, I didn't want to rely on that.

so, instead, I am forming a range from 0 to `chunk.size()` to loop over, and then retrieving the associated `[i, connected]` values inside the loop.

src/batch.cpp
bmcdanie: I patterned this on how `ASGARD_IO_HIGHFIVE` is set - or intended to. what changes would you like here?

bmcdanie: changed to autogenerated header definition per Slack convo

PR #139, opened 2019-08-13
Added gesv and test

stonecoldhughes:
The test for double precision float always works, but the test for single precision float version is off by tiny epsilon deviations - is it floating point representation limitations or is there an error I've missed?

### DIFF
--- a/src/lib_dispatch.cpp
+++ b/src/lib_dispatch.cpp
@@ -271,6 +271,35 @@ void gemm(char const *transa, char const *transb, int *m, int *n, int *k,
 }
 
 template<typename P>
+void gesv(int *n, int *nrhs, P *A, int *lda, int *ipiv, P *B, int *ldb,
+          int *info)
+{
+  assert(n);
+  assert(nrhs);
+  assert(A);
+  assert(lda);
+  assert(ipiv);
+  assert(B);
+  assert(ldb);
+  assert(info);
+
+  if constexpr (std::is_same<P, double>::value)
+  {
+    dgesv_(n, nrhs, A, lda, ipiv, B, ldb, info);
+  }
+
+  else if constexpr (std::is_same<P, float>::value)
+  {
+    sgesv_(n, nrhs, A, lda, ipiv, B, ldb, info);
+  }
+
+  else
+  {
+    std::cerr << "gesv not implemented for non-floating types" << std::endl;
+  }
+}
+
+template<typename P>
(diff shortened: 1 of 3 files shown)


grahamlopez:
okay to test

src/lib_dispatch.cpp
```diff
@@ -270,6 +270,35 @@ void gemm(char const *transa, char const *transb, int *m, int *n, int *k,
   }
 }
 
+template<typename P>
+void gesv(int *n, int *nrhs, P *A, int *lda, int *ipiv, P *B, int *ldb,
+          int *info)
+{
+  assert(n);
```
grahamlopez: can you make any stronger assertions for all of these?

src/lib_dispatch.cpp
```diff
@@ -270,6 +270,35 @@ void gemm(char const *transa, char const *transb, int *m, int *n, int *k,
   }
 }
 
+template<typename P>
+void gesv(int *n, int *nrhs, P *A, int *lda, int *ipiv, P *B, int *ldb,
```
grahamlopez: can you remind me why we want to expose `gesv()` in lib_dispatch? and then if yes, why not also in fast_math?
Put these kinds of things in the PR description

src/lib_dispatch.hpp
```diff
@@ -108,6 +118,10 @@ void gemm(char const *transa, char const *transb, int *m, int *n, int *k,
           P *alpha, P *A, int *lda, P *B, int *ldb, P *beta, P *C, int *ldc,
           resource const res = resource::host);
 
+template<typename P>
+void gesv(int *n, int *nrhs, P *A, int *lda, int *ipiv, P *B, int *ldb,
```
grahamlopez: need extern template decls as well

src/lib_dispatch_tests.cpp
```diff
@@ -9,8 +9,7 @@
 //
 TEMPLATE_TEST_CASE("matrix-matrix multiply (lib_dispatch::gemm)",
                    "[lib_dispatch]", float, double, int)
-{
-  // clang-format off
+{ // clang-format off
```
grahamlopez: restore this line to its original position

grahamlopez:
@stonecoldhughes let's get this dealt with or closed, whichever you think is the right path.

grahamlopez:
it looks like this might be obviated by #155

bmcdanie:
gesv w/tests implemented in #155

PR #149, opened 2019-08-26
Initial commit of mpi communication module.

stonecoldhughes:


bmcdanie:
I still need to wrap up some work before I get to the code that would consume this. Let me give some high level comments while I finish that.

## Naming

- MPI is an implementation detail, not part of the interface - let's not use that for naming. `distributed` or `communication` etc. is more descriptive of function.
- `endpoint`, `packet`, etc. are networking/distrib. comp. terms that have a meaning. I think (IIRC) your `endpoints` describe the arguments for a unidirectional, point to point message - `message` might be more appropriate? not sure, but we try to avoid names that are already oversubscribed.

## Usability

- Simplest api possible - don't pass objects by rvalue reference when it doesn't provide a performance benefit, and when we do need to do that, make it clear in the function name that we are taking ownership.

## Readability/style/conventions

- Remove spurious return statements in constructors/void return functions. I can see what an empty function body does / when a void function ends; the return just adds a line to the file. I could be persuaded in the case of a very long void function; but then, I'd suspect a function that large is overly bloated anyway.
- East `const`
- Composition is easier to read and understand than inheritance, but functional input/output is easier to read and understand than either.  Avoid superfluous classes, e.g. those that simply wrap a container. 
- Where possible, the code should express what it's doing; especially difficult/complex segments may require explanation in comments. Especially looking here at the loop in the interval translation code - that is difficult to decipher without explanation.

This will be rolled into the distribution component, so don't worry about the component naming issue. I should (hopefully) wrap up my portion of that today, and then we can make a plan for integrating this.

grahamlopez:
I've added most of Tyler's comments to https://github.com/project-asgard/asgard/wiki/coding-standards

PR #154, opened 2019-09-10
WIP Features/direct implicit

elwasif:
Use direct implicit solve for the time loop


src/batch.cpp
```diff
@@ -784,7 +784,7 @@ void build_implicit_system(PDE<P> const &pde, element_table const &elem_table,
         //A_view.print("A_view 1");
         A_view = A_view + k_tmp;
         A_view.print("A_view 2");
```
elwasif: @grahamlopez If I remove this print line, the compiler seems to optimize away the assignment to A_view and the results are different

grahamlopez:
okay to test

PR #155, opened 2019-09-10
Implicit direct solve

elwasif:
Direct implicit solve - from develop scratch


grahamlopez:
okay to test

grahamlopez:
I see this from the CI:
```
src/batch.cpp:770:55: error: no matching function for call to ‘fk::matrix<float, (mem_type)1, (resource)0>::matrix(const fk::matrix<float, (mem_type)0, (resource)1>&, int&, int, int&, int)’
```
I believe this is because you are trying to construct a host view from a device owner

elwasif:
> I see this from the CI:
> 
> ```
> src/batch.cpp:770:55: error: no matching function for call to ‘fk::matrix<float, (mem_type)1, (resource)0>::matrix(const fk::matrix<float, (mem_type)0, (resource)1>&, int&, int, int&, int)’
> ```
> 
> I believe this is because you are trying to construct a host view from a device owner

I see ` fk::matrix<P, mem_type::owner, resource::device> coefficients_;` So nowwe can only access the coefficients on the device?

grahamlopez:
> So nowwe can only access the coefficients on the device?

This is where the memory lives, so you cannot make a view from the host to the device. You either need a device view, or a copy of the data on the host.

elwasif:
So now I have this `working inside the loop
``` C++
          fk::matrix<P, mem_type::owner, resource::host> tmp;
          auto & t = pde.get_coefficients(k, d);
          tmp.clear_and_resize(t.nrows(), t.ncols());
          tmp.transfer_from(t);
          fk::matrix<P, mem_type::view> op_view = fk::matrix<P, mem_type::view>(
              tmp, operator_row(d),
```
It works but is inefficient, will probably cache in a ```std::map<>``` but I just wanted to check if this is the right way to do this

grahamlopez:
I think you could simplify a little bit with something like
```
fk::matrix<P, mem_type::owner, resource::host> tmp(pde.get_coefficients(k, d).clone_onto_host());
fk::matrix<P, mem_type::view> op_view = fk::matrix<P, mem_type::view>(tmp, operator_row(d),
```
but you are still doing a memory allocation and transfer in your loop, so this likely won't fix all of your performance issues.

src/batch.cpp
```diff
@@ -693,6 +705,113 @@ build_batches(PDE<P> const &pde, element_table const &elem_table,
   return batches;
 }
 
+// function to allocate and build implicit system.
+// given a problem instance (pde/elem table)
+template<typename P>
+void build_implicit_system(PDE<P> const &pde, element_table const &elem_table,
```
dlg0: While we build the system matrix and invert it to achieve an implicit time advance scheme, the matrix we invert (`A` here) is the same for implicit or explicit time advance (we just don't build it in the explicit machinery - although it's not a bad idea to have that capability too, so we could test the batch matvec against a vanilla matvec where we actually have A.

Anyway, my only point here is the name of this function could be changed to say `build_system_matrix` or some such since it isn't technically related to implicit time advance.

src/batch.cpp
```diff
@@ -693,6 +705,113 @@ build_batches(PDE<P> const &pde, element_table const &elem_table,
...
+  // FIXME eventually want to do this in parallel
+  for (const auto &[i, connected] : chunk)
+  {
+    // first, get linearized indices for this element
+    //
+    // calculate from the level/cell indices for each
+    // dimension
+    fk::vector<int> const coords = elem_table.get_coords(i);
+    assert(coords.size() == pde.num_dims * 2);
+    fk::vector<int> elem_indices = linearize(coords);
```
dlg0: There are multiple "linearizations" of coordinates in our code. There is one that occurs within the 1D coordinates, and then a second which maps the multi-D coordinates to a linear space. The variable naming here does not tell me which one this is. In the matlab I think I would do something like `coords_1D` for a 1D coordinate, and `coords_nD` for a multi-D coordinate.

src/batch.cpp
```diff
@@ -693,6 +705,113 @@ build_batches(PDE<P> const &pde, element_table const &elem_table,
...
+        kron_vals.push_back(kron0);
+        for (int d = 0; d < pde.num_dims; d++)
+        {
+          // printf("pde.get_coefficients(%d, %d) :\n", k, d);
+          // pde.get_coefficients(k, d).print("COEFF");
+          // printf("View: (%d, %d) -> (%d, %d) \n", operator_row(d),
+          //     operator_col(d),
+          //       operator_row(d) + degree - 1,
+          //       operator_col(d) + degree - 1);
+          // printf("=====================================\n");
```
dlg0: Let's remove unused / commented code from PRs.

src/batch.cpp
```diff
@@ -693,6 +705,113 @@ build_batches(PDE<P> const &pde, element_table const &elem_table,
...
+          fk::matrix<P, mem_type::view> op_view = fk::matrix<P, mem_type::view>(
+              coef_cache[key_type(k, d)], operator_row(d),
+              operator_row(d) + degree - 1, operator_col(d),
+              operator_col(d) + degree - 1);
+          // op_view.print("OP_VIEW");
+          // printf("$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$\n");
```
dlg0: As above (and some more below)

src/batch.cpp
```diff
@@ -693,6 +705,113 @@ build_batches(PDE<P> const &pde, element_table const &elem_table,
...
+
+          fk::matrix<P> k_new = kron_vals[d].kron(op_view);
```
dlg0: I don't understand this line. I think that `op_view` is the `deg`*`deg` matrix for each dimension, and here were are building the list of `num_dims` many of these to kron together later? Or is the `.kron()` actually performing the kron expansion right here? If so, I'm unsure why `kron` is a member of `kron_vals` ... Ahhhh, now I get it. `.kron` is a member function of the `fk::matrix` class, and calling it this way krons itself with the argument. I'm not sure if this naming is confusing or not. If doing it this way, why have a vector of mats at all? Why not just do something like ...

```
for (int d = 0; d < pde.num_dims; d++)
{
    ...
    kron_vals = kron_vals.kron(op_view);
}
```
since I'd think that the second to last element of `kron_vals` might get pretty big?

dlg0:
@elwasif Is this ready for re-review?

elwasif:
Yes
________________________________

elwasif:
retest this please

dlg0:
okay to test

dlg0:
Fusiont5 needs the decrypt pw. I’ll do it when I’m done in this meeting.

src/batch.cpp
```diff
@@ -693,6 +705,113 @@ build_batches(PDE<P> const &pde, element_table const &elem_table,
...
+        // printf("Writing into A(%d, %d) of size %d\n\n", global_row,
+        // global_col, k_tmp.nrows());
+
+        fk::matrix<P, mem_type::view> A_view(
+            A, global_row, global_row + k_tmp.nrows() - 1, global_col,
+            global_col + k_tmp.ncols() - 1);
+
+        // A_view.print("A_view 1");
+        A_view = A_view + k_tmp;
+        A_view.print("A_view 2");
```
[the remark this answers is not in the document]
grahamlopez: grr, that's a bummer. let me see if I can find a less troublesome way. My first thought was to try something like:

```
auto const dont_ignore = [](auto ignored) { (void)ignored; };
A_view = A_view + k_tmp; 
dont_ignore(A_view);
```

I think you are right: the compiler is determining that there are no side effects.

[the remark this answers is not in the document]
grahamlopez: So I was just trying out my suggestion, but first I tried to reproduce the failure with your original code (but without the `.print()`). However, running the tests all pass. How were you seeing the incorrect results? (If no tests failed as a result, we might need some better tests somewhere.)

grahamlopez:
@bmcdanie are you happy with the new tests for batch and time_advance?

src/lib_dispatch.hpp
```diff
@@ -76,6 +76,16 @@ extern "C"
 
   void sgetri_(int *n, float *A, int *lda, int *ipiv, float *work, int *lwork,
                int *info);
+
+  void dgesv_(int *n, int *nrhs, double *A, int *lda, int *ipiv, double *b,
```
grahamlopez: why no tests for these?

src/time_advance.cpp
```diff
@@ -119,6 +119,61 @@ apply_explicit(PDE<P> const &pde, element_table const &elem_table,
   }
 }
 
+// this function executes an explicit time step using the current solution
```
grahamlopez: this function seems out of place. let's ove it and leave another for your function.

src/time_advance.cpp
```diff
@@ -119,6 +119,61 @@ apply_explicit(PDE<P> const &pde, element_table const &elem_table,
...
+      for (int j = 0; j < A.ncols(); ++j)
+      {
+        A(i, j) *= -dt;
+      }
+      A(i, i) += 1.0;
+    }
+    // first_time = false;
+  }
+  // A.print("A");
+  fm::copy(host_space.x, host_space.x_orig);
```
grahamlopez: it looks like these lines can be moved before the first conditional, and then the two if()'s can be combined?

src/batch.hpp
```diff
@@ -132,6 +132,16 @@ std::vector<batch_operands_set<P>>
 build_batches(PDE<P> const &pde, element_table const &elem_table,
               rank_workspace<P> const &workspace, element_chunk const &chunk);
 
+template<typename P>
+void build_system_matrix(PDE<P> const &pde, element_table const &elem_table,
+                         element_chunk const &chunk, fk::matrix<P> &A);
+
+template<typename P>
+fk::matrix<P>
+build_system_matrix(PDE<P> const &pde, element_table const &elem_table,
```
grahamlopez: I don't see the definition for this overload; where is it?

elwasif: Leftover from earlier versions, will remove.

src/time_advance.cpp
elwasif: Not sure I understand

src/time_advance.cpp
elwasif: Yes. Will do

src/batch.cpp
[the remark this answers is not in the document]
elwasif: I saw this (before adding the tests)  when I manually compared the system  matrix with the matlab version, and also in the error reported in the time loop.  Now when I try it again, I don't see this anymore. Weird (or more likely PEBCAK)

src/lib_dispatch.hpp
elwasif: added test case in d057d9e7

src/time_advance.cpp
elwasif: Done in d057d9e7225666f6c6d7bffeed854ef50a90389a

src/batch.hpp
elwasif: Updated in a0675a501c1ef41cf5b09b85749bbbd81133676e

elwasif:
Retest this please

dlg0:
okay to test

src/batch.cpp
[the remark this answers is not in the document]
grahamlopez: Weird indeed. Compilers ::sigh::.

Can you go ahead and revert back to the simple statement?

src/time_advance.cpp
grahamlopez: sorry, I meant the comment is out of place. it is for the explicit.

src/time_advance.cpp
grahamlopez: nice. much better.

src/batch.cpp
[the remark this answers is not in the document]
elwasif: Updated in e59f8c3e

src/time_advance.cpp
elwasif: Fixed in 9016ae66ddc67329b46666c3492f5136351e8ff9

elwasif:
Retest this please

dlg0:
@elwasif You should now be added to the whitelist for people whose commits automatically get pushed through the CI. Also, I'm told that once a single PR has had a single "okay to test", it will automatically go to CI for subsequent commits :)

src/batch.cpp
```diff
@@ -693,6 +699,97 @@ build_batches(PDE<P> const &pde, element_table const &elem_table,
...
+    fk::vector<int> const coords = elem_table.get_coords(i);
+    assert(coords.size() == pde.num_dims * 2);
+    fk::vector<int> elem_indices = linearize(coords);
+
+    int const global_row = i * elem_size;
+
+    // calculate the row portion of the
+    // operator position used for this
+    // element's gemm calls
+    fk::vector<int> operator_row = get_operator_row(pde, degree, elem_indices);
```
grahamlopez: is this unused?

src/batch.cpp
grahamlopez: also, looks like `elem_indices` can be const

src/batch.cpp
elwasif: Used in lines 620 and 635

grahamlopez: ah, thanks. This was a tooling problem on my side.

grahamlopez:
Okay, once @dlg0's and @bmcdanie's comments are addressed, I'm good with this one.

src/batch.cpp
elwasif: Done in 4ed483a1

elwasif: @dlg0 So for building the system matrix, this is the nD coords, right ?

dlg0: I think that here `coords` is indeed a multi-dimensional coordinate, so yeah, `coords_nD` would work.

elwasif: Applied in 272358a2

src/batch.cpp
```diff
@@ -544,6 +544,31 @@ static fk::vector<int> linearize(fk::vector<int> const &coords)
   return elem_indices;
 }
 
+template<typename P>
+inline fk::vector<int> get_operator_row(const PDE<P> &pde, const int degree,
```
bmcdanie: for consistency with rest of code, please place `const` on the right side of the type. and maybe @dlg0 can order us some wristbands: http://slashslash.info/eastconst/

src/batch.cpp
```diff
@@ -693,6 +699,97 @@ build_batches(PDE<P> const &pde, element_table const &elem_table,
...
+        for (int d = 0; d < pde.num_dims; d++)
+        {
+          fk::matrix<P, mem_type::view> op_view = fk::matrix<P, mem_type::view>(
+              coef_cache[key_type(k, d)], operator_row(d),
+              operator_row(d) + degree - 1, operator_col(d),
+              operator_col(d) + degree - 1);
+          fk::matrix<P> k_new = kron_vals[d].kron(op_view);
+          kron_vals.push_back(k_new);
+
+          // k_new.print("K_NEW");
```
bmcdanie: please remove commented code

src/batch.cpp
```diff
@@ -693,6 +699,97 @@ build_batches(PDE<P> const &pde, element_table const &elem_table,
...
+                         element_chunk const &chunk, fk::matrix<P> &A)
+{
+  // assume uniform degree for now
+  int const degree    = pde.get_dimensions()[0].get_degree();
+  int const elem_size = static_cast<int>(std::pow(degree, pde.num_dims));
+  int const A_size    = elem_size * elem_table.size();
+  using key_type      = std::pair<int, int>;
+  using val_type      = fk::matrix<P, mem_type::owner, resource::host>;
+  std::map<key_type, val_type> coef_cache;
+
```
bmcdanie: can we add a comment here that the intent with this segment is to copy the coeff matrices onto the host?

src/batch.cpp
bmcdanie: we could definitely add such a function in `fast_math` at some point

src/fast_math.hpp
```diff
@@ -110,4 +110,73 @@ gemm(fk::matrix<P, amem, resrc> const &A, fk::matrix<P, bmem, resrc> const &B,
...
+
+  int lda = A.stride();
+  int ldb = B.size();
+
+  int info;
+  lib_dispatch::gesv(&rows_A, &cols_B, A.data(), &lda, ipiv.data(), B.data(),
+                     &ldb, &info);
+  if (info > 0)
+  {
+    printf("The diagonal element of the triangular factor of A,\n");
```
bmcdanie: can we use `std::cout` here?

elwasif:
@bmcdanie Addressed comments in 76e16c66

src/batch.cpp
```diff
@@ -545,8 +545,8 @@ static fk::vector<int> linearize(fk::vector<int> const &coords)
 }
 
 template<typename P>
-inline fk::vector<int> get_operator_row(const PDE<P> &pde, const int degree,
-                                        const fk::vector<int> &elem_indices)
+inline fk::vector<int> get_operator_row(PDE<P> const &pde, int const degree,
```
bmcdanie: thanks!

src/batch.cpp
dlg0: OK, optimization for later.

grahamlopez:
Thanks for the work, @elwasif! Merging.

PR #138, opened 2019-08-12
Feature/mpi

bmcdanie:
initial work to distribute work across nodes

also closes #146 .

PR #174, opened 2019-10-10
Refactor/standardize build (depends on #138 MPI)

### LINKED ISSUE #167: standardize build options
currently, using cuda is set using an explicit compile definition in the cmake, whereas openmp and the io option are set automatically by cmake. cuda option should be brought in line with the others.

bmcdanie:
closes #167.


bmcdanie:
rebase was more time-intensive than re-implementing using this as pattern

PR #175, opened 2019-10-10
Feature/parse stop (depends on #138 MPI)

### LINKED ISSUE #146: hard stop when failing to parse
if I enter

`./asgard continuity_6 -l 2 -d 2`

to run the code (I'm missing `-p` before the pde), an error will be printed, but we will continue with the default pde (continuity 2). we want to exit instead when parsing fails.

bmcdanie:
closes #146 .


bmcdanie:
fixed in #138; redundant

PR #178, opened 2019-10-10
WIP: Refactor/test (depends on #138 MPI)

### LINKED ISSUE #144: Remove duplication of code from _tests.cpp
In `time_advance_tests.cpp` we use a lot of duplicate setup code just to test each of the PDEs. This needs to be cleaned up such that we use the setup code from the main program itself, rather than this duplicate - but possibly different from the real code - version. 

It is also a pain to maintain if anything in the setup code changes, since every PDE test code has to be updated.

bmcdanie:
continues work started in #138; closes #144


bmcdanie:
reopened in another branch

Issue #184, opened 2019-10-29
bug in rank workspace allocation and chunk transfer


bmcdanie:
in `chunk.cpp` the function that determines batch input allocation size: https://github.com/project-asgard/asgard/blob/092b609d2f852ed5a37eab5ff491971f0d492a33/src/chunk.cpp#L14 incorrectly assumes that the largest `x` size per chunk is the maximum connected element per row. actually, the maximum x space could fall across rows, e.g. if one row within a chunk requires half of a row to half of the next, the entire x vector is required for the chunk.

additionally, in `chunk.cpp` the `x_view` vector: https://github.com/project-asgard/asgard/blob/092b609d2f852ed5a37eab5ff491971f0d492a33/src/chunk.cpp#L263 is implicitly assumed to match `rank_space.batch_input` in size; this is only true when the problem size is small and chunks all contain multiple rows. need to either 1) allow the destination for transfer function to be equal to or greater than source (i.e. change assert) or 2) create a view of the rank_space vector that matches the smaller `x_view` vector before copy.

found these bugs during performance testing with large 6d problems - need to be addressed.

bmcdanie:
discussed with @grahamlopez today about how this interacts with #166. The simplest solution for fixing the issues of chunks within a subgrid addressing discontinuous portions of the input vector is to fix #166 and hold the input (and output) vector in rank space for all chunks within a subgrid. More complex solutions (including z-style element assignment of elements to chunks and providing a mapping function for each chunk to (potentially discontinuous) portions of the input vector) exist if necessary.

bmcdanie:
this was fixed with #186 .

PR #186, opened 2019-10-31
Fix/chunking errors (depends on #183)

### LINKED ISSUE #165: cap RAM usage in chunk tests
chunk tests blow up mortal computers, RAM usage needs to be humbled

bmcdanie:
Fix the logic error that caused `rank_space.batch_input` to be mis-sized, and extend tests to catch this problem.
Also closes #165 .


grahamlopez:
there are some incorrect bits here. I discussed with @bmcdanie, and we're going to merge this as it will make the true solution (#184) a little more straightforward. so this is a temporary stop-gap that is partially correct, but #184 still needs to happen and should completely close this bug.

grahamlopez:
test this please

PR #187, opened 2019-10-31
Feature/default cfl (depends on #186)

### LINKED ISSUE #152: Changing CFL causes tests to fail
The default Courant–Friedrichs–Lewy condition (cfl) has a value of 0.1 set in program_options.hpp. Running problems from the command line with this default value causes enormous RMSE error. Reducing the cfl to 0.01 causes better RMSE for command-line runs but causes several tests in which the cfl is not explictly set to fail.

bmcdanie:
closes #152


grahamlopez:
@bmcdanie why did you decide to change the default, but then test on a different custom value? Why not test on the default as we were before?

bmcdanie:
@grahamlopez just to minimize the change to our existing generated inputs / generation scripts. I can update those and push new data if we'd prefer that.

grahamlopez:
yeah, I think let's do for consistency's sake. it's good to test the most common case

bmcdanie:
done. also will start a PR to fix gold generation in MATLAB once @dlg0 gives me repo access.

sorry about the delay, hit with interviews + paper rebuttals simultaneously

PR #189, opened 2019-10-31
Feature/batch perf (depends on #188)

bmcdanie:

PR #191, opened 2019-11-07
Feature/batch raw assign (depends on #189)

bmcdanie:
add capability to assign pointers directly into batch (skip use of view).

we lose some safety checks but the performance gains are huge. before/after callgraphs will be posted for comparison.

### DIFF
--- a/src/batch.cpp
+++ b/src/batch.cpp
@@ -697,10 +697,14 @@ build_batches(PDE<P> const &pde, element_table const &elem_table,
   std::vector<batch_operands_set<P>> batches =
       allocate_batches<P>(pde, elements_in_chunk);
 
-  // i: row we are addressing in element grid
-  // connected: start/stop for this row
-  for (auto const &[i, connected] : chunk)
+  // can't use structured binding; variables passed into outlined omp func
+  for (auto const &row_info : chunk)
   {
+    // i: row we are addressing in element grid
+    auto const i = row_info.first;
+    // connected: start/stop for this row
+    auto const connected = row_info.second;
+
     // first, get linearized indices for this element
     //
     // calculate from the level/cell indices for each
--- a/src/batch.cpp
+++ b/src/batch.cpp
@@ -697,28 +697,10 @@ build_batches(PDE<P> const &pde, element_table const &elem_table,
   std::vector<batch_operands_set<P>> batches =
       allocate_batches<P>(pde, elements_in_chunk);
 
-  std::vector<int> const index_to_key = [&chunk]() {
-    std::vector<int> builder;
-    for (auto const &[i, connected] : chunk)
-    {
-      builder.push_back(i);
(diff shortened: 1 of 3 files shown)


bmcdanie:
note to self: failure caused by structured binding outlined by openmp function

src/batch.hpp
```diff
@@ -128,6 +129,11 @@ void kronmult_to_batch_sets(
     PDE<P> const &pde);
 
 template<typename P>
```
grahamlopez: add a quick description of this overload

src/batch.cpp
```diff
@@ -422,6 +432,16 @@ static void kron_base(fk::matrix<P, mem_type::view, resource::device> const &A,
   batches[2].assign_entry(y_view, batch_offset);
 }
 
+// same function, high performance version
```
grahamlopez: "// same function, unsafe version"

src/batch.cpp
```diff
@@ -538,6 +559,75 @@ void kronmult_to_batch_sets(
   batches[pde.num_dims - 1][2].assign_entry(y_view, batch_offset);
 }
 
+template<typename P>
+void kronmult_to_batch_sets(std::vector<P *> const &A, P *const x, P *const y,
```
grahamlopez: give us a summary of what makes this version unsafe

grahamlopez: also, let's rename to `unsafe_kronmult_to_batch_sets()`

src/batch_tests.cpp
```diff
@@ -143,6 +143,37 @@ TEMPLATE_TEST_CASE_SIG("batch", "[batch]",
...
+
+      test.assign_raw(second_v.data(), 1);
+      ptr_list = test.get_list();
+      REQUIRE(ptr_list[0] == first_v.data());
+      REQUIRE(ptr_list[1] == second_v.data());
+      REQUIRE(ptr_list[2] == third_v.data());
+      REQUIRE(ptr_list[0] == test(0));
+      REQUIRE(ptr_list[1] == test(1));
+      REQUIRE(ptr_list[2] == test(2));
+    }
```
grahamlopez: no tests for the `unsafe_kronmult_to_batchsets()`?

src/batch.hpp
bmcdanie: done

src/batch.cpp
bmcdanie: done

src/batch.cpp
bmcdanie: done

src/batch_tests.cpp
bmcdanie: added. replicated safe section and shimmed calls into the unsafe version - the test refactor #193 will clean these up - I'll rebase that once this is merged.

PR #202, opened 2019-12-10
use 03 and g in all builds

### LINKED ISSUE #131: compiler warning audit
It would be worthwhile for someone to understand the yet unused warnings that are available, and to reason out which ones we want and which we don't.

bmcdanie:
closes #131 . tldr; we are using the right warning flags, just setting the default build to debug here. more info in the issue.


bmcdanie:
note to self: on thursday, check `fno-exceptions` performance to determine whether setting separate compile flags for the io component is worth doing

bmcdanie:
test this please

PR #203, opened 2019-12-10
clarify time advance naming

### LINKED ISSUE #141: Comments on apply_explicit
https://github.com/project-asgard/asgard/blob/7bedea70a8d57796a2058e65ac5ad1c0e6ef8082/src/time_advance.cpp#L88

In trying to interpret what's happening here I have the following comments ...

1. I'd suggest that `apply_explicit` should be named `apply_A`. Nominally it is not associated with either explicit or implicit time advance, but rather just applying the system matrix to the present solution vector. 
2. The naming and comments surrounding `copy_chunk_inputs` and `copy_chunck_ouputs` could be more descriptive, i.e., at present, it does not tell me what those inputs and outputs are. i.e., are they pieces of the input and output vectors, or are they the coefficient matrices, or what? 
3. Could `copy_chunk_inputs` and `copy_chunk_ouputs` be a single function with the order of the arguments reversed for the second call? Or at least have the source/destination argument order be the same in both calls?

bmcdanie:
clarify: `apply_explicit` -> `apply_system`.

didn't touch the naming of chunk data management functions; there is a PR in the pipeline #199 that addresses those.

closes #141 if you're happy with the name @dlg0 .

### DIFF
--- a/src/time_advance.cpp
+++ b/src/time_advance.cpp
@@ -113,10 +113,9 @@ scale_sources(PDE<P> const &pde,
 template<typename P>
 static void
 apply_A(PDE<P> const &pde, element_table const &elem_table,
-               element_subgrid const &grid,
-               std::vector<element_chunk> const &chunks,
-               host_workspace<P> &host_space, rank_workspace<P> &rank_space,
-               std::vector<batch_operands_set<P>> &batches)
+        element_subgrid const &grid, std::vector<element_chunk> const &chunks,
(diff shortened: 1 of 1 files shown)


bmcdanie:
@dlg0 I'm not opposed to sticking with apply_A - I agree A is canonically "that matrix" in the context of Ax = B, but I guess I'm trying to give some notion of what A is/what it represents. if you prefer apply_A I am happy to back this one out. the other functions have already been removed.

dlg0:
Let's leave it as `apply_A`

PR #205, opened 2019-12-10
Feature/realspace

stonecoldhughes:


bmcdanie:
is this WIP or not? change title if ready for review.

src/kron.cpp
```diff
@@ -0,0 +1 @@
+#include "kron.hpp"
```
bmcdanie: need to define public functions in the header and implementations in the cpp as in the other components.

src/kron.hpp
```diff
@@ -0,0 +1,212 @@
...
+    int const x_size)
+{
+  int greatest = x_size;
+  int r_prod   = 1;
+  int c_prod   = 1;
+
+  typename std::vector<
+      fk::matrix<P, mem_type::view, resrc>>::const_reverse_iterator iter;
+
+  for (iter = matrix.rbegin(); iter != matrix.rend(); ++iter)
```
bmcdanie: make this more expressive/readable if possible and/or add comments within function about the process you're going through here.

additionally, add comments before function signature describing the routine's purpose, the meaning of its arguments, etc.

finally, add asserts detailing preconditions for your arguments.

src/kron.hpp
```diff
@@ -0,0 +1,212 @@
...
+template<typename P, resource resrc>
+class batch_set
+{
+public:
+  batch_set(batch<P, resrc> &&left, batch<P, resrc> &&right,
+            batch<P, resrc> &&product)
+      : left(left), right(right), product(product)
+  {}
+
+  batch_set(batch<P, resrc> &&bs)
```
bmcdanie: if you're using the default move constructor, use `= default`

do we want a copy constructor?

this probably belongs in the batch component.

bmcdanie: may be something we can use for the other batches rather than the `using` we have there now

src/kron.hpp
```diff
@@ -0,0 +1,212 @@
...
+      : left(std::move(bs.left)), right(std::move(bs.right)),
+        product(std::move(bs.product))
+  {}
+
+  batch<P, resrc> left;
```
bmcdanie: if this is a public container for three `batches`, why not just use a container class?

is the entire purpose of this class to name the three `batch`es? not immediately opposed, just trying to figure out your reasoning.

src/kron.hpp
```diff
@@ -0,0 +1,212 @@
...
+            fk::vector<P, mem_type::view, resrc> const &x, int const y_size)
+      : alpha(alpha), beta(beta), in(0), out(1), y_size(y_size),
+        workspace({fk::vector<P, mem_type::owner, resrc>(workspace_len),
+                   fk::vector<P, mem_type::owner, resrc>(workspace_len)})
+  {
+    fk::vector<P, mem_type::view, resrc> init_w0(workspace[0], 0, x.size() - 1);
+    init_w0 = x;
+  }
+
+  batch_job(batch_job<P, resrc> &&bj)
```
bmcdanie: default?

src/kron.hpp
```diff
@@ -0,0 +1,212 @@
...
+        product(std::move(bs.product))
+  {}
+
+  batch<P, resrc> left;
+  batch<P, resrc> right;
+  batch<P, resrc> product;
+};
+
+template<typename P, resource resrc>
+class batch_job
```
bmcdanie: can this be amalgamated into the above class?

src/kron.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+  batch_job(batch_job<P, resrc> &&bj)
+      : batches(std::move(bj.batches)), alpha(bj.alpha), beta(bj.beta),
+        in(bj.in), out(bj.out), y_size(bj.y_size),
+        workspace(
+            {fk::vector<P, mem_type::owner, resrc>(std::move(bj.workspace[0])),
+             fk::vector<P, mem_type::owner, resrc>(std::move(bj.workspace[1]))})
+  {}
+
+  void add_batch_set(batch_set<P, resrc> &&bs)
```
bmcdanie: why are we passing this by rvalue reference?

src/kron.hpp
```diff
@@ -0,0 +1,212 @@
...
+  {
+    return workspace[in];
+  }
+
+  fk::vector<P, mem_type::owner, resrc> const &get_output_workspace()
+  {
+    return workspace[out];
+  }
+
+  std::vector<batch_set<P, resrc>> batches;
```
bmcdanie: do all these need to be public/non-const?

src/kron.hpp
```diff
@@ -0,0 +1,212 @@
...
+  P alpha;
+  P beta;
+  int in;
+  int out;
+  int y_size;
+  std::array<fk::vector<P, mem_type::owner, resrc>, 2> workspace;
+};
+
+template<typename P, resource resrc>
+fk::vector<P, mem_type::owner, resrc> execute_batch_job(batch_job<P, resrc> &bj)
```
bmcdanie: is this a copy of the `batched_gemm` in the `batch` component, except with your batch job api?

we don't need two such similar things - probably need to use the `batch job` stuff in both places or reuse the stuff there.

src/kron.hpp
```diff
@@ -0,0 +1,212 @@
...
+    batches.emplace_back(bs);
+
+    int const tmp = in;
+    in            = out;
+    out           = tmp;
+
+    return;
+  }
+
+  fk::vector<P, mem_type::owner, resrc> const &get_input_workspace()
```
bmcdanie: what is the benefit of this vs. two vector members named in and out?

src/kron.hpp
```diff
@@ -0,0 +1,212 @@
...
+{
+public:
+  batch_job(P const alpha, P const beta, int const workspace_len,
+            fk::vector<P, mem_type::view, resrc> const &x, int const y_size)
+      : alpha(alpha), beta(beta), in(0), out(1), y_size(y_size),
+        workspace({fk::vector<P, mem_type::owner, resrc>(workspace_len),
+                   fk::vector<P, mem_type::owner, resrc>(workspace_len)})
+  {
+    fk::vector<P, mem_type::view, resrc> init_w0(workspace[0], 0, x.size() - 1);
+    init_w0 = x;
```
bmcdanie: what is this variable?

src/kron.hpp
```diff
@@ -0,0 +1,212 @@
...
+                                         bj.y_size - 1);
+
+  fk::vector<P, mem_type::owner, resrc> r(v);
+
+  return r;
+}
+
+template<typename P, resource resrc>
+batch_job<P, resrc>
+kron_batch(std::vector<fk::matrix<P, mem_type::view, resrc>> const &matrix,
```
bmcdanie: describe in comments above routine: arguments, overall purpose of routine, relevant details

src/kron.hpp
```diff
@@ -0,0 +1,212 @@
...
+                           std::move(product));
+
+    job.add_batch_set(std::move(bs));
+  }
+
+  int stride = iter->nrows();
+  int v_size = x.size() / iter->ncols() * iter->nrows();
+  ++iter;
+
+  for (; iter != matrix.rend(); ++iter)
```
bmcdanie: let's improve readability and add comments throughout this routine

src/kron.hpp
```diff
@@ -0,0 +1,212 @@
...
+  {
+    int const read_stride  = stride * iter->ncols();
+    int const write_stride = stride * iter->nrows();
+    int const n_gemms      = v_size / read_stride;
+
+    /* create a batch of matrices */
+    batch<P, resrc> left(n_gemms, stride, iter->ncols(), stride, false);
```
bmcdanie: rather than creating these and then moving them into the batch set, why doesn't the batch set handle construction here, and then simply initialize the objects there in place?

src/kron_tests.cpp
```diff
@@ -0,0 +1,158 @@
+#include "kron.hpp"
+#include "tests_general.hpp"
+
+TEMPLATE_TEST_CASE("kron", "[kron]", double, float)
```
bmcdanie: even in tests, let's use const correctness

src/kron_tests.cpp
```diff
@@ -0,0 +1,158 @@
...
+  /* a simple 2 matrix and vector kron */
+  SECTION("kron_0")
+  {
+    fk::matrix<TestType, mem_type::owner, resource::device> a = {{2, 3},
+                                                                 {4, 5}};
+
+    fk::matrix<TestType, mem_type::owner, resource::device> b = {{6, 7},
+                                                                 {8, 9}};
+
+    const std::vector<fk::matrix<TestType, mem_type::view, resource::device>>
```
bmcdanie: east const

src/kron_tests.cpp
```diff
@@ -0,0 +1,158 @@
...
+        for (int j = 0; j < cols; ++j)
+        {
+          m(i, j) = 2;
+        }
+      }
+
+      return m.clone_onto_device();
+    };
+
+    auto m0 = matrix_all_twos(3, 4);
```
bmcdanie: const correctness throughout

src/kron_tests.cpp
```diff
@@ -0,0 +1,158 @@
...
+    REQUIRE(r.clone_onto_host() == correct);
+  }
+
+  SECTION("kron_1")
+  {
+    auto matrix_all_twos = [](int rows, int cols)
+        -> fk::matrix<TestType, mem_type::owner, resource::device> {
+      fk::matrix<TestType, mem_type::owner, resource::host> m(rows, cols);
+
+      for (int i = 0; i < rows; ++i)
```
bmcdanie: or: for (auto & elem : matrix) { elem = 2 };

src/kron_tests.cpp
```diff
@@ -0,0 +1,158 @@
...
+    fk::vector<TestType, mem_type::owner, resource::device> r =
+        execute_batch_job(bj);
+
+    fk::vector<TestType> correct(
+        std::vector<TestType>(y_size, x_size * (1 << (matrix.size() - 1)) * 3));
+
+    REQUIRE(r.clone_onto_host() == correct);
+  }
+
+  SECTION("kron_2")
```
bmcdanie: for testability let's make a kron test function that accepts matrices/vector to kron and performs the common operations. in these test cases, just init the matrices.

src/kron_tests.cpp
bmcdanie: if these functions work on device or host, need to cross product template these tests.

src/transformations.cpp
```diff
@@ -129,14 +130,8 @@ wavelet_to_realspace(PDE<P> const &pde, fk::vector<P> const &wave_space,
...
   std::vector<dimension<P>> const &dims = pde.get_dimensions();
-  /* determine the length of the real-space vector */
-  int prod = 1;
-  for (int i = 0; i < static_cast<int>(dims.size()); i++)
-  {
-    prod *= (dims[i].get_degree() * std::pow(2, dims[i].get_level()));
-  }
 
-  fk::vector<P> real_space(prod);
+  std::vector<batch_job<P, resource::host>> bjs;
```
bmcdanie: no problem with this routine overall. of course, our final (scalable) version needs to overlap much better with the workspace allocations we are already making. additionally, we don't want to perform these allocations every time step.

src/transformations.cpp
```diff
@@ -165,12 +160,28 @@ wavelet_to_realspace(PDE<P> const &pde, fk::vector<P> const &wave_space,
...
+    prod *= (dims[i].get_degree() * std::pow(2, dims[i].get_level()));
+  }
+  /* execute the batch job for each partial sum */
+  fk::vector<P> real_space(prod);
+  for (auto &bj : bjs)
+  {
+    real_space = real_space + execute_batch_job<P, resource::host>(bj);
   }
 
   return real_space;
```
bmcdanie: tests?

bmcdanie (changes requested):
aside from specific concerns, we need to see where we have functionality duplicated (e.g. execute batch job vs. batched gemm in batch component) and flatten our object structure here (batch sets + batch job is probably unnecessary). 

of course, the scalable routine will need to understand distribution, chunking, and the use of allocated workspaces already present. we may need to inform the allocation routines about our realspace transformation requirements.

are the existing realspace transform tests what you're using to test the top level method?

in CASL today if you want to discuss.

src/transformations.cpp
bmcdanie: probably reusing the old tests?

stonecoldhughes: Yes, I'm using the old tests for these.

src/kron.hpp
stonecoldhughes: I did this to maintain a flattened design while still getting the data-organizational benefit of a class. The batch_set object's purpose is simply to bundle together 3 batches. Inserting them into the class and then reaching down into the class later to initialize them makes the design slightly more hierarchical and I was trying to avoid that at all costs in this file.

src/kron.hpp
stonecoldhughes: I am making a copy here. The source is fk::vector view "x" and the destination is fk::vector owner workspace[ 0 ]. The destination vector may be longer than the source, so to copy I use this correctly sized intermediate view variable.

src/kron.hpp
stonecoldhughes: Because a physical vector labeled "in" would not always the input vector and the other physical one labeled "out" would not always be the output vector. They switch roles many times. This function frees the programmer to deal with logical "input" and "output" without worrying about which physical object "input" or "output" refers to on any given iteration.

src/kron.hpp
stonecoldhughes: this should disappear in the final scalable version.

src/kron.hpp
bmcdanie: but you aren't actually doing the work for the progammer - the progammer is still determining which is in and which is out, just through publicly accessible ints with no bounds checking.

if we want to provide this abstraction so the programmer doesn't have to worry about it, you could have these functions ingest the current dimension number and return the appropriate workspace.

src/kron.hpp
stonecoldhughes: yes, the entire purpose is to name the three batches

src/transformations.cpp
stonecoldhughes: yes, I am using the old tests. They just tested correct input/output of kron products so nothing needed to be changed.

src/transformations.cpp
stonecoldhughes: the final scalable version will not have allocations like this and will overlap better

src/kron.hpp
```diff
@@ -0,0 +1,212 @@
...
+  batch_job(batch_job<P, resrc> &&bj)
+      : batches(std::move(bj.batches)), alpha(bj.alpha), beta(bj.beta),
+        in(bj.in), out(bj.out), y_size(bj.y_size),
+        workspace(
+            {fk::vector<P, mem_type::owner, resrc>(std::move(bj.workspace[0])),
+             fk::vector<P, mem_type::owner, resrc>(std::move(bj.workspace[1]))})
+  {}
+
+  void add_batch_set(batch_set<P, resrc> &&bs)
+  {
```
bmcdanie: members should have a single responsibility - this function adds a batch set, but also swaps the input and output spaces. this makes the function less flexible, tying which storage areas I'm using for in/out with the order I am placing batches. 

also, if you intend for in/out to be managed exclusively within the class, they should be private members.

src/kron.cpp
stonecoldhughes: done

src/kron.hpp
stonecoldhughes: I do not think it's possible to use the default move constructor here.

src/kron.hpp
stonecoldhughes: the compiler claims that this function can't be defaulted

bmcdanie: that's because this isn't a move constructor - the argument isn't of same type.

I thought this was a move constructor - despite the rhs being named `bs`. is this intended to be a move constructor?

src/kron.hpp
bmcdanie: why?

src/kron.hpp
bmcdanie: is this ever invoked?

src/kron.hpp
bmcdanie:
```
batch_job(batch_job<P, resrc> const &) = default;
batch_job(batch_job<P, resrc> &&bj) = default;
```

src/kron.hpp
bmcdanie: this pattern seems inconsistent with RAII?

it isn't a problem for a class to manage its members.

src/kron.hpp
stonecoldhughes: asserts and extensive comments added

src/kron.hpp
stonecoldhughes: I think it could be, but there is a clear separation of tasks between the two and keeping them separate aligns with their separate purposes.

src/kron.hpp
stonecoldhughes: The final destination for a batch_set object is a container within batch_job. It is not used or needed anywhere else, and any reference to it is only useful for getting it there. This being the case, I wanted to minimize copies. Move semantics does those things.

bmcdanie: That to me says that batch_set should be a member of batch_job, rather than constructed externally and moved into place. The batch_job should be in a valid state when constructed unless we've knowingly decided to violate RAII for solid reasons. In that case, we'd want to document that we know we are and why because it's a red flag to anyone reading the code.

At any rate, functions with rvalue ref arguments outside of move constructors are potential bug sources - we need to clearly indicate in the name that we are taking ownership.

src/kron.hpp
stonecoldhughes: They either have to be public or I need to add getter functions for each one. Making them public seemed less bloated. I was able to make some of them const.

src/kron.hpp
stonecoldhughes: No, it actually calls batched_gemm underneath

src/kron.hpp
stonecoldhughes: done

src/kron.hpp
stonecoldhughes: done

src/kron.hpp
bmcdanie: Needing a one line getter isn't bloat - least privilege is worth it

src/kron.hpp
stonecoldhughes: The programmer never accesses "in" or "out" directly anywhere, so I went ahead and made them private. I don't understand your statement 

> the progammer is still determining which is in and which is out, just through publicly accessible ints with no bounds checking

The member variables "in" and "out" are not accessed anywhere outside of the class.

src/kron.hpp
stonecoldhughes: I made in/out private members and moved the functionality to swap the spaces into a different member function called after each batch_set is added to the batch_job

src/kron.hpp
stonecoldhughes: done

src/kron_tests.cpp
stonecoldhughes: done

stonecoldhughes: done

src/kron_tests.cpp
stonecoldhughes: done

src/kron_tests.cpp
stonecoldhughes: done

src/kron_tests.cpp
stonecoldhughes: done

src/kron_tests.cpp
stonecoldhughes: The bulk of all these test cases is initializing the matrices, vector, and correct result. The only common operations across the test cases are the two function calls `kron_batch()` and `execute_batch_job()` which by itself doesn't seem worth it. I'm not sure what else I could really offload to a function here.

src/transformations.cpp
stonecoldhughes: Removed the allocation at every timestep. Full integration with existing infrastructure is not possible until a scalable algorithm for the realspace transform is implemented.

src/kron_tests.cpp
bmcdanie: Well obviously you'd have to increase the share of common code. But, these are still edge cases, and if you prefer them this way I'd say it's fine. Trimming 5k+ lines of test code a couple weeks ago has me a little paranoid about test bloat.

But, on second look, definitely take out the bit shifting and add a comment about why this is the correct answer. Let's be as clear and straightforward as possible about what we are doing unless we have a really good reason not to be.

src/kron.hpp
stonecoldhughes: done

src/kron.hpp
stonecoldhughes: well when I deleted it, everything still compiled and ran so I supposed not...

CMakeLists.txt
```diff
@@ -183,7 +183,7 @@ target_link_libraries (time_advance PRIVATE batch fast_math pde tensors INTERFAC
 
 target_link_libraries (transformations
   PRIVATE connectivity distribution matlab_utilities pde program_options
-  quadrature tensors)
+  quadrature tensors batch)
```
bmcdanie: minor note - place in alpha order

src/batch.cpp
```diff
@@ -8,6 +8,274 @@
...
+          j * stride * matrix[i].nrows());
+
+      left.back().assign_entry(input, j);
+      right.back().assign_entry(fk::matrix<P, mem_type::view, resrc>(matrix[i]),
+                                j);
+      product.back().assign_entry(output, j);
+    }
+
+    /* output and input space will switch roles for the next iteration */
+    int const tmp = in;
```
bmcdanie: in future consider `std::swap`

src/batch.cpp
```diff
@@ -8,6 +8,274 @@
...
+
+  return;
+}
+
+template class batch_chain<double, resource::device>;
+template class batch_chain<double, resource::host>;
+template class batch_chain<float, resource::device>;
+template class batch_chain<float, resource::host>;
 
 // object to store lists of operands for batched gemm/gemv.
```
bmcdanie: what is this comment related to?

src/main.cpp
```diff
@@ -223,11 +223,24 @@ int main(int argc, char **argv)
 #ifdef ASGARD_IO_HIGHFIVE
   auto output_dataset = initialize_output_file(initial_condition);
 
-  fk::vector<prec> const initial_condition_real = wavelet_to_realspace<prec>(
-      *pde, initial_condition, table, default_workspace_cpu_MB);
+  int prod = real_solution_size(*pde);
```
bmcdanie: const and/or roll into below function call

src/transformations.cpp
```diff
@@ -165,15 +168,37 @@ wavelet_to_realspace(PDE<P> const &pde, fk::vector<P> const &wave_space,
...
+
+template<typename P>
+int real_solution_size(PDE<P> const &pde)
+{
+  /* determine the length of the realspace solution */
+  std::vector<dimension<P>> const &dims = pde.get_dimensions();
+  int prod                              = 1;
+  for (int i = 0; i < static_cast<int>(dims.size()); i++)
+  {
+    prod *= (dims[i].get_degree() * std::pow(2, dims[i].get_level()));
```
bmcdanie: in future consider std::accumulate

src/batch.hpp
```diff
@@ -76,6 +77,69 @@ class batch
...
+extern template class batch_chain<float, resource::host>;
+
+/* Calculates necessary workspace length for the Kron algorithm. See .cpp file
+ * for more details */
+template<typename P, resource resrc>
+int calculate_workspace_len(
+    std::vector<fk::matrix<P, mem_type::view, resrc>> const &matrix,
+    int const x_size);
+
+/* external explicit instantiations */
```
bmcdanie: these are not instantiations - in fact, they prevent instantiations.

https://isocpp.org/wiki/faq/cpp11-language-templates#extern-templates

src/batch.cpp
```diff
@@ -8,6 +8,274 @@
...
+  assert(matrix.size() > 0);
+
+  /* ensure "x" is correct size - should be the product of all the matrices
+     respective number of columns */
+  assert(x.size() == std::accumulate(matrix.begin(), matrix.end(), 1,
+                                     [](int const i, auto const &m) {
+                                       return i * m.ncols();
+                                     }));
+
+  int in  = 0;
```
bmcdanie, line 105: put a comment about what these are here at the declaration site

src/batch.cpp
```diff
@@ -8,6 +8,274 @@
...
+  int out = 1;
+
+  /* ensure the workspaces are big enough for the problem */
+  int const workspace_len = calculate_workspace_len(matrix, x.size());
+  assert(workspace[0].size() >= workspace_len);
```
bmcdanie, line 110: should probably assert `workspace.size() == 2`

bmcdanie (changes requested):
fix these minor notes and let's get this merged in please

src/batch.cpp
stonecoldhughes: I like that

src/batch.hpp
stonecoldhughes: They do prevent implicit instantiations. Since I define the class template's functions outside of the .hpp where it's declared, I already would not be getting implicit instantiations right?

bmcdanie: Yes, that's my theory. The compiler doesn't have the information necessary to instantiate them in other translation units. I believe we could strip the externs out, although we would still need the explicit instantiations in the cpp. At some point I am going to look into that.

CMakeLists.txt
bmcdanie: @stonecoldhughes place to in alpha order

stonecoldhughes: done

src/batch.cpp
stonecoldhughes: vestigial comment removed

src/main.cpp
stonecoldhughes: done

src/batch.cpp
stonecoldhughes: done up above as the 2nd check

src/batch.cpp
stonecoldhughes: done

src/kron_tests.cpp
stonecoldhughes: done

src/batch.cpp
```diff
@@ -8,8 +8,274 @@
...
+
+Max space needed is maximum of the above sizes.
+
+Improvement idea: Each stage alternates between 2 workspaces,
+so the first workspace needs to be the maximum of round 0 and 2, while the other
+needs to be the max of round 1 and 3. As it stands now, each workspace is the
+size of the max of all rounds.
+*/
+template<typename P, resource resrc>
+int calculate_workspace_len(
```
grahamlopez: it's a long name already (I approve); let's just take `_len`-->`_length`

src/batch.cpp
```diff
@@ -8,8 +8,274 @@
 #include "connectivity.hpp"
 #include "lib_dispatch.hpp"
 #include "tensors.hpp" // for views
+/*
```
grahamlopez, line 12: very nice comment; thank you!

src/batch.cpp
```diff
@@ -8,8 +8,274 @@
...
+Max space needed is maximum of the above sizes.
+
+Improvement idea: Each stage alternates between 2 workspaces,
+so the first workspace needs to be the maximum of round 0 and 2, while the other
+needs to be the max of round 1 and 3. As it stands now, each workspace is the
+size of the max of all rounds.
+*/
+template<typename P, resource resrc>
+int calculate_workspace_len(
+    std::vector<fk::matrix<P, mem_type::view, resrc>> const &matrix,
```
grahamlopez: `matrix` --> `matrices`?
(a `std::vector<>` is not a matrix in this case)

src/batch.cpp
```diff
@@ -8,8 +8,274 @@
...
+
+    int const size = x_size / c_prod * r_prod;
+    if (size > greatest)
+      greatest = size;
+  }
+
+  return greatest;
+}
+
+/* explicit instantions of function above */
```
grahamlopez: previous convention has been to group these at the bottom of the file

src/main.cpp
```diff
@@ -286,12 +306,15 @@ int main(int argc, char **argv)
     {
       update_output_file(output_dataset, host_space.x);
     }
-    /* write realspace output to file */
+
+    /* transform from wave space to real space */
```
grahamlopez: s/wave/wavelet

src/batch.cpp
stonecoldhughes: name extended

src/batch.cpp
stonecoldhughes: gone

bmcdanie:
test this please

src/batch.cpp
stonecoldhughes: done

src/main.cpp
stonecoldhughes: changed

src/batch.cpp
stonecoldhughes: done

bmcdanie (approved):

PR #208, opened 2019-12-17
Fix time advance bug

bmcdanie:


tyounkin:
@bmcdanie, did the sign of this comparison need to be changed?

bmcdanie:
> @bmcdanie, did the sign of this comparison need to be changed?

! good catch @tyounkin .

tyounkin:
@bmcdanie @dlg0 I anticipate this will result in some of the time advance testing breaking which will lead to a discussion about: What is the right precision to compare PDE solutions between Matlab and C++ versions?

bmcdanie:
all of the time advance tests pass within the threshhold still except for fokkerplanck 2d, which has always required a looser tolerance than the rest - could be a potential bug?

dlg0:
Indeed it could be. I'd suggest @tyounkin compare the matlab with the C++ for that case to see if anything obvious shows up. Perhaps this is a useful case to actually plot the 2D output from the C++ in matlab?

PR #209, opened 2019-12-18
Fix TA bug

bmcdanie:
actually fix it this time - thanks @tyounkin

### DIFF
--- a/src/time_advance_tests.cpp
+++ b/src/time_advance_tests.cpp
@@ -28,7 +28,10 @@ template<typename P>
 void time_advance_test(int const level, int const degree, PDE<P> &pde,
                        int const num_steps, std::string const filepath,
                        bool const full_grid                            = false,
-                       std::vector<std::string> const &additional_args = {})
+                       std::vector<std::string> const &additional_args = {},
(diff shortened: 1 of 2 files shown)


bmcdanie:
test this please

src/time_advance_tests.cpp
```diff
@@ -28,7 +28,10 @@ template<typename P>
 void time_advance_test(int const level, int const degree, PDE<P> &pde,
                        int const num_steps, std::string const filepath,
                        bool const full_grid                            = false,
-                       std::vector<std::string> const &additional_args = {})
+                       std::vector<std::string> const &additional_args = {},
+                       double const tol_factor                         = 1e4)
```
grahamlopez: this gets set locally as `eps_multiplier` down below. should this default parameter become `eps_multiplier`?

Also, other places in the file that call this are using `tol_fac` locally. probably those need to change as well.

bmcdanie: this is something weird git did with the rebase, should be fixed.

bmcdanie:
test this please

grahamlopez:
test this please

PR #210, opened 2019-12-19
Default quad num of degree + 1

### LINKED ISSUE #82: make quad_num global and set to deg+1
ASGarD
    asgardasgard
    Issues
    #82

Open
Opened 1 week ago by Green, David L
Close issue
New issue
make quad_num global and set to deg+1

The quad_num variable is used in a few places. It sets the number of quadrature points over which to perform an integration. The Legendre functions over which the integrations are being performed have a maximum degree (set by the global deg). So, to avoid extra work at low deg and errors at high deg, we should set quad_num = deg+1 consistently throughout the code.

I'd have a go at this, but I am 100% certain that my approach to a global variable will be embarrassing.

@bgl had suggested some set of global variables to handle other things. Perhaps we can get one variable done so I can see where to add this?
Related issues
0

Green, David L @dg6 added bug feature hygeine labels 1 week ago

Graham Lopez @bgl · 1 week ago
Owner

At one point, I had suggested some kind of globally-accessible const singleton object, but either it never gained priority attention from us, or we tried and couldn't get it to work.

I don't want to compromise with a set of global variables; that's bad karma - the C++ gods will frown on us and our crops will struggle.

Graham Lopez
Graham Lopez @bgl · 1 week ago
Owner

I gave this a go and modified the first argument to legendre_weights() to be degree + 1 in both coefficients.cpp and transformations.hpp, but NOT basis.cpp since these are set much higher (2^6).

Not only will it require a change to some of the matlab test input generation (coefficients_tests.cpp), but also some tolerances we are using (time_advance_tests.cpp and batch_tests.cpp) that we might want to think a little bit about.

bmcdanie:
replaces #206, closes #82


src/quadrature.cpp
```diff
@@ -138,92 +138,89 @@ legendre(fk::vector<P> const domain, int const degree,
...
   // of the roots x_roots done next
   fk::vector<P> const x_linspace =
-      linspace(static_cast<P>(-1.0), static_cast<P>(1.0), polynomial_degree);
+      linspace(static_cast<P>(-1.0), static_cast<P>(1.0), num_points);
 
   //% Initial guess at the roots for the Legendre polynomial of degree
-  // polynomial_degree
-  // x_roots=cos((2*(0:polynomial_degree-1)'+1)*pi/(2*(polynomial_degree-1)+2))+(0.27/polynomial_degree)*sin(pi*x_linspace*((polynomial_degree-1)/(polynomial_degree+1);
+  // num_points
+  // x_roots=cos((2*(0:num_points-1)'+1)*pi/(2*(num_points-1)+2))+(0.27/num_points)*sin(pi*x_linspace*((num_points-1)/(num_points+1);
```
grahamlopez: fix this line-wrapping

src/coefficients.cpp
```diff
@@ -64,15 +64,10 @@ generate_coefficients(dimension<P> const &dim, term<P> const &term_1D,
...
-  // FIXME should this be order dependent?
-  // FIXME is this a global quantity??
-  int const quad_num = 10;
-
   // get quadrature points and quadrature_weights.
   // we do the two-step store because we cannot have 'static' bindings
-  static auto const legendre_values =
-      legendre_weights<double>(quad_num, -1.0, 1.0);
+  static auto const legendre_values = legendre_weights<double>(
+      default_quad_number(dim.get_degree()), -1.0, 1.0);
```
grahamlopez: should this be degree + 1?

bmcdanie: fixed

src/quadrature.hpp
```diff
@@ -15,10 +15,20 @@ std::enable_if_t<std::is_floating_point<P>::value, std::array<fk::matrix<P>, 2>>
 legendre(fk::vector<P> const domain, int const degree,
          legendre_normalization const norm = legendre_normalization::lin);
 
+inline int default_quad_number(int const degree)
+{
+  assert(degree > 0);
+  static int constexpr minimum_quadrature = 10;
+  return std::max(minimum_quadrature, degree);
```
grahamlopez: build in the `+1` into this function instead?

src/quadrature.hpp
```diff
@@ -15,10 +15,20 @@ std::enable_if_t<std::is_floating_point<P>::value, std::array<fk::matrix<P>, 2>>
 legendre(fk::vector<P> const domain, int const degree,
          legendre_normalization const norm = legendre_normalization::lin);
 
+inline int default_quad_number(int const degree)
```
grahamlopez: i don't think I like this helper function. Why not just have `legendre_weights()` do this for us?
change my mind.

grahamlopez: to be specific, I don't like it because the user of the APIs that need it have to know about it and remember to use it

src/coefficients.cpp
grahamlopez: not fixed.

src/quadrature.hpp
bmcdanie: I would also prefer to build it in.

the problem is here: https://github.com/bmcdanie/ASGarD/blob/b6a120730adc4db68fa37839d75906034dcc3374/src/transformations.cpp#L106.

we need to use a quad num == degree.

src/quadrature.cpp
grahamlopez: fix this line-wrapping (2nd call)

bmcdanie: done

PR #211, opened 2019-12-20
Ensure working Fortran compiler if CMake needs to build OpenBLAS

stonecoldhughes:
Otherwise don't even care

edit: closes #211 .

### DIFF
--- a/contrib/contrib.cmake
+++ b/contrib/contrib.cmake
@@ -123,6 +123,17 @@ if (NOT LINALG_LIBS_FOUND)
   else (NOT DEFINED LINALG_LIBS)
     message (STATUS "OpenBLAS not found. We'll build it from source.")
 
+    # OpenBLAS will build without Fortran, but it will silently neglect to build vital functions
+    include( CheckLanguage )
+    message( STATUS "Checking for Fortran compiler..." )
+    check_language(Fortran)
+    if(CMAKE_Fortran_COMPILER)
+      enable_language(Fortran)
+      message( STATUS "Fortran compiler found" )
+    else()
+      message( FATAL_ERROR "Fortran compiler missing - required to build OpenBLAS" )
+    endif()
+
     include (ExternalProject)
     ExternalProject_Add (openblas-ext
       UPDATE_COMMAND ""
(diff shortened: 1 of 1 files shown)


contrib/contrib.cmake
```diff
@@ -107,6 +107,17 @@ if (NOT LINALG_LIBS)
   else (NOT DEFINED LINALG_LIBS)
     message (STATUS "OpenBLAS not found. We'll build it from source.")
 
+    # OpenBLAS will build without Fortran, but it will silently neglect to build vital functions
+    include( CheckLanguage )
```
bmcdanie: does this just print the warning? I think we want to fail completely here, right?

stonecoldhughes: It does print the warning but I don't see any reason for it to fail there.

bmcdanie: the point of this PR is to make explicit the problems that occur when we build openblas without a fortran compiler. if we don't have fortran, but are trying to build openblas, we want to fail loudly - one warning message behind several screens of openblas compile barf won't help us.

grahamlopez: if it doesn't fail, won't the build just fail later on when trying to link? Generally, we'd like for things to fail at the cause, not the result.

bmcdanie: it doesn't help us to continue building under those conditions, right?

contrib/contrib.cmake
```diff
@@ -107,6 +107,17 @@ if (NOT LINALG_LIBS)
...
+    message( STATUS "Checking for Fortran compiler..." )
+    check_language(Fortran)
+    if(CMAKE_Fortran_COMPILER)
+      enable_language(Fortran)
+      message( STATUS "Fortran compiler found" )
+    else()
+      message( FATAL_ERROR "Fortran compiler missing - required to build OpenBLAS" )
```
bmcdanie: or does this line cause build abort?

contrib/contrib.cmake
bmcdanie: to be clear, when I say "fail here", I mean fail when we don't have a fortran compiler, not at this specific line.

contrib/contrib.cmake
bmcdanie: @stonecoldhughes update - this is the line that causes abort.

PR #214, opened 2019-12-31
improve perf of combine dims using views

### LINKED ISSUE #76: views for combine_dimensions
ASGarD
    asgardasgard
    Issues
    #76

Open
Opened 2 weeks ago by Graham Lopez
Close issue
New issue
views for combine_dimensions

combine_dimensions() is still in the top 3 time consumers in setup.

This is just a hunch, but maybe using views instead of .set_subvector() could help us out here
Related issues
0

Graham Lopez @bgl added performance label 2 weeks ago
Green, David L
Green, David L @dg6 · 2 weeks ago
Owner

Did the profiling get rolled into the cmake? If so, how would I go about using it? (I'm thinking about taking this task on)
Green, David L
Green, David L @dg6 · 2 weeks ago
Owner

From what I can tell (using Allinea Map), the call to kron_d dominates the runtime of combine_dimensions. I did take a look at implementing a view instead of the .set_subvector(), but from what I could see, the .set_subvector() doesn't require you telling it both the start and end position, only the start. Views require both, as well as creating the view (which has to happen every time within the loop?) which makes for more, perhaps less clear code.

So perhaps we can close this one out?
Graham Lopez
Graham Lopez @bgl · 1 week ago
Owner

The profiling is rolled into the cmake. All of the related options to enable it are named with the pattern ASGARD_PROFILE_* You can also see more information about it on the profiling wiki page: https://code.ornl.gov/asgard/asgard/wikis/user-guide/profiling

It makes sense that the bulk of the time is in kron_d(); this is where the math happens. If we want to speed that up, it will likely require an algorithmic change.

I think there is still the possibility to use a view for combine_dimensions() and possibly still a win to be had there.
1. We would make a view for each loop iteration, but this is a very cheap operation (no heap allocations).
2. We would need to provide the start and stop, but these are directly calculable from the loop index. It might be something along the lines of:

    start_idx = i * std::pow(degree, num_dims);
    end_idx = (i+1) * std::pow(degree, num_dims) -1;
    fk::vector<P, memtype::view>(combined, start_idx, end_idx);

bmcdanie:
closes #76.

we could also consider optimizing `kron_d` if this becomes a performance bottleneck in future - it isn't now.

### DIFF
--- a/src/transformations.cpp
+++ b/src/transformations.cpp
@@ -219,10 +219,11 @@ combine_dimensions(int const degree, element_table const &table,
           degree > 1 ? (((id + 1) * degree) - 1) : index_start;
       kron_list.push_back(vectors[j].extract(index_start, index_end));
     }
-    fk::vector<P> const partial_result =
-        kron_d(kron_list, kron_list.size()) * time_scale;
-    combined.set_subvector((i - start_element) * std::pow(degree, num_dims),
-                           partial_result);
+    int const start_index = (i - start_element) * std::pow(degree, num_dims);
+    int const stop_index  = start_index + std::pow(degree, num_dims) - 1;
+    fk::vector<P, mem_type::view> combined_view(combined, start_index,
+                                                stop_index);
+    combined_view = kron_d(kron_list, kron_list.size()) * time_scale;
   }
   return combined;
 }


grahamlopez:
test this please

PR #216, opened 2019-12-31
Fix deg and lev parsing; fix for deg=1; allow subvectors of size 1

### LINKED ISSUE #215: specifying level or degree alone causes crash
If either level or degree is specified as a command line option, and both are not specified, the program will crash. not great.

bmcdanie:
closes #215

### DIFF
--- a/src/pde/pde_base.hpp
+++ b/src/pde/pde_base.hpp
@@ -369,11 +369,13 @@ public:
       // supported
       for (dimension<P> &d : dimensions_)
       {
-        if (num_levels != -1) {
+        if (num_levels != -1)
+        {
           assert(num_levels > 1);
           d.set_level(num_levels);
         }
-        if (degree != -1) {
+        if (degree != -1)
+        {
           assert(degree > 0);
           d.set_degree(degree);
         }
--- a/src/pde/pde_base.hpp
+++ b/src/pde/pde_base.hpp
@@ -365,16 +365,18 @@ public:
     // if default lev/degree not used
     if (num_levels != -1 || degree != -1)
     {
-      assert(num_levels > 1);
-      assert(degree > 0);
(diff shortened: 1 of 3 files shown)


src/tensors.hpp
```diff
@@ -1306,7 +1306,7 @@ fk::vector<P, mem, resrc>::extract(int const start, int const stop) const
 {
   assert(start >= 0);
   assert(stop < this->size());
-  assert(stop > start);
+  assert(stop >= start);
```
grahamlopez: I'm not convinced that our constructors handle allocations of `-1` very well

src/pde/pde_base.hpp
```diff
@@ -365,16 +365,20 @@ class PDE
     // if default lev/degree not used
     if (num_levels != -1 || degree != -1)
     {
```
grahamlopez: ah, so when you said that we crashed before, we were actually hitting `asserts()`? If so, that still sucks, but not nearly as troubling.

bmcdanie: yes

src/tensors.hpp
bmcdanie: we do appear to be missing some asserts in some of the constructors, I can add those

bmcdanie: added

grahamlopez: I'm referring to line 1314

bmcdanie: I think the least possible value of `sub_size` there is one, unless I've made a mistake. it used to permit a minimum vector size of 2, but I changed that to allow for single element subvects.

grahamlopez: you are right; I'm doing too many things at once

grahamlopez:
test this please

Issue #156, opened 2019-09-18
OpenBLAS check thwarts Cray compiler wrappers


bcfriesen:
Compiler wrappers on Cray XC systems (`cc`, `CC`, `ftn`) include appropriate include and link flags for BLAS/LAPACK implementations provided by Cray LibSci. These libraries sometimes have obscure names, e.g., `libsci_intel_mpi.so`. However, the OpenBLAS check in `contrib/contrib.cmake` makes it difficult to convince CMake that the BLAS requirement in ASGard is satisfied without providing explicit links to a linear algebra library.

Would it be possible to add a 'last-resort' mechanism to the BLAS check during configuration which simply attempts to compile a BLAS code with the given compiler and compiler flags? If that fails, then CMake could proceed to download OpenBLAS. Currently the way I work around this on the Cori system at NERSC is by removing the entire BLAS check in `contrib/contrib.cmake`.

grahamlopez:
Did you try `-DASGARD_BLAS_PATH` or `-DASGARD_LAPACK_PATH`? I'm not sure that it would be sufficient, but we should understand the limitation.
I think a better 'last-resort' (if necessary) might be to allow the above options to take the full path to a file, including the file name. Then the responsibility moves to the user to ensure that whatever this file is provides the necessary symbols to compile/link.
Does this sound reasonable?

bcfriesen:
Hi @grahamlopez,

> Did you try -DASGARD_BLAS_PATH or -DASGARD_LAPACK_PATH? 

Trying either/both of those CMake flags did not work.

> I think a better 'last-resort' (if necessary) might be to allow the above options to take the full path to a file, including the file name. Then the responsibility moves to the user to ensure that whatever this file is provides the necessary symbols to compile/link.
Does this sound reasonable?

I'm not sure this is the best approach. The problem is that LibSci provides different implementations based on whether the code is serial, threaded, or MPI-parallel. The compiler wrapper figures out which one to link to your program based on the flags that you provide. E.g., with the Intel compiler:

```
friesen@cori12:dgemm> cc dgemm_libsci.c
friesen@cori12:dgemm> ldd a.out|grep libsci
        libsci_intel.so.5 => /opt/cray/pe/lib64/libsci_intel.so.5 (0x00002aaaab52e000)
friesen@cori12:dgemm> cc -qopenmp dgemm_libsci.c
friesen@cori12:dgemm> ldd a.out|grep libsci
[...]

PR #218, opened 2020-01-03
CMake: adjust logic to use Cray LibSci instead of OpenBLAS

bcfriesen:
CMake plays funny tricks on the `find_package(LAPACK)` and `find_package(BLAS)`
commands when trying to detect Cray LibSci. So the CMake logic needed to be
adjusted to handle this case, in order to avoid building OpenBLAS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,14 @@
-cmake_minimum_required (VERSION 3.13)
+cmake_minimum_required (VERSION 3.13) # remove cray check below when >= 3.16
 project (asgard
   VERSION 0.1.0
   LANGUAGES CXX
 )
+
+if(${CMAKE_SYSTEM_NAME} STREQUAL "CrayLinuxEnvironment")
+  if (CMAKE_VERSION VERSION_LESS "3.16")
+    message (WARNING "To use Cray LibSci instead of building OpenBLAS, you must use CMake >= 3.16.")
+  endif ()
+endif ()
 
 ###############################################################################
 ## Set up the compiler and general global build options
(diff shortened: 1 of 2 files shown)


grahamlopez:
okay to test

grahamlopez:
@bcfriesen did you want to bump the `cmake_minimum_required (VERSION 3.13)` accordingly?

bcfriesen:
I can if you'd like - it will probably make building the code harder since v3.16 is the newest release, and I doubt it is included in any `apt-get`/`zypper`/`dnf` repository yet. So almost anyone building the code will need to first build 3.16 by hand. But if that's not a deal-breaker for, then I'll make the change.

grahamlopez:
sorry I wasn't clear. Is 3.16 needed for Asgard to build on Cori, or does your fix work with the current version as well?

bcfriesen:
> Is 3.16 needed for Asgard to build on Cori, or does your fix work with the current version as well?

3.16 is needed required if one wants ASGarD to use LibSci on Cori. With any version older than 3.16, it will not detect LibSci correctly, and will build OpenBLAS.

grahamlopez:
Thanks, that makes sense.

anyone have opinions? I would be fine to leave the versioning as it is right now, as our current list of Cori users are adept enough to load the right version of cmake if they want libsci.

bmcdanie:
Prefer current versioning, for the reason you state.

bmcdanie:
closes #156 @bcfriesen @grahamlopez ?

bcfriesen:
Looks good to me, thanks!

grahamlopez:
Hang on, I'm looking into simplifying the logic a bit

grahamlopez:
test this please

grahamlopez:
test this please

grahamlopez:
test this please

grahamlopez:
test this please

Issue #185, opened 2019-10-31
assert on cpu workspace limit when making host space allocation


bmcdanie:

bmcdanie:
closed with #227 .

PR #227, opened 2020-01-13
assert on cpu workspace limit in host workspace constructor (issue #185)

stonecoldhughes:
…space.

### DIFF
--- a/src/chunk.cpp
+++ b/src/chunk.cpp
@@ -106,9 +106,11 @@ rank_workspace<P>::get_unit_vector() const
 
 template<typename P>
 host_workspace<P>::host_workspace(PDE<P> const &pde,
-                                  element_subgrid const &grid)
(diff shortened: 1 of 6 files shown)


bmcdanie (approved):
good work.

we want to continue to evolve the use of these limits - eventually should be settable from the cmake and consistently applied as we've gotten to w/ the gpu limits.

src/chunk.cpp
```diff
@@ -100,9 +100,11 @@ rank_workspace<P>::get_unit_vector() const
 
 template<typename P>
 host_workspace<P>::host_workspace(PDE<P> const &pde,
-                                  element_subgrid const &grid)
+                                  element_subgrid const &grid,
+                                  int const memory_limit_MB)
```
grahamlopez, line 110: should this be defaulted?

Issue #126, opened 2019-07-25
Notes on main


dlg0:
Some notes on `main.cpp`
- [x] it does not work :slightly_smiling_face: i.e., while the time_advance tests pass, i get all zeros out of running the actual code for the `fokkerplanck_1d_4p2` PDE (see my `fokkerplanck1_4p1a` branch).
- [x] system() should be renamed to something more descriptive.
- [x] system.batch_output is the solution vector. Perhaps we could name it as such? (i.e., avoiding lines like `diff = system.batch_output - analytic_solution_t`
- [x] what is `x_orig` and where is it used?

bmcdanie:
we no longer have these items in main @dlg0. I may use this issue as an opportunity to re-evaluate main for readability, though.

bmcdanie:
closed with #229 - but need to periodically reevaluate main for conciseness and clarity.

PR #229, opened 2020-01-15
issue #126 - Clean up main()

stonecoldhughes:
Fixes issue #126 , which was outdated. Cleaned up main by moving combine_dimensions blocks into a function.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -104,16 +104,18 @@ int main(int argc, char **argv)
 
   // -- generate initial condition vector.
   node_out() << "  generating: initial conditions..." << '\n';
+
   fk::vector<prec> const initial_condition = [&pde, &table, &subgrid,
                                               degree]() {
-    std::vector<fk::vector<prec>> initial_conditions;
+    std::vector<vector_func<prec>> v_functions;
+
     for (dimension<prec> const &dim : pde->get_dimensions())
     {
-      initial_conditions.push_back(
-          forward_transform<prec>(dim, dim.initial_condition));
+      v_functions.push_back(dim.initial_condition);
     }
-    return combine_dimensions(degree, table, subgrid.col_start,
-                              subgrid.col_stop, initial_conditions);
+
+    return transform_and_combine_dimensions(
+        *pde, v_functions, table, subgrid.col_start, subgrid.col_stop, degree);
   }();
 
   // -- generate source vectors.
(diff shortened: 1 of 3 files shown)


grahamlopez:
Please make the title of pull requests say what they do, and if they are fixing an issue, include that in the description (which should also expand on the PR title).

bmcdanie (commented):
+67/-18 is not a great net for a DRY'ing out PR. hmm...

may need to rethink changing this, or have the pass in the function arrays (`source.source_funcs[i]`, for instance) so that we can extract more of the shared code. I'm in CASL today if you want to discuss.

bmcdanie:
actually, it looks like the PDE holds these functions in vectors already - is it redundant to repackage them? they can probably be passed straight in.

bmcdanie:
example:

```  
  121   // -- generate source vectors.
  122   // these will be scaled later according to the simulation time applied
  123   // with their own time-scaling functions
  124   node_out() << "  generating: source vectors..." << '\n';
  125   std::vector<fk::vector<prec>> const initial_sources = [&pde, &table, &subgrid,
  126                                                          degree]() {
  127     std::vector<fk::vector<prec>> initial_sources;
  128     for (source<prec> const &source : pde->sources)
  129     {
  130       initial_sources.push_back(transform_and_combine_dimensions(
  131           *pde, source.source_funcs, table, subgrid.row_start, subgrid.row_stop,
  132           degree));
  133     }
  134     return initial_sources;
  135   }();
```

src/transformations.hpp
```diff
@@ -31,6 +31,24 @@ fk::vector<P>
 combine_dimensions(int const, element_table const &,
                    std::vector<fk::vector<P>> const &, P const = 1.0);
 
+template<typename P>
+fk::vector<P>
+transform_and_combine_dimensions(PDE<P> const &pde,
```
bmcdanie: let's accept the functions by const ref and const the int args here

src/main.cpp
```diff
@@ -104,16 +104,18 @@ int main(int argc, char **argv)
 
   // -- generate initial condition vector.
   node_out() << "  generating: initial conditions..." << '\n';
+
   fk::vector<prec> const initial_condition = [&pde, &table, &subgrid,
                                               degree]() {
-    std::vector<fk::vector<prec>> initial_conditions;
+    std::vector<vector_func<prec>> v_functions;
```
bmcdanie, line 110: trim this unnecessary step if possible, see my example

src/transformations.cpp
```diff
@@ -9,6 +9,35 @@
...
 #include <numeric>
 
+template<typename P>
+fk::vector<P>
+transform_and_combine_dimensions(PDE<P> const &pde,
+                                 std::vector<vector_func<P>> v_functions,
+                                 element_table const &table, int start,
+                                 int stop, int degree)
+{
+  std::vector<fk::vector<P>> dimension_components;
```
bmcdanie: assert stop and start >= 0, stop >= start, degree  > 0, etc.

bmcdanie: also that the v_functions match num pde dims

bmcdanie (changes requested):
add these few items and let's get a simple test for this in transformations tests since it's a public function.

src/transformations.hpp
stonecoldhughes: done

src/transformations.cpp
stonecoldhughes: done

src/main.cpp
bmcdanie: can't, because no vector of functions directly available here as in the source case. need to consider in pde spec rewrite.

bmcdanie (approved):
thanks.

PR #252, opened 2020-03-05
Fix performance bug

bmcdanie:
silly mistake placing this calculation in the innermost batch building loop...fixed


stonecoldhughes (approved):
good catch

PR #255, opened 2020-03-10
ax extern declarations

### LINKED ISSUE #254: remove extern template declarations when not required to prevent extraneous instantiation

bmcdanie:
closes #254 .

we'd discussed previously removing these and I think it's time.

the out-of-line definitions for these functions in a .cpp means that the compiler can't arbitrarily re-instantiate them outside their home translation unit anyway.


stonecoldhughes (commented):
approved

stonecoldhughes (approved):

PR #257, opened 2020-03-10
let cmake handle flags for openmp

### LINKED ISSUE #247: Building issues on macOS Catalina
I tried to build ASGarD on macOS Catalina with Apple clang version 11.0.0 (clang-1100.0.33.17). I followed the instructions on wiki - Specific build instructions for OS X, which was tested for Mojave before. When I cmake, I got the following error:clang: error: 

unsupported option '-fopenmp'

Commenting out Line 70 of CMakeLists.txt (add_compile_options ("-fopenmp")) solved this problem.

Then, I got the following error:

[ 56%] Building CXX object CMakeFiles/time_advance-tests.dir/src/time_advance_tests.cpp.o
In file included from /Users/ocekmer/workspace/ASGarD/asgard/src/time_advance_tests.cpp:6:
/Users/ocekmer/workspace/ASGarD/asgard/testing/tests_general.hpp:73:22: error: implicit
   instantiation of undefined template 'std::__1::basic_ostringstream<char,
   std::__1::char_traits<char>, std::__1::allocator<char> >'
 std::ostringstream out;
           ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/iosfwd:137:32: note: 
   template is declared here
  class _LIBCPP_TEMPLATE_VIS basic_ostringstream;
                ^

This one was solved by adding sstream library in testing/tests_general.hpp.

bmcdanie:
some compilers (like Apple's clang) use different openmp flags - cmake appears to handle this well, so we can rely on the build system rather than manually specifying flags.

closes #247.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,8 +65,17 @@ if(FOUND_MKL GREATER_EQUAL 0)
   set(ASGARD_USE_MKL TRUE)
 endif()
 
+
 if(ASGARD_USE_OPENMP)
-  find_package(OpenMP REQUIRED)
+  find_package(OpenMP REQUIRED)  
+  if(ASGARD_USE_MKL)
+    add_compile_options ("-fopenmp")
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+	message(FATAL_ERROR 
+	  "MKL and apple clang++ require conflicting 
+	  openmp flags - build configuration not supported")
+    endif()
+  endif()
 endif()
 
 if(ASGARD_USE_CUDA)


stonecoldhughes (approved):
literally what cmake is for - approved

bmcdanie:
> literally what cmake is for - approved

unfortunately, I just tested on Cori, and this is broken when we link with MKL.

ugh.

the worst build issues are always intel library related.

note to self - need to cover both apple case (CMake only flags) and intel case (must manually specify `-fopenmp`) or print some warning

PR #263, opened 2020-03-21
WIP: refactor/batch

### LINKED ISSUE #85: improve batches clarity
ASGarD
    asgardasgard
    Issues
    #85

Open
Opened 1 week ago by Graham Lopez
Close issue
New issue
improve batches clarity

Right now, it is fairly opaque how the batch objects, operands sets, input/output vectors, time_advance() routine all hang together. But this is one of the main cores of our algorithm, so we need to improve its accessibility.

One way to start on this would be to make the relationship between x/y (input/output to time advance) and the batch inputs/outputs clearer. This could be done by replacing the `std::vector<batch_operand_sets>`` with a struct that also held the "batch_input" and "batch_output", and probably a setter. Then the arguments to the time advance are truly only its input/output, and within we would explicitly and more clearly set the input as we wanted.

This would allow us to further clean up the time advance signature by getting rid of x_orig and making x const as it should be.
Related issues
0

Related merge requests
1

    Feature/improved split
    asgard/asgard !83

Graham Lopez @bgl added hygeine label 1 week ago

McDaniel, Benjamin T
McDaniel, Benjamin T @3bm · 4 days ago
Developer

some of this was addressed in !83 (merged).

bmcdanie:
use the same machinery for batch building/execution in time advance and realspace conversion where possible.

clean up batch set interface and enforce RAII if no performance penalty.

closes #85 .

### DIFF
--- a/src/batch.cpp
+++ b/src/batch.cpp
@@ -261,6 +261,11 @@ batch<P, resrc>::batch(int const capacity, int const nrows, int const ncols,
   assert(nrows > 0);
   assert(ncols > 0);
   assert(stride > 0);
+
+  // FIXME removing separate allocation
+  // so, this needs to go along with assert checks
+  // on exec for nullptr
+  // essentially, RAII and this becomes unnecessary
   for (P *&ptr : (*this))
   {
     ptr = nullptr;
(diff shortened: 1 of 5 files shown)


bmcdanie:
replaced by #269

PR #266, opened 2020-03-25
Feature/faster batching

bmcdanie:
noticed batch building scaled poorly with threads.

this is a nested loop process - outer loop over rows in a chunk, inner loop over connected items. we were parallelizing the inner loop, but significant work was also done in the outer loop.

so, two improvements: 1) hoisted the pragma into outer loop to parallelize more work. 
                                     2) ensured realloc wasn't called on operator ptrs.

update: playing with tcmalloc and trying to improve performance of heap allocation across threads failed.

so, allocate all variables inside batch building openmp loop on thread stacks.

this yields multiple x speedup and better scaling.

### DIFF
--- a/contrib/contrib.cmake
+++ b/contrib/contrib.cmake
@@ -141,7 +141,7 @@ if (NOT LINALG_LIBS_FOUND)
       PREFIX contrib/blas/openblas
       URL https://github.com/xianyi/OpenBLAS/archive/v0.3.4.tar.gz
       DOWNLOAD_NO_PROGRESS 1
-      CONFIGURE_COMMAND ""
+      CONFIGURE_COMMAND "USE_OPENMP=1"
       BUILD_COMMAND make
       BUILD_IN_SOURCE 1
       INSTALL_COMMAND make PREFIX=${OpenBLAS_PATH} install
--- a/src/batch.cpp
+++ b/src/batch.cpp
@@ -906,6 +906,8 @@ void unsafe_kronmult_to_batch_sets(P *const *const A, P *const x, P *const y,
 }
 
 // helper for calculating 1d indices for elements
+
+// performant (no-alloc) version
 inline void linearize(fk::vector<int> const &coords, int output[])
 {
   int const output_size = coords.size() / 2;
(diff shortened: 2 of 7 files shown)


bmcdanie:
this is please

stonecoldhughes (approved):

contrib/contrib.cmake
```diff
@@ -141,7 +141,7 @@ if (NOT LINALG_LIBS_FOUND)
       PREFIX contrib/blas/openblas
       URL https://github.com/xianyi/OpenBLAS/archive/v0.3.4.tar.gz
       DOWNLOAD_NO_PROGRESS 1
-      CONFIGURE_COMMAND ""
+      CONFIGURE_COMMAND "USE_OPENMP=1"
```
dlg0, line 144: How is the threading within the BLAS build related to this PR?

src/batch.cpp
```diff
@@ -837,8 +837,8 @@ void kronmult_to_batch_sets(
 
 // unsafe version; uses raw pointers rather than views
 template<typename P>
-void unsafe_kronmult_to_batch_sets(std::vector<P *> const &A, P *const x,
-                                   P *const y, std::vector<P *> const &work,
+void unsafe_kronmult_to_batch_sets(P *const *const A, P *const x, P *const y,
```
dlg0, line 840: the `std::vector` gets allocated on the heap? i.e., why switch from the vector container to a pointer to a pointer?

src/batch.cpp
```diff
@@ -847,7 +847,7 @@ void unsafe_kronmult_to_batch_sets(std::vector<P *> const &A, P *const x,
   int const degree = pde.get_dimensions()[0].get_degree();
 
   // check workspace sizes
-  assert(static_cast<int>(work.size()) == std::min(pde.num_dims - 1, 2));
+  // assert(static_cast<int>(work.size()) == std::min(pde.num_dims - 1, 2));
```
dlg0: what's the deal with this commented assert? we need it or we don't

contrib/contrib.cmake
bmcdanie: we should've been doing it all along: `If the application is parallelized by OpenMP, please build OpenBLAS with USE_OPENMP=1`

from: https://github.com/xianyi/OpenBLAS/wiki/faq.

I changed it in this PR because with one of the compilers I tested, openblas printed a warning message suggesting I enable `USE_OPENMP`

Once all the timers are in, I want to look at how the gemms are scaling to make sure everything in playing nicely WRT threading. we may need to manually set our blas library to use only one thread during the batched gemm loop (which is already parallelized over gemms).

src/batch.cpp
bmcdanie: the calling function `build_batches` was allocating these small vectors on the heap and passing them into this function.

I could not find a way to make these mallocs performant.

So, the options were 1) preallocate all the vectors on the heap, and have each loop calculate an index into its own subvectors, or 2) allocate on thread stack. option 2 seemed more readable/maintainable, so I chose that.

src/batch.cpp
bmcdanie: removed, good catch. it was useful, but switching from vectors to pointers means we cannot assert on this size any longer.

PR #267, opened 2020-03-31
Feature/timer depends on #266

bmcdanie:
add component to wrap and time function calls we want to profile.

generate a report with summary stats on each wrapped function (min/med/max/avg/num calls) at the end of execution.


stonecoldhughes (approved):
looks good, ready for merge

PR #271, opened 2020-04-07
update timer to provide for tic/toc interface

bmcdanie:
our timing method relies on passing a function into the timer w/ its arguments forwarded.

this does not work for constructors, see std: https://eel.is/c++draft/class.ctor#6.

so, extend timer with additional interface - classic tic/toc.


stonecoldhughes (approved):
Clever way of testing it

PR #272, opened 2020-04-09
serial in-place transpose for column major data layout - closes #221

### LINKED ISSUE #221: implement in-place transpose for tensors
current transpose creates a new matrix, transposes values, and then performs a quasi-"move".

this does not work for views (delete is called on the original `data` ptr) and is inefficient.

implement in place transpose (preferable) or implement another version enabled for views.

stonecoldhughes:


src/tensors.hpp
```diff
@@ -2148,29 +2148,69 @@ operator*(matrix<P, omem> const &B) const
...
+     index position of the element in the matrix's transpose */
+  auto remap_index = [this](int const pos) -> int {
+    int row         = pos % nrows_;
+    int col         = pos / nrows_;
+    int destination = row * ncols_ + col;
+    return destination;
+  };
+
+  /* The first and last elements never change position and can be ignored */
+  for (int pos = 1; pos < size() - 1; ++pos)
```
bmcdanie, line 2199: this iteration style will not work for views, whose data is not necessarily contiguous. need to use the matrix iterators instead, or account for stride != nrows.

src/tensors.hpp
```diff
@@ -2148,29 +2148,69 @@ operator*(matrix<P, omem> const &B) const
...
-
-  // inelegant manual "move assignment"
-  // unlike actual move assignment, need to delete the old pointer
-  nrows_  = temp.nrows();
-  ncols_  = temp.ncols();
-  stride_ = nrows();
-  delete[] data_;
-  data_      = temp.data();
-  temp.data_ = nullptr;
+  /* vector pretending to be a matrix */
```
bmcdanie, line 2164: I'm not convinced this will work for views.

src/tensors_tests.cpp
```diff
@@ -3538,3 +3538,50 @@ TEMPLATE_TEST_CASE("fk::matrix device transfer functions", "[tensors]", double,
     REQUIRE(mat_h == gold_view);
   }
 }
+
+TEMPLATE_TEST_CASE("fk::matrix transpose", "[tensors]", double, float)
```
bmcdanie, line 3542: the case that motivated this PR is transposing a view submatrix of a larger matrix - this is what the existing solution would not work for. so, we need tests that cover this case (I think these will expose problems with your implementation).

bmcdanie (changes requested):
need to account for non-contiguous matrices.

src/tensors.hpp
```diff
@@ -2148,29 +2148,69 @@ operator*(matrix<P, omem> const &B) const
...
+  }
+
+  /* spot for each element, true ~ visited, false ~ unvisited */
+  std::vector<bool> visited(size() - 2, false);
+
+  /* Given index "pos" in a linear array interpreted as a matrix of "nrows_"
+     rows, n_cols_ columns, and column-major data layout, return the linear
+     index position of the element in the matrix's transpose */
+  auto remap_index = [this](int const pos) -> int {
+    int row         = pos % nrows_;
```
bmcdanie: this lambda and all its calculated values can be const.

stonecoldhughes: updated

bmcdanie (approved):
good. in the event we need non-contiguous transpose will need to update loop to be stride-aware or use iterators.

PR #273, opened 2020-04-09
remove wrap timer interface

bmcdanie:
now that we have the tic/toc interface, I'm removing the more complex wrap interface to simply the timer.


stonecoldhughes (approved):
even better

PR #275, opened 2020-04-16
Feature/kronmult

bmcdanie:
1) set up Ed's kronmult library as an external project with auto download/build

2) set up workspace for call into these routines

3) integrate into explicit time advance as an alternative to batching


stonecoldhughes (approved):
looks good

PR #288, opened 2020-05-27
Feature/batch with cuda

bmcdanie:
profiling indicates building batches on CPU and transferring to GPU for kronmult is a bottleneck.

fix this by building batches on GPU with a new kernel with cuda is enabled.

add an additional kernel for staging element workspaces on GPU.


stonecoldhughes (approved):
ready for merge

PR #296, opened 2020-06-09
relative test with division by max

bmcdanie:
use relative RMSE test similar to the one in main for tests.


bmcdanie:
test this please

stonecoldhughes (approved):

PR #300, opened 2020-06-12
WIP: Feature/vectorized stage

### LINKED ISSUE #293: vectorize kronmult stage kernel
recent profiling results suggest kronmult stage kernel is the most expensive part of kronmult setup for large problems.

vectorize the load/stores that dominate this kernel to increase performance.

from slack:
essentially what we want to do is reduce our number of loads by 2 (for double) and 4 (for float) by reading vectors of doubles/float rather than single values at a time

in the kernel, each thread copies values one by one:

https://github.com/project-asgard/asgard/blob/7c4dca77275bccafabe339759db92b18729b01c2/src/device/kronmult_cuda.cpp#L50

we'd need to adjust the iteration to occur over pairs of elements (for double) or fours of elements (for float), being aware that the size may not be a multiple of these numbers (so may have to handle stragglers).

in the inside of the loop, instead of the simple copy statement, we use the vector load.

for example:

instead of `a[i] = b[i]`, we have `reinterpret_cast<double2*>(a)[i] = reinterpret_cast<double2*>(b)[i];`

we may need to double check the vector size, but IIRC it's 128 bits. that means we can use double2 and float4.

our pointers must be aligned to do this, but I believe cudaMalloc does this. we can check that with ptr % 128 == 0

bmcdanie:
vectorize staging kernel.

closes #293 .

### DIFF
--- a/src/device/kronmult_cuda.cpp
+++ b/src/device/kronmult_cuda.cpp
@@ -109,6 +109,10 @@ void stage_inputs_kronmult(P const *const x, P *const workspace,
 {
   assert(num_elems > 0);
   assert(num_copies > 0);
+  // num_elems should always have a factor of two in it.
+  // this simplifies vector reads.
+  // let's double check
+  assert(num_elems % 2 == 0);
 
 #ifdef ASGARD_USE_CUDA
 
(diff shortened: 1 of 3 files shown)


bmcdanie:
number 300!

https://www.youtube.com/watch?v=QkWS9PiXekE

PR #301, opened 2020-06-12
pde and unit test inconsistencies

### LINKED ISSUE #270: fokkerplanck_2d_complete inconsistent between matlab & C++
**Describe the bug**
In time_advance-tests, in the test case for the fokkerplanck_2d_complete, the error tolerances are
absurdly high. One possible cause is that the implementation of the PDE in Matlab used for gold data is different than the one in C++.

**To Reproduce**
1. commit 8348b289814bfb72b7b8cf8cd62afba489c98a5b
2. cmake -DASGARD_USE_CUDA=ON -DASGARD_USE_MPI=ON .. ; make -j
3. mpirun -n 2 ./time_advance-tests

**Expected behavior**
The fokkerplanck_2d_complete test will fail if the tolerance is any higher than it currently is.

stonecoldhughes:
fokkerplanck_2 passes time_advance_tests in C++. Several other PDEs are either inconsistent or the gold data generation parameters in Matlab differ from the C++ ones used in tests.

closes #270 .


bmcdanie (changes requested):
please trim all the unnecessary/unrelated files from this PR @stonecoldhughes . there are hundreds of cmake files etc. that should not have been committed, and it's burying the diffs for the files I need to review.

src/time_advance_tests.cpp
```diff
@@ -24,192 +24,97 @@ static distribution_test_init const distrib_test_info;
 static auto constexpr num_steps = 5;
 
 template<typename P>
-void time_advance_test(int const level, int const degree, PDE<P> &pde,
-                       int const num_steps, std::string const filepath,
-                       bool const full_grid = false, P const tol_factor = 1e-10)
+void time_advance_test(parser const &parse, std::string const &filepath,
```
bmcdanie, line 27: why did you change this function?

src/time_advance_tests.cpp
```diff
@@ -223,112 +128,145 @@ void implicit_time_advance_test(int const level, int const degree, PDE<P> &pde,
...
-    int const level      = 4;
-    bool const full_grid = false;
-    auto pde = make_PDE<TestType>(PDE_opts::diffusion_1, level, degree, cfl);
+    int const degree            = 4;
+    int const level             = 4;
     std::string const gold_base = "../testing/generated-inputs/time_advance/"
-                                  "diffusion1/diffusion1_e_sg_l4_d4_t";
+                                  "diffusion1_sg_l4_d4_t";
+    TestType const tol_factor =
+        std::is_same<TestType, double>::value ? 1e-12 : 1e-2;
```
bmcdanie, line 236: is this one looser than it used to be?

.gitignore
```diff
@@ -18,3 +18,4 @@ contrib/lperftools
 contrib/valgrind
 contrib/hdf5
 contrib/highfive
+contrib/kronmult
```
bmcdanie, line 21: thanks

src/pde/pde_diffusion1.hpp
```diff
@@ -21,16 +21,18 @@ class PDE_diffusion_1d : public PDE<P>
   static bool constexpr do_poisson_solve_  = false;
   static bool constexpr has_analytic_soln_ = true;
 
+  static P constexpr nu = 0.01;
+
   static fk::vector<P>
   initial_condition_dim0(fk::vector<P> const &x, P const t = 0)
   {
-    static double p     = -2.0 * PI * PI;
+    static double p     = -2.0 * nu * nu;
```
bmcdanie: could be constified.

src/pde/pde_diffusion1.hpp
```diff
@@ -93,21 +87,21 @@ class PDE_diffusion_1d : public PDE<P>
...
+      return coefficient * std::cos(nu * x_value);
     });
 
     return fx;
   }
 
   static P source_0_t(P const t)
   {
-    static double neg_two_pi_squared = -2.0 * PI * PI;
+    static double coefficient = -2.0 * nu * nu;
```
bmcdanie: const

src/pde/pde_fokkerplanck2_complete.hpp
```diff
@@ -73,16 +74,16 @@ class PDE_fokkerplanck_2d_complete : public PDE<P>
   static auto constexpr psi = [](P x) {
     auto dphi_dx = 2.0 / std::sqrt(M_PI) * std::exp(-std::pow(x, 2));
```
bmcdanie: constify

src/pde/pde_fokkerplanck2_complete.hpp
```diff
@@ -114,13 +115,51 @@ class PDE_fokkerplanck_2d_complete : public PDE<P>
...
-    std::transform(x.begin(), x.end(), ret.begin(), [](auto const elem) {
-      P const a = 2;
-      return 2.0 / (std::sqrt(M_PI) * std::pow(a, 3)) *
-             std::exp(-std::pow(elem, 2) / std::pow(a, 2));
-    });
+    P N = 1000.0;
+    P h = 20.0 / N;
+    P Q = 0;
+
+    auto function = [](fk::vector<P> const &p) -> fk::vector<P> {
```
bmcdanie: const

src/time_advance_tests.cpp
```diff
@@ -499,151 +480,182 @@ TEMPLATE_TEST_CASE("time advance - fokkerplanck_1d_4p1a", "[time_advance]",
...
+  SECTION("fokkerplanck_2d_complete, level 5, degree 3, sparse grid")
   {
-    int const degree     = 3;
-    int const level      = 3;
-    bool const full_grid = false;
-    auto pde = make_PDE<TestType>(PDE_opts::diffusion_1, level, degree, cfl);
+    int const level  = 5;
+    int const degree = 3;
+
+    TestType const tol_factor =
```
bmcdanie, line 542: good

src/time_advance_tests.cpp
```diff
@@ -499,151 +480,182 @@ TEMPLATE_TEST_CASE("time advance - fokkerplanck_1d_4p1a", "[time_advance]",
...
+
+    time_advance_test(parse, gold_base, tol_factor);
   }
 }
 
+/* Explicit time advance is not a fruitful approach to this problem */
 TEMPLATE_TEST_CASE("time advance - fokkerplanck_2d_complete", "[time_advance]",
                    float, double)
 {
   /* FIXME - these tolerances are way too high. Different parameters are likely
```
bmcdanie: take this comment out, it's fixed

bmcdanie (changes requested):
looks good overall. let's tighten up the const in a couple of areas (these were not const prior but we can fix now). also, let's discuss briefly why the test function change. not opposed, just want to know why we switched.

src/time_advance_tests.cpp
stonecoldhughes: All the arguments to the original version of the function are now encapsulated in the parser class, so it is more concise this way. With this configuration, the body of the unit tests are only responsible for defining state via a parser object and then this routine builds the PDE from it. Previously, there were 2 separate test functions for time_advance - for explicit and implicit. Now the unified version infers which should be called from the parser object.

bmcdanie: ok, that's an improvement.

src/time_advance_tests.cpp
stonecoldhughes: Yes - after updating the pde to match the reference implementation, the tolerances had to be loosened.

bmcdanie: ok - did we see any of that in other component tests for this pde? we're loosening here, are we loosening for this PDE in coefficient generation, source vectors, boundary conditions?

src/pde/pde_diffusion1.hpp
stonecoldhughes: done

src/pde/pde_diffusion1.hpp
stonecoldhughes: done

src/pde/pde_fokkerplanck2_complete.hpp
stonecoldhughes: done

src/pde/pde_fokkerplanck2_complete.hpp
stonecoldhughes: done

src/time_advance_tests.cpp
stonecoldhughes: done

src/pde/pde_diffusion1.hpp
```diff
@@ -21,16 +21,18 @@ class PDE_diffusion_1d : public PDE<P>
...
   static bool constexpr has_analytic_soln_ = true;
 
+  static P constexpr nu = 0.01;
+
   static fk::vector<P>
   initial_condition_dim0(fk::vector<P> const &x, P const t = 0)
   {
-    static double p     = -2.0 * PI * PI;
-    P const coefficient = std::exp(p * t);
+    static const double p = -2.0 * nu * nu;
```
bmcdanie: west const! you wound me

bmcdanie (approved):

src/pde/pde_diffusion1.hpp
stonecoldhughes: lol fixed

src/time_advance_tests.cpp
bmcdanie: checked - the answer is no.

PR #314, opened 2020-06-30
WIP: Feature/faster stage

bmcdanie:
improvements so far:

1) more threads.
2) remove for loop in kernel - tile blocks over problem such that only one pass is required.


bmcdanie:
redundant, closed - optimized kernel during hackathon

PR #318, opened 2020-07-10
Send Asgard real space solution vector to a shared matlab session

stonecoldhughes:


stonecoldhughes:
Requires a shared Matlab session to already be running. A running session is not shared by default - it must be made shared by typing "matlab.engine.shareEngine" in the Matlab command prompt.

dlg0:
I get the following error when trying to build using ...
```
cmake ../ -DASGARD_IO_MATLAB_DIR=/Applications/MATLAB_R2020a.app
```

```
-- kronmult library not found - dl and build from src
-- build without CUDA support
-- Found Matlab: /Applications/MATLAB_R2020a.app/extern/include (found version "9.8")
-- Found OpenMP_CXX: -fopenmp (found version "4.5")
-- Found OpenMP: TRUE (found version "4.5")
CMake Error at CMakeLists.txt:225 (target_include_directories):
  Cannot specify include directories for target "io" which is not built by
  this project.


CMake Error at CMakeLists.txt:226 (target_link_libraries):
  Cannot specify link libraries for target "io" which is not built by this
  project.
```

bmcdanie:
being implemented by @ckendrick .

PR #319, opened 2020-07-13
point to new kronmult lib with less syncthreads

bmcdanie:
during the hackathon, we discovered that we could hoist syncthreads out of kgemm functions and into the transition between dimensions (i.e. between `kronmult[x]` and `kronmult[x-1]` functions. point to the new version of the kronmult library that does this.


bmcdanie:
added more to kronmult lib - consolidating in one PR

PR #320, opened 2020-07-13
Feature/stage launch (depends on #319)

bmcdanie:
Stage kronmults kernel is not launching with nearly enough threads to saturate GPU bandwidth for the 6d case. Set `num_blocks` = `num_total_copies/threads_per_block` (with rounding guard) to drastically improve performance for high dimension problems.


bmcdanie:
replaced by #325

PR #322, opened 2020-07-13
use new kronmult lib that doesn't block over m and n

bmcdanie:
removing the blocking loops over m and n have decreased our integer instruction mix and shown good performance during hackathon. point to version of kronmult lib without m/n blocking.


bmcdanie:
note to self: unbreak the non-cuda implementation

bmcdanie:
this method worked for our benchmark problem at first, but scaled poorly as we increased degree. additionally, when using one-warp blocks, this method is slower.

PR #339, opened 2020-08-25
non-uniform level

bmcdanie:
add support for non-uniform number of basis levels across dimensions.


bmcdanie:
this is essentially complete - the only failing tests are `fokkerplank_2d_complete`'s `coefficients` and (to a lesser degree) `time_advance` tests.

at this point, the features I've added to the C++ to enable performant adaptivity have created divergence between C++/MATLAB that makes debugging this issue difficult.

so, I'm placing this on hold while I work to add these features to the MATLAB and bring the code paths closer together to enable debugging. concurrently, I'm going to begin adding adaptivity to the C++.

PR #353, opened 2020-12-04
Set default CMAKE_BUILD_TYPE and populate list of options

quantumsteve:

## Proposed changes

Also, I noticed that CMAKE_BUILD_TYPE doesn't have a default or a list of values in the cmake gui. Do we want to use this setting? What defaults do we want?

I had to add the `<unordered_map>` header to  `src/elements.hpp` in order to successfully build on MacOS with clang/libc++.

- [x] Bugfix
- [ ] New feature
- [x] Build related changes

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,10 +17,26 @@ endif ()
 set (CMAKE_CXX_STANDARD 17)
 set (CMAKE_CXX_STANDARD_REQUIRED ON)
 
-message (STATUS "CMAKE_BUILD_TYPE is ${CMAKE_BUILD_TYPE}")
+# Set a default build type if none was specified
+# https://blog.kitware.com/cmake-and-the-default-build-type/
+set (default_build_type "Release")
+if (EXISTS "${CMAKE_SOURCE_DIR}/.git")
+  set (default_build_type "Debug")
+endif ()
+
+if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+  message (STATUS "Setting build type to '${default_build_type}' as none was specified.")
+  set (CMAKE_BUILD_TYPE "${default_build_type}" CACHE
+       STRING "Choose the type of build." FORCE)
+  # Set the possible values of build type for cmake-gui
+  set_property (CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
+    "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
+else ()
+  message (STATUS "CMAKE_BUILD_TYPE is ${CMAKE_BUILD_TYPE}")
+endif ()
 
 # add compiler flags we always want to use
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic -O3 -g")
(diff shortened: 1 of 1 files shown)


hbrunie:
Hi Steve,
Why removing -O3 and -g from CXX compile flags in the CMakeLists.txt?

bmcdanie:
Thanks @quantumsteve. We had considered build types earlier and decided at that time to always build O3 g. But, we need -DNDEBUG anyway for good peformance on GPU (asserts expensive there), so this might be a good time to add these.  Release/-DNDEBUG builds might throw a lot of warnings for unused variables we prepare for asserts...we may need to add a layer there if we care about compiler spam in Release.

@hbrunie, this may be because the build types carry default flags that may cover these (e.g. -03 is default in Release, -g is part of Debug defaults). We may need to check whether these defaults are platform dependent.

bmcdanie:
ah and I need to add you to the CI whitelist to run these tests.

bmcdanie:
whitelisted - please test with `git commit --allow-empty -m "Trigger Build"` if you would - we can squash it out after

@dlg0 can we get them cloudwatch log perms on AWS?

dlg0:
Y. People - slack me your preferred AWS email address. 

Also - I don't want compiler spam!

bmcdanie:
> Also - I don't want compiler spam!

yyyyyeah...I'm not a huge fan either. I think we want Debug to be the default build type anyway, but we should probably add an abstraction layer over `assert` so that we don't get unused variable warnings.

something like 

```
inline void asgard_assert(bool const condition) {
  (void)condition;
  assert(condition);
}
```

bmcdanie:
@quantumsteve can you push an empty commit to run the CI? also, if you have time to add the `asgard_assert` layer so that we don't get a bunch of warnings in release, that'd be good. otherwise I can add this one as is and open another issue.

dlg0:
Yeah, let's make sure we don't introduce compiler warnings. As i recall - compiler warnings were at one point a cause of failure for CI tests, is that still the case?

bmcdanie:
@dlg0 no, I didn't carry that one over to AWS, but I can add it back in if we want that. develop is still warning-free.

dlg0:
I think that the clang formatting and the compiler warnings are good to have at the CI stage - the cost for both is very low if we start from a place where we have formatted code and no warnings (which we do now), but the cost for both is high if we let those slip and want to bring them back later. So yes, I'd prefer to keep them enabled.

bmcdanie:
great, adaptivity should be wrapped up soon, I'll bring up jenkins :grimacing: and check how I implemented that

bmcdanie:
after #362 is merged, we should rebase this and merge it in. #362 prevents a load of compiler warnings when build type is set to `Release`.

quantumsteve:
@bmcdanie I merged upstream/develop into this branch.

bmcdanie:
not to be a bother @quantumsteve but could you roll that last one back and rebase onto develop instead? I want to keep the history consistent, and we have previously always done rebases vs. merge develop into development branch.

bmcdanie (approved):
excellent. thank you @quantumsteve !

PR #355, opened 2020-12-08
Pr kronsubmodule

hbrunie:
## Proposed changes

Change kronmult external library as a submodule of Asgard.

- [ ] Bugfix
- [ ] New feature
- [x] Build related changes

- [ ] Yes
- [x] No

Cori GPU
## Checklist

- [x ] this PR is up to date with current the current state of 'develop'

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "contrib/kronmult"]
+	path = contrib/kronmult
+	url = https://github.com/project-asgard/kronmult.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,8 +105,7 @@ if(ASGARD_USE_CUDA)
 endif()
 
 # build component to interface with Ed's kronmult lib
-##TODO: insert the interface directly inside kronmult dir
-## and manage dependance with cmake targets and interface
+##TODO: link to kronmult as interface library
 add_library(kronmult_cuda SHARED src/device/kronmult_cuda.cpp)
 if(ASGARD_USE_CUDA)
   set_source_files_properties( src/device/kronmult_cuda.cpp PROPERTIES LANGUAGE CUDA ) # no .cu extension
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,14 +105,16 @@ if(ASGARD_USE_CUDA)
 endif()
 
 # build component to interface with Ed's kronmult lib
+##TODO: insert the interface directly inside kronmult dir
+## and manage dependance with cmake targets and interface
 add_library(kronmult_cuda SHARED src/device/kronmult_cuda.cpp)
 if(ASGARD_USE_CUDA)
   set_source_files_properties( src/device/kronmult_cuda.cpp PROPERTIES LANGUAGE CUDA ) # no .cu extension
   set_target_properties( kronmult_cuda PROPERTIES CUDA_ARCHITECTURES OFF)
   set_target_properties( kronmult_cuda PROPERTIES COMPILE_FLAGS "-arch sm_70 -g -lineinfo --ptxas-options=-O3")
-  set_target_properties( kronmult_cuda PROPERTIES LINK_FLAGS "-Wl,-rpath,${KRON_PATH}")
+  set_target_properties( kronmult_cuda PROPERTIES LINK_FLAGS "-Wl,-rpath,${CMAKE_BINARY_DIR}")
 endif()
-target_include_directories (kronmult_cuda PRIVATE ${KRON_PATH} ${CMAKE_BINARY_DIR})
+target_include_directories (kronmult_cuda PRIVATE ${CMAKE_SOURCE_DIR}/contrib/kronmult/ ${CMAKE_BINARY_DIR})
 
 if(ASGARD_USE_MKL)
   if(ASGARD_USE_CUDA)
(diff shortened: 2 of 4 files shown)


hbrunie:
The tree of commit in the PR is a bit of a mess... As the PR should be pretty simple, I suggest we **squash** and merge them all into just one commit.

bmcdanie:
@hbrunie that sounds good to me.

CMakeLists.txt
```diff
@@ -105,14 +105,16 @@ if(ASGARD_USE_CUDA)
 endif()
 
 # build component to interface with Ed's kronmult lib
+##TODO: insert the interface directly inside kronmult dir
```
bmcdanie: can you help me understand this comment?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "contrib/kronmult"]
+	path = contrib/kronmult
+	url = https://github.com/project-asgard/kronmult.git
```
bmcdanie, line 3: is this always going to pull the latest commit of master for kronmult?

contrib/contrib.cmake
```diff
@@ -184,37 +184,36 @@ endif ()
...
+                    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
 
-endif()
+    set(KRON_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/contrib/kronmult/
+        CACHE PATH "kronmult include directory")
 
+    # also install it
+    install(DIRECTORY ${KRON_INCLUDE_DIR}/kronmult DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+
+    # for convenience setup a target
```
bmcdanie: also, can you help me understand these commands? thanks.

bmcdanie (changes requested):
Hugo - looks good, thank you! if you could help me understand a bit some of the comments, and whether this setup will always pull latest commit from master as written, we can get this in.

.gitmodules
hbrunie: My understanding is that it will not. 
To update submodule I found this commands: 
```
git pull --recurse-submodules
git submodule update --remote --recursive
```
There are other ways.

CMakeLists.txt
hbrunie: This comment is very unclear, and we should probably delete it, or rewrite it.
I am not enough familiar with CMAKE to know the best way to implement the link between kronmult and asgard. So I just made something that works. But I think there is another way to implement this using CMAKE interface feature.

.gitmodules
bmcdanie: ok - I do not want that behavior (automatically pulling latest), because I don't want development changes there to implicitly ripple into ASGarD.

rather, I'd prefer we set the gitlink to the latest release. I'm not super familiar with that process, but looks like it can be done this way? https://stackoverflow.com/questions/43671101/how-to-add-a-specific-release-of-a-repo-as-a-submodule

contrib/contrib.cmake
hbrunie: That is linked to the comment I made above, you made a comment on this.
I don't understand exactly how INTERFACE works in CMAKE, but I understood it was one way of dealing with dependences. 
I built the kronmult_interface (CMAKE INTERFACE) to be able to use it in the asgard CMakeLists.txt directly instead of linking "by hand" the library. But I gave up to finish it properly.

So: I can either continue reading on it and trying to make it using CMAKE INTERFACE (changing Asgard CMakeLists.txt) or we can delete these lines.

CMakeLists.txt
bmcdanie: I think this way is good for me - we do have a `PRIVATE` kronmult_cuda dependency on kron

bmcdanie: I've only used the `INTERFACE` to link something that needs to export symbols, but not consume them itself. you may mean something else though - I'm thinking you're referring to linking as `INTERFACE` rather than `PRIVATE`

contrib/contrib.cmake
bmcdanie: ah I can help some with this (as far as I know)

`INTERFACE` flag when linking means the target does not itself need the symbols linked in, but things that link to the target do.

in our case, we need `PRIVATE`, because `kronmult_cuda` does need the symbols defined in `kron`. if `kronmult_cuda` did not need the symbols, but things that link to `kronmult_cuda` did, that would be `INTERFACE`. `PUBLIC` would mean both `kronmult_cuda` and things linking to `kronmult_cuda` need those symbols.

bmcdanie: ah but you're talking about `INTERFACE` as it relates to add library

bmcdanie: so this is different, and yes because it's a header only library, we could use that.

hbrunie: Ok. If you look at line 223 of CMakeLists.txt in this PR, `PRIVATE` is indeed used.

.gitmodules
bmcdanie: after reading up on this, I see the gitlink will be fixed to a particular commit in the submodule repo unless we alter it, so I am good with this.

bmcdanie (approved):

bmcdanie (approved):
thanks.

PR #372, opened 2021-01-21
fix time in analytic solution generation.

bmcdanie:
also handle nonseparable (time+space) sources

## Proposed changes
1) fix time analytic solutions are generated with.
2) fix source generation to handle nonseparable sources.

- [x] Bugfix
- [ ] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -169,7 +169,7 @@ int main(int argc, char **argv)
   for (auto i = 0; i < opts.num_time_steps; ++i)
   {
     // take a time advance step
-    auto const time          = i * pde->get_dt();
+    auto const time          = (i + 1) * pde->get_dt();
     auto const update_system = i == 0;
     auto const method = opts.use_implicit_stepping ? time_advance::method::imp
                                                    : time_advance::method::exp;
@@ -185,14 +185,16 @@ int main(int argc, char **argv)
     // print root mean squared error from analytic solution
     if (pde->has_analytic_soln)
     {
+      // get analytic solution at time(step+1)
       auto const subgrid           = adaptive_grid.get_subgrid(get_rank());
+      auto const time_multiplier   = pde->exact_time(time + pde->get_dt());
       auto const analytic_solution = transform_and_combine_dimensions(
           *pde, pde->exact_vector_funcs, adaptive_grid.get_table(), transformer,
-          subgrid.col_start, subgrid.col_stop, degree);
-      auto const time_multiplier     = pde->exact_time((i + 1) * pde->get_dt());
-      auto const analytic_solution_t = analytic_solution * time_multiplier;
-      auto const diff                = f_val - analytic_solution_t;
-      auto const RMSE                = [&diff]() {
+          subgrid.col_start, subgrid.col_stop, degree, time, time_multiplier);
+
+      // calculate root mean squared error
+      auto const diff = f_val - analytic_solution;
+      auto const RMSE = [&diff]() {
         fk::vector<prec> squared(diff);
         std::transform(squared.begin(), squared.end(), squared.begin(),
                        [](prec const &elem) { return elem * elem; });
@@ -200,11 +202,11 @@ int main(int argc, char **argv)
                           squared.size();
         return std::sqrt(mean);
       }();
-      auto const relative_error = RMSE / inf_norm(analytic_solution_t) * 100;
+      auto const relative_error = RMSE / inf_norm(analytic_solution) * 100;
       auto const [rmse_errors, relative_errors] =
           gather_errors(RMSE, relative_error);
       tools::expect(rmse_errors.size() == relative_errors.size());
-      for (int i = 0; i < rmse_errors.size(); ++i)
+      for (auto i = 0; i < rmse_errors.size(); ++i)
       {
         node_out() << "Errors for local rank: " << i << '\n';
         node_out() << "RMSE (numeric-analytic) [wavelet]: " << rmse_errors(i)
--- a/src/time_advance.cpp
+++ b/src/time_advance.cpp
@@ -12,40 +12,25 @@
 namespace time_advance
 {
 template<typename P>
-static std::vector<fk::vector<P>>
+static fk::vector<P>
 get_sources(PDE<P> const &pde, adapt::distributed_grid<P> const &grid,
-            basis::wavelet_transform<P, resource::host> const &transformer)
+            basis::wavelet_transform<P, resource::host> const &transformer,
+            P const time)
 {
-  std::vector<fk::vector<P>> source_vects;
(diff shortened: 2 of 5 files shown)


src/time_advance.cpp
```diff
@@ -371,38 +353,35 @@ template fk::vector<float> adaptive_advance(
     options const &program_opts, fk::vector<float> const &x, float const time,
     int const workspace_size_MB, bool const update_system);
 
-template fk::vector<double>
```
hbrunie, line 374: Why this tabulation change? Is this mandatory with clang-format?

src/time_advance.cpp
```diff
@@ -12,40 +12,25 @@
...
-            basis::wavelet_transform<P, resource::host> const &transformer)
+            basis::wavelet_transform<P, resource::host> const &transformer,
+            P const time)
 {
-  std::vector<fk::vector<P>> source_vects;
   auto const my_subgrid = grid.get_subgrid(get_rank());
   // FIXME assume uniform degree
   auto const degree = pde.get_dimensions()[0].get_degree();
+  auto const dof    = std::pow(degree, pde.num_dims) * my_subgrid.nrows();
+  fk::vector<P> sources(dof);
```
hbrunie, line 24: Could you explain me why get_sources only returns a vector, and not a vector of vectors anymore?
How the previous information has been refactored?

src/main.cpp
```diff
@@ -185,26 +185,28 @@ int main(int argc, char **argv)
     // print root mean squared error from analytic solution
     if (pde->has_analytic_soln)
     {
+      // get analytic solution at time(step+1)
       auto const subgrid           = adaptive_grid.get_subgrid(get_rank());
+      auto const time_multiplier   = pde->exact_time(time + pde->get_dt());
       auto const analytic_solution = transform_and_combine_dimensions(
           *pde, pde->exact_vector_funcs, adaptive_grid.get_table(), transformer,
-          subgrid.col_start, subgrid.col_stop, degree);
-      auto const time_multiplier     = pde->exact_time((i + 1) * pde->get_dt());
```
hbrunie, line 192: Time was already incremented previously, so this is not bug origin, right?

hbrunie (commented):
It is not clear to me where the time bug in analytic solution generation is coming from. Could you pinpoint?

src/time_advance.cpp
```diff
@@ -60,16 +45,15 @@ adaptive_advance(method const step_method, PDE<P> &pde,
...
+    auto const my_subgrid     = adaptive_grid.get_subgrid(get_rank());
+    auto const unscaled_parts = boundary_conditions::make_unscaled_bc_parts(
         pde, adaptive_grid.get_table(), transformer, my_subgrid.row_start,
         my_subgrid.row_stop);
     return (step_method == method::exp)
-               ? explicit_advance(pde, adaptive_grid, program_opts,
-                                  unscaled_sources, unscaled_parts, x_orig,
-                                  workspace_size_MB, time)
-               : implicit_advance(pde, adaptive_grid, unscaled_sources,
+               ? explicit_advance(pde, adaptive_grid, transformer, program_opts,
```
hbrunie, line 53: Is there a reason why `time_advance` can not directly call `boundary_conditions::make_unscaled_bc_parts` ?
Because we are already passing `pde, adaptive_grid, transformer` to `time_advance`. 
So it seems like all the code above (computing `my_subgrid` and `unscaled_parts`) could be encapsulated/embedded inside `time_advance`.

src/time_advance.hpp
```diff
@@ -31,8 +31,8 @@ template<typename P>
 fk::vector<P>
 explicit_advance(PDE<P> const &pde,
                  adapt::distributed_grid<P> const &adaptive_grid,
+                 basis::wavelet_transform<P, resource::host> const &transformer,
```
hbrunie, line 34: If we do need transformer, we can encapsulate more code in explicit/implicit_time_advance, I made a comment on `time_advance.cpp`. What do you think?

hbrunie (commented):
I would need some explanation to understand how is the bug fixed. 
Right now, I don't understand the origin of the bug nor how it has been fixed.
This is due to my understanding of the code (before the PR), this is not due to the code changes in this PR.

src/time_advance.cpp
bmcdanie: this is correct. this was done previously because we were keeping everything in the time advance on GPU, so it was advantageous to precompute the source vectors and boundary conditions and score them on the GPU, scaling them later for time. this is only possible for time and space separable sources/boundary conditions. the physicists have since thought up nonseparable source vectors :). And I imagine they will eventually with the boundary conditions as well. So, this technique no longer works. Moreover, it's not even really necessary with our current setup, where much of time advance outside of kronmult is on CPU.

src/time_advance.cpp
bmcdanie: yes

src/time_advance.hpp
bmcdanie: this is correct, and I'd like to do that for boundary conditions as well. I have done some of it with sources.

src/time_advance.cpp
bmcdanie: previously, we created `num_sources` many unscaled source vectors, and combined/scaled them later for time into a single source vector. however, this is not possible in some of the new pdes (see `separable_adapt_test_2d.m`). since I already needed to modify these functions to address the time bug, I abandoned the unscaled vectors + scaling and just regenerate source vector as needed.

src/main.cpp
bmcdanie: correct, it was just that the time and time multiplier were not correctly passed into `transform_and_combine_dimensions`

src/time_advance.cpp
hbrunie: https://github.com/project-asgard/DG-SparseGrid/blob/20448e62cfb796cbc50ba721cc04821105301e56/PDE/seperable_adapt_test_2d.m#L73

So first we built unscaled source vectors (same with boundaries and analytic solution), and then combined them into a single source vector (using time multiplier to scale them correctly). But this is not always possible, so now we decided to abandoned the unscaled vectors and generate source, boundary conditions and analytic solution vectors depending on time (separable).

src/time_advance.cpp
hbrunie: Thus, will you commit a change to bring the computation of `my_subgrid` and `unscaled_parts` inside `explicit_advance` and `implicit_advance`?

src/time_advance.hpp
hbrunie: Do you want to make it in this PR, or should I open an issue with this (hardening) ?

hbrunie (approved):
Thank you for this fix! 👍

PR #392, opened 2021-05-04
Add option to use scalapack p?gesv and p?getrs

quantumsteve:

## Proposed changes

This PR adds the `ASGARD_USE_SCALAPACK` build option and "scalapack" solver option to ASGarD. The build option is only available when `ASGARD_USE_MPI=ON`. If built with `ASGARD_USE_SCALAPACK=OFF`, the "scalapack" solver option will print an error and exit.

With this option, matrices are distributed in a block-cyclic row distribution. The c functions initializing this grid are called in the in the `cblacs_grid` class. This class contains member functions for obtaining the row and column indices of a particular process as well as memory that needs to be allocated on that process. Generally, it is constructed once,  stored as a shared_ptr in `scalapack_matrix_info` and `scalapack_vector_info`, and otherwise ignored by the client code.

`scalapack_matrix_info` contains the additional scalapack information associated with  `fk::matrix<P>`. I initially tried including it as a member function, but changed my mind because of the additional complexity it created. The constructor taking two ints is for matrices on rank 0 that are not distributed and `get_desc()` returns a pointer to a 9-element array that can be used for a scatter operation to distribute this on multiple processes.  The second constructor also includes the row group size mb, column group size nb and `cblacs_grid` object described above.

https://www.netlib.org/scalapack/slug/node75.html 

`scalapack_vector_info` is a decorator containing `scalapack_matrix_info` with the number of columns is 1. 

`parallel_solver` contains `gather_matrix` and `scatter_matrix` functions to redistribute a matrix from 1 to multiple processes and back again. This is currently only used by tests, but may be useful if we want to change row and column group sizes in the matrix generated by the implicit time advance.

In fast_math, we add `gesv` and `getrs` functions that also take fk::scalapack_matrix_info and fk::scalapack_vector_info. They use this information to call the parallel solvers p?gesv and p?getrs.

In batch.hpp, a build_system_matrix function was added that only build the portion needed by that particular process. 

in time_advance.cpp, `implicit_advance` was updated to work with the scalapack solver. Block sizes are set to match information in `element_subgrid`.

ASGarD stores vectors a row with many columns, while scalapack expects a column with many rows. ASGarD's distributed vectors contains multiple copies of the same data, while scalapack only has data on some processes.  Functions `col_to_row_major` and `row_to_col_major` work around these differences. 

In time_advance_tests.cpp, I added an add the scalapack solver to all existing implicit tests. Scalapack tests are skipped when  get_num_ranks() return 1.

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

## Checklist

- [x ] this PR is up to date with current the current state of 'develop'


CMakeLists.txt
```diff
@@ -81,7 +81,8 @@ option (ASGARD_IO_HIGHFIVE "Use the HighFive HDF5 header library for I/O" OFF)
 option (ASGARD_USE_CUDA "Optional CUDA support for asgard" OFF)
 option (ASGARD_USE_OPENMP "Optional openMP support for asgard" ON)
 option (ASGARD_USE_MPI "Optional distributed computing support for asgard" OFF)
-option (ASGARD_USE_SLATE "Use optional slate linear algebra library" OFF)
+include(CMakeDependentOption)
+cmake_dependent_option(ASGARD_USE_SCALAPACK "Use optional scalapack linear algebra library" OFF "ASGARD_USE_MPI" OFF)
```
cianciosa, line 85: Since scalapack requires mpi why not just force `ASGARD_USE_MPI` on when `ASGARD_USE_SCALAPACK` is on?

src/time_advance.cpp
```diff
@@ -239,6 +243,87 @@ explicit_advance(PDE<P> const &pde,
...
+  x.resize(size);
+
+  int const num_subgrid_cols = get_num_subgrid_cols(get_num_ranks());
+  for (int row_rank = 1; row_rank < num_subgrid_cols; ++row_rank)
+  {
+    int col_rank = row_rank * num_subgrid_cols;
+    if (get_rank() == row_rank)
+    {
+      MPI_Send(x_orig.data(), x_orig.size(), mpi_type, col_rank, 0,
+               MPI_COMM_WORLD);
```
cianciosa: Should we assume `MPI_COMM_WORLD` internally? A future use case may intend to run multiple parallel instances of asgard. A pattern, I've used before was to define an internal communicator like`asgard_comm_world`. The end user could the set that to `MPI_COMM_WORLD` or a subset of it from `MPI_Comm_split`.

quantumsteve: I moved these functions into distribution.cpp where the global communicator is defined.

CMakeLists.txt
quantumsteve: We also have to protect against the case where `ASGARD_USE_MPI` is turned off after the initial configuration. I think it's better to use the CMake module instead of doing it ourselves.

I don't like that `ASGARD_USE_SCALAPACK` doesn't initially show up in cmake-gui. What do you think about defaulting `ASGARD_USE_MPI` to `ON`?

ckendrick: I think it would be nice to allow setting `ASGARD_USE_SCALAPACK` to automatically turn on MPI, or print a message saying to configure with MPI enabled as well.

It is a bit misleading to me that running `cmake -DASGARD_USE_SCALAPACK=ON` will configure fine, but it will not actually enable scalapack since MPI is disabled. I think this case should be handled somehow (warning saying scalapack wasn't actually enabled, or turn on MPI) otherwise people could do this and expect scalapack to be enabled.

CMakeLists.txt
```diff
@@ -283,9 +293,13 @@ if (ASGARD_IO_HIGHFIVE)
   target_link_libraries (io PUBLIC highfive tensors PRIVATE hdf5)
 endif ()
 
-if (ASGARD_USE_SLATE)
-  target_link_libraries (lib_dispatch PRIVATE slate_lapack_api)
-  target_compile_definitions (lib_dispatch PUBLIC ASGARD_USE_SLATE)
+if (ASGARD_USE_SCALAPACK)
+  find_package(scalapack REQUIRED)
```
ckendrick, line 295: When lapack is not installed, I got a lot of linking errors. I assumed that finding scalapack would look for lapack as well. Should a check be added to make sure lapack is installed?

ckendrick (commented):
Running asgard with `mpirun -n 2 ./asgard -i -s scalapack` is crashing with this error:
```
*** The MPI_Comm_size() function was called before MPI_INIT was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
```
Am I running it correctly?

CMakeLists.txt
quantumsteve: What OS are you using and how did you install scalapack? On Ubuntu, libblas and liblapack are automatically installed with scalapack.
https://packages.debian.org/sid/libscalapack-openmpi2.1

ckendrick: I am using Ubuntu 20.04 and installed `libscalapack-openmpi2.1` and `libscalapack-openmpi-dev`.
CMake configured without any errors, but I got this linking message for all targets: `No rule to make target '/usr/lib/x86_64-linux-gnu/liblapack.so'`. It looks like LAPACK_LIB was not found by CMake.

After installing `liblapack-dev` these problems were fixed.

It looks like `libscalapack-openmpi2.1` installs `liblapack3` which provides `liblapack.so.3` but no link is being made to `liblapack.so`?

quantumsteve: Yes, it looks like `/usr/lib/cmake/scalapack-2.1.0.openmpi/scalapack-targets-none.cmake` points to `liblapack.so` instead of `liblapack.so.3`. I'll add that to my [existing bug report about the file](https://bugs.launchpad.net/ubuntu/+source/scalapack/+bug/1917534).

quantumsteve: I added a comment to our existing scalapack bug report.

https://bugs.launchpad.net/ubuntu/+source/scalapack/+bug/1917534

CMakeLists.txt
quantumsteve: Changing the behavior of CmakeDependentOption is proving much more difficult than I expected. In out last meeting we agreed to leave it as is for now.

I added some documentation to the README.md and [wiki page](https://github.com/project-asgard/asgard/wiki/platforms) to minimize installation issues.

ckendrick (approved):

Issue #373, opened 2021-01-21
extend test coverage in time_advance-tests


bmcdanie:
currently, we test only that the wavelet space solution for each problem at a given timestep matches the MATLAB solution for that problem/timestep. we could fairly simply extend these tests to also test that the difference between wavelet space solution and analytic solution (for problems with analytic solution) also matches. this will ensure we have consistent analytic solutions generated at every timestep.

hbrunie:
Are these tests extended to double and single precision ? (are we testing ASGARD_DOUBLE_PRECISION OFF)?

bmcdanie:
yes - that macro is only used to control precision in main, but time advance tests are templated to run in single and double precision.

mkstoyanov:
New API tests use the L-2 norm (not just the wavelet error) as well as point-wise values. New PDEs have much more tests included.

One can always use more tests, but this is OK for now.

PR #394, opened 2021-05-26
Feature/advect blob

hbrunie:
Adding some PDE test cases to highlight adaptivity usefulness. 

## Proposed changes
Adding advect blob 1 2 3 and 4. Inspired from David commit on DG-SparseGrid

- [X] No

Cori Haswell and GPU Nvidia V100 and laptop INTEL i5 8th generation

## Todo list
- [x] Build related changes: Adding pde to ./asgard -a (helper)
- [ ] Documentation content changes: Adding advect_blob PDE description

### DIFF
--- a/src/pde.hpp
+++ b/src/pde.hpp
@@ -12,6 +12,10 @@
 #include "program_options.hpp"
 #include "quadrature.hpp"
 
+#include "pde/pde_advect_blob1.hpp"
+#include "pde/pde_advect_blob2.hpp"
+#include "pde/pde_advect_blob3.hpp"
+#include "pde/pde_advect_blob4.hpp"
 #include "pde/pde_base.hpp"
 #include "pde/pde_continuity1.hpp"
 #include "pde/pde_continuity2.hpp"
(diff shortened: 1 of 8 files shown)


hbrunie:
No need for tests based on golden result from matlab. (discussion with @dlg0 )
Only test that the RMSE between simulated and analytical solution does not exceed arbitrary threshold.
What should be this threshold?

dlg0:
something that will allow the test to be fast, so a relative error of 1e-3 or so is fine, running at lev=3, deg=3 or so. Just double check that if you increase lev by 1 that the error goes down by about a factor of (deg+1)

hbrunie:
> something that will allow the test to be fast, so a relative error of 1e-3 or so is fine, running at lev=3, deg=3 or so. Just double check that if you increase lev by 1 that the error goes down by about a factor of (deg+1)

Looking at the tests `pde_tests.cpp` I can not find any test on "equality" between exact and simulated solution.
For every PDEs, the exact solution is compared to the exact golden results from DG-SparseGrid. 
But the C++ simulated solution is never compared to the C++ exact solution.

This is already highlighted in the issue:  #373.

dlg0:
That's something I did not know. We can certainly fix that - although it's not like it is really "wrong", but I'd prefer it to be testing against the C++ generated exact solution. So for your new one, just test against the C++ exact solution.

hbrunie:
There is another PR adding the tests between C++ simulated solution and analytic one (#402 ).
Once the #402  PR is accepted and merged, I will rebase this one #394  on top of it and add the corresponding tests for advect_blob.

mkstoyanov:
Stalled PR, feel free to resubmit based on the latest developments.

Issue #387, opened 2021-03-09
Compiling errors Apple Clang


cianciosa:
**Describe the bug**
Asgard fails to compile using the Apple clang with 12.4.

#### First Bug ####
```
Scanning dependencies of target basis
[ 27%] Building CXX object CMakeFiles/basis.dir/src/basis.cpp.o
In file included from /Users/m4c/Projects/asgard/src/basis.cpp:2:
In file included from /Users/m4c/Projects/asgard/src/distribution.hpp:3:
/Users/m4c/Projects/asgard/src/elements.hpp:115:8: error: no template named 'unordered_map' in namespace 'std'; did you mean 'unordered_set'?
  std::unordered_map<int64_t, fk::vector<int>> id_to_coords_;
  ~~~~~^~~~~~~~~~~~~
       unordered_set
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/unordered_set:382:28: note: 'unordered_set' declared here
class _LIBCPP_TEMPLATE_VIS unordered_set
                           ^
In file included from /Users/m4c/Projects/asgard/src/basis.cpp:2:
In file included from /Users/m4c/Projects/asgard/src/distribution.hpp:3:
/Users/m4c/Projects/asgard/src/elements.hpp:75:26: error: no member named 'at' in 'std::__1::unordered_set<long long, fk::vector<int, mem_type::owner,
      resource::host>, std::__1::equal_to<long long>, std::__1::allocator<long long> >'
    return id_to_coords_.at(active_element_ids_[index]);
           ~~~~~~~~~~~~~ ^
2 errors generated.
make[2]: *** [CMakeFiles/basis.dir/src/basis.cpp.o] Error 1
make[1]: *** [CMakeFiles/basis.dir/all] Error 2
make: *** [all] Error 2
```
##### Fix #####
The line
https://github.com/project-asgard/asgard/blob/d43b3bdaf64db8385d5035f558acad7683ad59cf/src/elements.hpp#L10
Should be replaced with 
`#include <unordered_map>`

#### Second bug ####
```
Scanning dependencies of target adapt
[ 47%] Building CXX object CMakeFiles/adapt.dir/src/adapt.cpp.o
In file included from /Users/m4c/Projects/asgard/src/adapt.cpp:1:
In file included from /Users/m4c/Projects/asgard/src/adapt.hpp:2:
[...]

PR #398, opened 2021-07-13
Fix const error when compiling in clang.

cianciosa:
## Proposed changes

This fixes the second bug in https://github.com/project-asgard/asgard/issues/387. The bug arrises because `plan_` contains elements that are marked `const` and the line `plan_ = new_plan` tries to overwrite them which would violate the `const` property. This pr works around that by deleting the old elements of `plan_` first then inserting the elements of `new_plan`

- [x] Bugfix
- [ ] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Mac OS and Linux. Running the unit tests locally result in several test failures of unknown origin.

## Checklist

### DIFF
--- a/src/adapt.cpp
+++ b/src/adapt.cpp
@@ -312,7 +312,8 @@ fk::vector<P> distributed_grid<P>::refine_elements(
   auto const new_plan = get_plan(get_num_ranks(), table_);
   auto const remapper = remap_for_addtl(table_.size() - added);
   auto const y        = redistribute_vector(x, plan_, new_plan, remapper);
-  plan_               = new_plan;
+  plan_.clear();
+  plan_.insert(new_plan.cbegin(), new_plan.cend());
 
   return y;
 }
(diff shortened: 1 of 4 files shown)


src/adapt.cpp
```diff
@@ -312,7 +312,8 @@ fk::vector<P> distributed_grid<P>::refine_elements(
   auto const new_plan = get_plan(get_num_ranks(), table_);
   auto const remapper = remap_for_addtl(table_.size() - added);
   auto const y        = redistribute_vector(x, plan_, new_plan, remapper);
-  plan_               = new_plan;
```
quantumsteve, line 315: Please add a comment about why we're doing this.

src/adapt.cpp
```diff
@@ -338,7 +339,9 @@ fk::vector<P> distributed_grid<P>::remove_elements(
   auto const new_plan = get_plan(get_num_ranks(), table_);
   auto const remapper = remap_for_delete(all_remove_indices, table_.size());
   auto const y        = redistribute_vector(x, plan_, new_plan, remapper);
-  plan_               = new_plan;
+
```
quantumsteve, line 346: Please add a comment about why we're doing this.

quantumsteve (changes requested):
Please add a code comment and squash your commits.

cianciosa:
🤬🤬🤬 Every time I try to squash the commits, it just keeps adding more.

quantumsteve (approved):

PR #414, opened 2022-01-25
Add new matlab changes

ckendrick:

## Proposed changes

This starts to add the new [changes made to the Matlab repo ](https://github.com/project-asgard/DG-SparseGrid/commit/d6b2073a2aa4062d3d69a8f9e9950873c62df9a3) into Asgard.

Major changes include:
- New coefficient type was added: `coefficient_type::div`.
- Adds a `moment_dV` gfunc to the `dimension` class.
- Adds a gfunc to the `partial_term` class for specifying the lhs mass term.
- All PDE definitions were changed to be written with new additions of div, dV, etc.
- Partial term coefficients are precomputed and stored up a max level. With adaptivity enabled, this is the same max level as specified with the `-m` option. Without adaptivity, this max level defaults to the largest level found across all dimensions. This variable is added to the `PDE` class as `max_level`.
- New `--terms` option added to enable/disable individual terms of a PDE at runtime.
- Dimension mass matrices are computed with the `generate_dimension_mass_mat` function, which should get called before using `generate_all_coefficients`.
- Other cases for the `fokkerplanck_2d_complete` PDE were added and are named `fokkerplanck_2d_complete_case[1-4]`.
- MATLAB interface:
  - The initial condition vector is copied to the Matlab workspace.
  - A representation of the PDE is copied to Matlab as a struct containing the terms, pterms, dimensions, and other data fields.
  - Plotting error with adaptivity enabled was fixed.

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [ ] No

## Checklist


src/pde/pde_base.hpp
```diff
@@ -104,8 +108,14 @@ class dimension
     degree_ = degree;
   }
 
+  void set_mass_matrix(fk::matrix<P> const &new_mass)
+  {
+    this->mass_.clear_and_resize(new_mass.nrows(), new_mass.ncols()) = new_mass;
```
quantumsteve, line 114: Do we have to call `clear_and_resize` before the copy assignment operator?

ckendrick: I believe so since the matrix copy assignment operator does not handle resizing and `mass_` is initially empty. I am copying how [the coefficients are being set](https://github.com/project-asgard/asgard/blob/8baa1e6c379a2f8925c50b56181c702b5b729237/src/pde/pde_base.hpp#L197-L198), but let me know if there is a better way to do this.

src/pde/pde_base.hpp
```diff
@@ -33,6 +33,8 @@ using vector_func = std::function<fk::vector<P>(fk::vector<P> const, P const)>;
 template<typename P>
 using scalar_func = std::function<P(P const)>;
 
+using g_func_type = std::function<double(double const, double const)>;
```
quantumsteve:
```suggestion
using g_func_type = std::function<P(P const, P const)>;
```

quantumsteve:
You have a precision issue on MacOS
```
% ctest -R pde-test --verbose
UpdateCTestConfiguration  from :/Users/svh/Documents/asgard/build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/Users/svh/Documents/asgard/build/DartConfiguration.tcl
Test project /Users/svh/Documents/asgard/build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 12
    Start 12: pde-test

12: Test command: /Users/svh/Documents/asgard/build/pde-tests
12: Environment variables: 
12:  OMP_NUM_THREADS=2
12: Test timeout computed to be: 10000000
12: 
12: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
12: pde-tests is a Catch v3.0.0-preview.4 host application.
12: Run with -? for options
12: 
12: Randomness seeded to: 1515874252
12: 
12: -------------------------------------------------------------------------------
12: testing fokkerplanck2_complete_case4 implementations - double
12:   fp2 complete pterm funcs
12: -------------------------------------------------------------------------------
12: /Users/svh/Documents/asgard/src/pde_tests.cpp:296
12: ...............................................................................
12: 
12: /Users/svh/Documents/asgard/testing/tests_general.hpp:41: FAILED:
12:   REQUIRE( (diff_norm / max) < (tolerance * std::sqrt(v0.size())) )
12: with expansion:
12:   0.00000000000000593632
12:   <
12:   0.00000000000000496507
12: 
12: ===============================================================================
12: test cases:  14 |  13 passed | 1 failed
12: assertions: 281 | 280 passed | 1 failed
12: 
1/1 Test #12: pde-test .........................***Failed   13.46 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) =  13.46 sec

The following tests FAILED:
	 12 - pde-test (Failed)
Errors while running CTest
Output from these tests are in: /Users/svh/Documents/asgard/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
```

src/boundary_conditions.cpp
```diff
@@ -186,7 +191,7 @@ fk::vector<P> boundary_conditions::compute_left_boundary_condition(
 
   P const scale_factor = (1.0 / std::sqrt(domain_per_cell)) *
                          bc_func(fk::vector<P>({domain_min}), time)(0) * g *
-                         -1.0;
+                         dV * -1.0;
```
quantumsteve: consider using std::negate
https://en.cppreference.com/w/cpp/utility/functional/negate

src/boundary_conditions.cpp
```diff
@@ -169,11 +171,14 @@ fk::vector<P> boundary_conditions::compute_left_boundary_condition(
 
   fk::vector<P> bc(dof);
 
-  P g = g_func(domain_min, time);
+  P g  = g_func(domain_min, time);
+  P dV = dv_func(domain_min, time);
+  expect(std::isfinite(dV));
   if (!std::isfinite(g))
```
quantumsteve, line 176: I think we want either the `expect` or this if statement. If we keep the if statement, there should be an `else` that throws an exception.

src/boundary_conditions.cpp
```diff
@@ -254,51 +262,82 @@ std::vector<fk::vector<P>> boundary_conditions::generate_partial_bcs(
...
+       ++dim_num)
   {
-    partial_bc_vecs.emplace_back(forward_transform(
-        dimensions[dim_num], bc_funcs[dim_num], transformer, time));
+    auto const degrees_freedom_1d_other =
+        dimensions[dim_num].get_degree() *
+        fm::two_raised_to(dimensions[dim_num].get_level());
+    auto const &f = bc_funcs[dim_num];
+    auto const &g = terms[dim_num].get_partial_terms()[p_index].g_func;
+    vector_func<P> const bc_func = [f, g](fk::vector<P> const x, P const t) {
```
quantumsteve: May we make it explicit that x is only a view and doesn't allocate anything?

src/boundary_conditions.cpp
```diff
@@ -254,51 +262,82 @@ std::vector<fk::vector<P>> boundary_conditions::generate_partial_bcs(
...
-  if (p_index > 0)
-  {
-    // FIXME assume uniform level across dimensions
-    auto const degrees_freedom_1d =
-        dimensions[d_index].get_degree() *
-        fm::two_raised_to(dimensions[d_index].get_level());
-    fk::matrix<P> chain = eye<P>(degrees_freedom_1d, degrees_freedom_1d);
+  partial_bc_vecs.at(d_index) = std::move(trace_bc);
+  partial_bc_vecs.at(d_index) = transformer.apply(
+      partial_bc_vecs.at(d_index), dimensions[d_index].get_level(),
```
quantumsteve: do we need the slower bounds-checking function here?

src/coefficients.cpp
```diff
@@ -304,81 +362,86 @@ fk::matrix<P> generate_coefficients(
...
       }
     }
   }
 
+  if (pterm.coeff_type == coefficient_type::grad)
+  {
+    // take the negative transpose of div
+    coefficients.transpose();
+    std::transform(coefficients.begin(), coefficients.end(),
+                   coefficients.begin(), [&](P const elem) { return -elem; });
```
quantumsteve: Use std::negate instead of lambda?

src/main.cpp
```diff
@@ -184,6 +192,18 @@ int main(int argc, char **argv)
   ml_plot.init_plotting(*pde, adaptive_grid.get_table());
   ml_plot.plot_fval(*pde, adaptive_grid.get_table(), real_space,
                     analytic_solution_realspace);
+
+  // send initial condition to matlab
+  std::vector<size_t> sizes(pde->num_dims);
+  for (int i = 0; i < pde->num_dims; i++)
+  {
+    sizes[i] = pde->get_dimensions()[i].get_degree() *
+               std::pow(2, pde->get_dimensions()[i].get_level());
```
quantumsteve: https://github.com/project-asgard/asgard/blob/develop/src/fast_math.hpp#L15

src/matlab_plot.cpp
```diff
@@ -316,6 +330,159 @@ void matlab_plot::plot_fval(PDE<P> const &pde, elements::table const &table,
...
+  matlab::data::CellArray ml_terms = factory_.createCellArray(
+      {static_cast<size_t>(pde.num_dims), static_cast<size_t>(pde.num_terms)});
+  for (auto i = 0; i < pde.num_dims; ++i)
+  {
+    for (auto j = 0; j < pde.num_terms; ++j)
+    {
+      ml_terms[i][j] = make_term(pde.get_terms()[j][i], pde.max_level);
+    }
+  }
+  ml_pde[0]["terms"] = ml_terms;
```
quantumsteve:
```suggestion
  ml_pde[0]["terms"] = std::move(ml_terms);
```

src/pde/pde_base.hpp
```diff
@@ -153,53 +163,130 @@ class partial_term
...
+    {
+      auto const dof = deg * fm::two_raised_to(level);
+      fk::matrix<P> result(dof, dof);
+      auto mass_tmp = mass_.extract_submatrix(0, 0, dof, dof);
+
+      fm::gemm(mass_tmp.invert(),
+               fk::matrix<P, mem_type::const_view>(new_coefficients, 0, dof - 1,
+                                                   0, dof - 1),
+               result);
+      coefficients_.emplace_back(result);
```
quantumsteve:
```suggestion
      coefficients_.push_back(std::move(result));
```

src/pde/pde_base.hpp
```diff
@@ -380,6 +486,20 @@ class PDE
...
 
+    auto const num_active_terms = cli_input.get_active_terms().size();
+    if (num_active_terms != 0)
+    {
+      auto const active_terms = cli_input.get_active_terms();
+      for (auto i = num_terms_ - 1; i >= 0; --i)
+      {
+        if (active_terms(i) == 0)
+        {
+          terms_.erase(terms_.begin() + i);
```
quantumsteve, line 497: Use the [erase-remove idiom](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom)

src/transformations.hpp
```diff
@@ -123,17 +131,11 @@ fk::vector<P> forward_transform(
...
-                       return static_cast<P>(1e-12);
-                     }
-                     return static_cast<P>(1e-4);
-                   }();
-                   return std::abs(elem) < compare ? static_cast<P>(0.0) : elem;
-                 });
+  std::transform(
+      transformed.begin(), transformed.end(), transformed.begin(), [](P &elem) {
+        P const compare = [] { return std::numeric_limits<P>::epsilon(); }();
+        return std::abs(elem) < compare ? static_cast<P>(0.0) : elem;
```
quantumsteve:
```suggestion
        return std::abs(elem) < std::numeric_limits<P>::epsilon() ? static_cast<P>(0.0) : elem;
```

src/transformations.hpp
```diff
@@ -123,17 +131,11 @@ fk::vector<P> forward_transform(
...
-                     {
-                       return static_cast<P>(1e-12);
-                     }
-                     return static_cast<P>(1e-4);
-                   }();
-                   return std::abs(elem) < compare ? static_cast<P>(0.0) : elem;
-                 });
+  std::transform(
+      transformed.begin(), transformed.end(), transformed.begin(), [](P &elem) {
+        P const compare = [] { return std::numeric_limits<P>::epsilon(); }();
```
quantumsteve:
```suggestion

```

quantumsteve:
@ckendrick Would you rebase against upstream/develop so we can add the additional MacOS MPI test. I'm getting some warnings on my local machine and want to ensure they go away with a more recent version on AppleClang.

```
[87/90] Building CXX object CMakeFiles/kronmult-tests.dir/src/kronmult_tests.cpp.o
In file included from /Users/svh/Documents/asgard/src/kronmult_tests.cpp:2:
In file included from /Users/svh/Documents/asgard/src/coefficients.cpp:10:
In file included from /Users/svh/Documents/asgard/src/transformations.hpp:4:
/Users/svh/Documents/asgard/src/distribution.hpp:7:32: warning: unknown warning group '-Wcast-function-type', ignored [-Wunknown-warning-option]
#pragma GCC diagnostic ignored "-Wcast-function-type"
                               ^
1 warning generated.
```

src/pde/pde_base.hpp
ckendrick: I might be missing how to do this, but I could not get a version to work using `remove`/`remove_if` since `terms_` is a `vector<vector<term<P>>>` and we want to remove when the corresponding index of `active_terms == 0`. There didn't seem to be an easy way to tie the elements of these vectors together.

src/coefficients.cpp
```diff
@@ -30,15 +30,65 @@ void generate_all_coefficients(
...
+        for (int level = 0; level <= pde.max_level; ++level)
+        {
+          auto const dof = dim.get_degree() * fm::two_raised_to(level);
+          fk::matrix<P> result(dof, dof);
+          auto mass_tmp = mass_coeff.extract_submatrix(0, 0, dof, dof);
+
+          auto pterm_coeff = generate_coefficients<P>(
+              dim, term_1D, partial_terms[k], transformer, level, time, rotate);
+
+          fm::gemm(mass_tmp.invert(), pterm_coeff, result);
```
cianciosa, line 57: Is there a reason to do a matrix invert + a matrix matrix multiply instead of just doing a single matrix solve?

src/coefficients_tests.cpp
```diff
@@ -318,4 +319,48 @@ TEMPLATE_TEST_CASE("fokkerplanck2_complete terms", "[coefficients]", double,
...
+    TestType const time = 1.0;
+    generate_dimension_mass_mat(*pde, transformer);
+    generate_all_coefficients(*pde, transformer, time, true);
+
+    int row = 0;
+    for (auto i = 0; i < pde->num_dims; ++i)
+    {
+      for (auto j = 0; j < pde->num_terms; ++j)
+      {
+        auto const &term_1D       = pde->get_terms()[j][i];
```
cianciosa, line 347: Since the [I] index of get terms in continuous in memory, it's would be better for cache usage to swap the two for loops.

src/coefficients_tests.cpp
```diff
@@ -318,4 +319,48 @@ TEMPLATE_TEST_CASE("fokkerplanck2_complete terms", "[coefficients]", double,
...
+        auto const &partial_terms = term_1D.get_partial_terms();
+        for (auto k = 0; k < static_cast<int>(partial_terms.size()); ++k)
+        {
+          int const dof = degree * fm::two_raised_to(levels(i));
```
cianciosa, line 351: Since dof has no dependence on the k index and degree is const, this can be moved outside of the loop.

src/matlab_plot.cpp
```diff
@@ -316,6 +330,159 @@ void matlab_plot::plot_fval(PDE<P> const &pde, elements::table const &table,
...
+    }
+    else if (type == boundary_condition::neumann)
+    {
+      return "N";
+    }
+    else
+    {
+      return " ";
+    }
+  };
```
cianciosa, line 433: This could also be coded as a switch statement

```
switch(type)
{
  case boundary_condition::periodic: return "P";
  case boundary_condition::dirichlet: return "D";
  case boundary_condition::neumann: return "N";
}
```
There's no need for a default case since type has to be one of the cases in the boundary condition enum.

Alternatively, you could store the return character in the enum value itself. In pde_base.hpp you could do
```
enum class boundary_condition : char
{
  periodic = 'P',
  dirichlet = 'D',
  neumann = 'N'
};
```
Then the lambda could be simplified to
```
auto const get_bc_str = [](boundary_condition const &type) -> std::string {
     return type;
   };
```
or eliminated all together.

src/pde/pde_base.hpp
```diff
@@ -153,53 +163,130 @@ class partial_term
...
+    for (int level = 0; level <= max_level; ++level)
+    {
+      auto const dof = deg * fm::two_raised_to(level);
+      fk::matrix<P> result(dof, dof);
+      auto mass_tmp = mass_.extract_submatrix(0, 0, dof, dof);
+
+      fm::gemm(mass_tmp.invert(),
+               fk::matrix<P, mem_type::const_view>(new_coefficients, 0, dof - 1,
+                                                   0, dof - 1),
+               result);
```
cianciosa, line 238: Can this be done as a linear solve directly instead of an invert and matrix matrix multiply?

src/pde_tests.cpp
```diff
@@ -292,4 +292,47 @@ TEMPLATE_TEST_CASE("testing fokkerplanck2_complete implementations", "[pde]",
...
+        read_matrix_from_txt_file(base_dir + "gfuncs.dat"));
+    auto const gold_dvs = fk::matrix<TestType>(
+        read_matrix_from_txt_file(base_dir + "dvfuncs.dat"));
+
+    int row = 0;
+    for (auto i = 0; i < pde->num_dims; ++i)
+    {
+      for (auto j = 0; j < pde->num_terms; ++j)
+      {
+        auto const &term_1D       = pde->get_terms()[j][i];
```
cianciosa, line 308: The pde->get_terms()[j][I] is continuous in memory so switching the order of the loops should access memory better.

cianciosa (commented):
I've added some suggestions I don't believe they are critical for merging though.

src/coefficients_tests.cpp
ckendrick: I agree, although changing this would also require changing how `row` is being calculated to index into the gold data file (or changing the Matlab script to generate in a different order). Since this test is only run once, and there are only 12 total iterations, I don't think this is necessary.

src/pde/pde_base.hpp
ckendrick: I think you are correct, this could probably be replaced with a single linear solve call. When looking into this, I noticed [our `gesv` implementation](https://github.com/project-asgard/asgard/blob/develop/src/fast_math.hpp#L190-L221) seems to only support a matrix and vector, rather than two matrices as input. This might need to be extended to support two matrices in order to make this replacement?

quantumsteve (approved):
At the Dev meeting we agreed to merge this PR as is.

PR #426, opened 2022-07-08
documented physics of 1d continuity equation

jburz2001:
## Proposed changes

Provides useful comments to serve as physics background for understanding the 1-dimensional continuity equation.

- [ ] Bugfix
- [ ] New feature
- [ ] Build related changes
- [x] Documentation content changes

- [ ] Yes
- [x] No

Linux (Fedora 35)

## Checklist

- [x] documentation has been added (if appropriate)

### DIFF
--- a/src/pde/pde_continuity1.hpp
+++ b/src/pde/pde_continuity1.hpp
@@ -26,28 +26,33 @@
 // This relates the change in f over time to the change in f over space.
 // More specifically, this equation says that the amount that f decreases by
 // over time is equivalent to the amount of f that escapes from the region.
-// In other words, f decreases over time at some location if f diverges from that location (i.e., if there is a source there).
-// In this case, the divergence, df/dx, is 1-dimensional.
-// Equivalently, one can negate both sides to write the equation in terms of the convergence, -df/dx:
+// In other words, f decreases over time at some location if f diverges from
+// that location (i.e., if there is a source there). In this case, the
+// divergence, df/dx, is 1-dimensional. Equivalently, one can negate both sides
+// to write the equation in terms of the convergence, -df/dx:
 //      df/dt = -df/dx.
-// This says that f increases over time at a point if f converges to that point (i.e., if there is a sink there).
+// This says that f increases over time at a point if f converges to that point
+// (i.e., if there is a sink there).
 //
-// In a slightly more complex sense, let's consider the continuity equation that is explicitly evaluated at a specific point, x',
-// on some 1-dimensional grid:
+// In a slightly more complex sense, let's consider the continuity equation that
+// is explicitly evaluated at a specific point, x', on some 1-dimensional grid:
 //      -df(x')/dt = df(x')/dx.
-// This says that when our code takes a discrete time step, dt, then the value of f at x=x', f(x'), changes by df(x').
-// Importantly, in this equation, -df(x')/dt = df(x')/dx, which implies that the df(x') with respect to time 
-// is exactly compensated by a -df(x') with respect to space.
-// The Continuity Equation is thus an example of a conservation law. The amount of f(x') that changes from one point of time 
-// to another does not disappear, the value is simply displaced to a location x'+dx.
-// In the case of a fluid (e.g. water) that's flowing, this means that the amount of mass at some location will change over time
-// since the fluid is displaced over the neighboring region.
+// This says that when our code takes a discrete time step, dt, then the value
+// of f at x=x', f(x'), changes by df(x'). Importantly, in this equation,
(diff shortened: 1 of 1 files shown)


quantumsteve:
I added your GitHub user id to our AWS CodeBuild filter group. Would you please push an empty commit?
@dlg0 Please give Justin access to view the CI output.

jburz2001:
Pardon the delay @quantumsteve . I just pushed an empty commit to the physics-documentation branch. Is this okay or should I instead push the empty commit to a different branch?

quantumsteve:
Below is the output from the clang-format-12 diff. 

I added your user id to `asgard-unit-clang`, so that should build next time. 

`asgard-unit-cuda-gxx` shows "Internal Service Error: CodeBuild is experiencing issues"

```
[Container] 2022/07/18 16:11:10 Running command bash -c "diff  -u <(cat src/pde/*pp) <(clang-format-12 src/pde/*pp)"
26 | --- /dev/fd/63  2022-07-18 16:11:10.962529994 +0000
27 | +++ /dev/fd/62  2022-07-18 16:11:10.962529994 +0000
28 | @@ -772,28 +772,33 @@
29 | // This relates the change in f over time to the change in f over space.
30 | // More specifically, this equation says that the amount that f decreases by
31 | // over time is equivalent to the amount of f that escapes from the region.
32 | -// In other words, f decreases over time at some location if f diverges from that location (i.e., if there is a source there).
33 | -// In this case, the divergence, df/dx, is 1-dimensional.
34 | -// Equivalently, one can negate both sides to write the equation in terms of the convergence, -df/dx:
35 | +// In other words, f decreases over time at some location if f diverges from
36 | +// that location (i.e., if there is a source there). In this case, the
37 | +// divergence, df/dx, is 1-dimensional. Equivalently, one can negate both sides
38 | +// to write the equation in terms of the convergence, -df/dx:
39 | //      df/dt = -df/dx.
40 | -// This says that f increases over time at a point if f converges to that point (i.e., if there is a sink there).
41 | +// This says that f increases over time at a point if f converges to that point
42 | +// (i.e., if there is a sink there).
43 | //
44 | -// In a slightly more complex sense, let's consider the continuity equation that is explicitly evaluated at a specific point, x',
45 | -// on some 1-dimensional grid:
46 | +// In a slightly more complex sense, let's consider the continuity equation that
47 | +// is explicitly evaluated at a specific point, x', on some 1-dimensional grid:
48 | //      -df(x')/dt = df(x')/dx.
49 | -// This says that when our code takes a discrete time step, dt, then the value of f at x=x', f(x'), changes by df(x').
50 | -// Importantly, in this equation, -df(x')/dt = df(x')/dx, which implies that the df(x') with respect to time
51 | -// is exactly compensated by a -df(x') with respect to space.
52 | -// The Continuity Equation is thus an example of a conservation law. The amount of f(x') that changes from one point of time
53 | -// to another does not disappear, the value is simply displaced to a location x'+dx.
54 | -// In the case of a fluid (e.g. water) that's flowing, this means that the amount of mass at some location will change over time
55 | -// since the fluid is displaced over the neighboring region.
56 | +// This says that when our code takes a discrete time step, dt, then the value
57 | +// of f at x=x', f(x'), changes by df(x'). Importantly, in this equation,
58 | +// -df(x')/dt = df(x')/dx, which implies that the df(x') with respect to time is
59 | +// exactly compensated by a -df(x') with respect to space. The Continuity
60 | +// Equation is thus an example of a conservation law. The amount of f(x') that
61 | +// changes from one point of time to another does not disappear, the value is
62 | +// simply displaced to a location x'+dx. In the case of a fluid (e.g. water)
63 | +// that's flowing, this means that the amount of mass at some location will
64 | +// change over time since the fluid is displaced over the neighboring region.
65 | //
66 | -// Note that f can represent various physical quantities and f can be either a vector or scalar field.
67 | -// f can be be mass density (a scalar) in the case of hydrodynamics;
68 | -// it can be charge density (a scalar) in the case of electrodynamics;
69 | -// and it can even be the electromagnetic energy density (a scalar) that is stored in electric and magnetic fields,
70 | -// as in the case of electromagnetism (see Poynting's Theorem).
71 | +// Note that f can represent various physical quantities and f can be either a
72 | +// vector or scalar field. f can be be mass density (a scalar) in the case of
73 | +// hydrodynamics; it can be charge density (a scalar) in the case of
74 | +// electrodynamics; and it can even be the electromagnetic energy density (a
75 | +// scalar) that is stored in electric and magnetic fields, as in the case of
76 | +// electromagnetism (see Poynting's Theorem).
77 | //
78 | // ---------------------------------------------------------------------------
79 | template<typename P>
80

<br class="Apple-interchange-newline">
```

quantumsteve (approved):

PR #463, opened 2022-11-07
Moment class and vlasov PDE

ckendrick:
Please review the [developer documentation](https://github.com/project-asgard/asgard/wiki/developing)
on the wiki of this project that contains help and requirements.

## Proposed changes

This PR adds the following changes:
- Adds a new `moment` class
- `vlasov_lb_full_f` PDE
- Multiple functions can now be specified for the initial conditions in each dimension. Fixes #446.
- The `PDE` class now holds a vector of `moment`s. These can be optionally provided in the PDE spec (see the vlasov PDE for specifying moments in a PDE) 

## What type(s) of changes does this code introduce?
_Put an `x` in the boxes that apply._

- [ ] Bugfix
- [x] New feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

### Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Checklist

_Put an x in the boxes that apply. You can also fill these out after creating
the PR. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- [x] this PR is up to date with current the current state of 'develop'
- [ ] code added or changed in the PR has been clang-formatted
- [x] this PR adds tests to cover any new code, or to catch a bug that is being fixed
- [ ] documentation has been added (if appropriate)

PR #474, opened 2022-12-09
Field class

mkstoyanov:
Please review the [developer documentation](https://github.com/project-asgard/asgard/wiki/developing)
on the wiki of this project that contains help and requirements.

## Proposed changes

Describe what this PR changes and why.  If it closes an issue, link to it here
with [a supported keyword](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

## What type(s) of changes does this code introduce?
_Put an `x` in the boxes that apply._

- [ ] Bugfix
- [X] New feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

### Does this introduce a breaking change?

- [ ] Yes
- [X] No

## Checklist

_Put an x in the boxes that apply. You can also fill these out after creating
the PR. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- [X] this PR is up to date with current the current state of 'develop'
- [ ] code added or changed in the PR has been clang-formatted
- [ ] this PR adds tests to cover any new code, or to catch a bug that is being fixed
- [ ] documentation has been added (if appropriate)

PR #475, opened 2022-12-13
IMEX time advance and PDE parameters

### LINKED ISSUE #445: params input in C++
Current MATLAB branch has a parameters "p" input in each 1D function "@(x,p,t)" where p is a parameter struct that gives user defined quantities for the run.  The parameters implementation is hacked to allow operators to change in time as the solution changes.

**Additional context**
Would need an "update_params_each_timestep" option implemented as well.  See https://github.com/project-asgard/DG-SparseGrid/blob/cory_branch/asgard_run_pde.m#L539-L587 for more information.  

The goal would to be depreciate this feature once the PDE_system work is finalized, but this is needed to get some results out soon.

ckendrick:

## Proposed changes

This finishes the implementation of the IMEX time advance started in #463 for the vlasov PDE, and adds a way to set/use simple parameters in PDEs.

An example of setting PDE parameters can be found [here](https://github.com/ckendrick/asgard/blob/5ada4529a64ef90dda3bafc6005f2a356d4543a4/src/pde/pde_vlasov_lb_full_f.hpp#L22-L24), using a parameter inside a gfunc can be found [here](https://github.com/ckendrick/asgard/blob/5ada4529a64ef90dda3bafc6005f2a356d4543a4/src/pde/pde_vlasov_lb_full_f.hpp#L285-L290), and updating a parameter can be found [here](https://github.com/ckendrick/asgard/blob/5ada4529a64ef90dda3bafc6005f2a356d4543a4/src/time_advance.cpp#L514-L518).

IMEX time advance is added using the `-x` switch and currently works only with the `vlasov` PDE. An example using IMEX: `./asgard -p vlasov -d 3 -l "4 4" -t 0.0015 -f -x`

Closes #445.

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist


src/time_advance.cpp
```diff
@@ -57,9 +58,14 @@ adaptive_advance(method const step_method, PDE<P> &pde,
...
-                                  unscaled_parts, x_orig, time,
-                                  program_opts.solver, update_system);
+               : ((step_method == method::imp)
+                      ? implicit_advance(pde, adaptive_grid, transformer,
+                                         unscaled_parts, x_orig, time,
+                                         program_opts.solver, update_system)
+                      : imex_advance(pde, adaptive_grid, transformer,
+                                     program_opts, unscaled_parts, x_orig,
+                                     workspace_size_MB, time,
+                                     program_opts.solver, update_system));
```
quantumsteve: Nested ternary operators are hard to read. I think an if/else tree or a switch statement would be clearer.

src/time_advance.cpp
```diff
@@ -83,8 +89,13 @@ adaptive_advance(method const step_method, PDE<P> &pde,
...
                                unscaled_parts, y, workspace_size_MB, time)
-            : implicit_advance(pde, adaptive_grid, transformer, unscaled_parts,
-                               y, time, program_opts.solver, refining);
+            : ((step_method == method::imp)
+                   ? implicit_advance(pde, adaptive_grid, transformer,
+                                      unscaled_parts, y, time,
+                                      program_opts.solver, refining)
+                   : imex_advance(pde, adaptive_grid, transformer, program_opts,
+                                  unscaled_parts, y, workspace_size_MB, time,
+                                  program_opts.solver, refining));
```
quantumsteve: Nested ternary operators are hard to read. I think an if/else tree or a switch statement would be clearer.

src/time_advance.cpp
```diff
@@ -392,48 +403,228 @@ implicit_advance(PDE<P> const &pde,
...
     first_time = false;
 
     A.clear_and_resize(A_local_rows, A_local_cols);
     build_system_matrix(pde, table, A, grid);
+    // AA = I - dt*A;
+    for (int i = 0; i < A.nrows(); ++i)
+    {
+      for (int j = 0; j < A.ncols(); ++j)
+      {
+        A(i, j) *= -dt;
```
quantumsteve: I thin that for column major order, these two loops should be interchanged.

src/time_advance.cpp
```diff
@@ -392,48 +403,228 @@ implicit_advance(PDE<P> const &pde,
...
+  // function in x
+  std::cout << " generating moment reduced matrices\n";
+  for (auto &m : pde.moments)
+  {
+    m.createMomentReducedMatrix(pde, adaptive_grid.get_table());
+  }
+
+  for (auto m : pde.moments)
+  {
+    expect(m.get_moment_matrix().nrows() > 0);
```
quantumsteve: It would be better to explicitly remove the loop from Release builds instead of assuming it'll be optimized out.

I think you can just add the `expect(...)` line to the previous loop. Otherwise, check `NDEBUG` instead of using `expect(...)`.

src/time_advance.cpp
```diff
@@ -392,48 +403,228 @@ implicit_advance(PDE<P> const &pde,
...
+
+  tools::timer.start("kronmult_setup");
+  fx = kronmult::execute(pde, table, program_opts, grid, workspace_size_MB,
+                         f2_x);
+  tools::timer.stop(apply_id);
+  reduce_results(fx, reduced_fx, plan, get_rank());
+
+  fk::vector<P> f_3s(x_orig.size());
+  exchange_results(reduced_fx, f_3s, elem_size, plan, get_rank());
+  fm::axpy(f_3s, x, (P)0.5 * dt); // f0 + 0.5*dt*apply_A(f0+f_2)
```
```suggestion
  fm::axpy(f_3s, x, static_cast<P>(0.5) * dt); // f0 + 0.5*dt*apply_A(f0+f_2)
```

src/time_advance.cpp
```diff
@@ -392,48 +403,228 @@ implicit_advance(PDE<P> const &pde,
...
+
+  tools::timer.start("kronmult_setup");
+  fx =
+      kronmult::execute(pde, table, program_opts, grid, workspace_size_MB, f_2);
+  tools::timer.stop(apply_id);
+  reduce_results(fx, reduced_fx, plan, get_rank());
+
+  exchange_results(reduced_fx, f_3s, elem_size, plan, get_rank());
+  fm::axpy(f_3s, x, (P)0.5 * dt);
```
```suggestion
  fm::axpy(f_3s, x, static_cast<P>(0.5) * dt);
```

src/transformations.cpp
```diff
@@ -128,28 +130,75 @@ std::vector<fk::matrix<P>> gen_realspace_transform(
...
+  P const h          = (max - min) / n;
+  auto const lgwt    = legendre_weights(degree, -1.0, 1.0, true);
+  auto const roots   = lgwt[0];
+
+  unsigned int const dof = roots.size();
+
+  fk::vector<P> nodes(mat_dims);
+  for (int i = 0; i < n; i++)
+  {
+    auto p_val = legendre(roots, degree, legendre_normalization::lin);
```
quantumsteve, line 146: Can/should we use the standard library for Legendre polynomials?
https://en.cppreference.com/w/cpp/numeric/special_functions/legendre

src/transformations.cpp
```diff
@@ -128,28 +130,75 @@ std::vector<fk::matrix<P>> gen_realspace_transform(
...
+    fk::vector<P> &real_space, int const transform_dims)
 {
   expect(memory_limit_MB > 0);
 
   std::vector<batch_chain<P, resource::host>> chain;
 
+  // generate transform for each dimension, or only for up to transform_dims
+  // dimensions
+  int const actual_num_dims =
+      transform_dims > 0 ? transform_dims : pde.num_dims;
```
quantumsteve: Can we avoid magic values form `transform_dims`? 

`std::optional<int>` would be more explicit but still ugly. Also, pde.num_dims goes away in #474.

I think the C++17 solution would be to pass a begin and end iterator for `std::vector<dimension<P>>`. C++20 adds std::span and ranges.

src/transformations.cpp
ckendrick: I think that would be a good idea. I'll check if our implementation matches and will swap it out if it does.

quantumsteve: Boost has a few more functions, including one for finding zeros.
https://www.boost.org/doc/libs/1_80_0/libs/math/doc/html/math_toolkit/sf_poly/legendre.html

mkstoyanov: We are not using boost, if we can avoid it. The dependence is enormous, for something as simple as root-fining.

We can pull some tricks from Tasmanian and quadrature build.

src/transformations.cpp
mkstoyanov: If we want to make this expressive, we can turn it into a flag. In all cases (right now), we are calling this with 1 or `pde.num_dims`. We can have an enum or `constexpt` constant that will denote `transform-one` or `transform-all` 

Also the `pde.num_dims` is not going away exactly, it will be converted to different signature, but we can keep both signatures.

src/kronmult.cpp
```diff
@@ -276,7 +276,8 @@ fk::vector<P, mem_type::view, resource::device>
 execute(PDE<P> const &pde, elements::table const &elem_table,
         options const &program_opts, element_subgrid const &my_subgrid,
         fk::vector<P, mem_type::const_view, resource::device> const &x,
-        fk::vector<P, mem_type::view, resource::device> &fx)
+        fk::vector<P, mem_type::view, resource::device> &fx,
+        imex_flag const &flag)
```
mkstoyanov: Don't call the variable `flag`, call it `imex` instead.

The type is `imex_flag` where `imex` refers to a specific algorithm while `flag` refers to the role of the variable (flag to switch to the `imex` algorithm).

When you say just `flag` that could be a switch to pretty much anything, `imex` is the part that really matters.

Alternative names are `imex_mode` or `imex_variant`, there are other possibilities, but `imex` should be part of the name.

src/kronmult.cpp
```diff
@@ -356,7 +372,8 @@ fk::vector<P, mem_type::owner, resource::host>
 execute(PDE<P> const &pde, elements::table const &elem_table,
         options const &program_opts, element_subgrid const &my_subgrid,
         int const workspace_size_MB,
-        fk::vector<P, mem_type::owner, resource::host> const &x)
+        fk::vector<P, mem_type::owner, resource::host> const &x,
+        imex_flag const &flag)
```
mkstoyanov: When it's just a flag, why are we passing by const-reference instead of value?

src/kronmult.hpp
```diff
@@ -19,6 +19,7 @@ fk::vector<P, mem_type::owner, resource::host>
 execute(PDE<P> const &pde, elements::table const &elem_table,
         options const &program_options, element_subgrid const &my_subgrid,
         int const workspace_size_MB,
-        fk::vector<P, mem_type::owner, resource::host> const &x);
+        fk::vector<P, mem_type::owner, resource::host> const &x,
+        imex_flag const &flag = imex_flag::exp);
```
mkstoyanov: Taking a reference to an enum? (I figured how it compiled, but no need to pass by ref)

I don't think imex_flag::exp is a defined variable with an address, there's some compiler trickery here.

src/main.cpp
```diff
@@ -236,13 +247,18 @@ int main(int argc, char **argv)
...
     auto const update_system = i == 0;
-    auto const method        = opts.use_implicit_stepping
-                                   ? asgard::time_advance::method::imp
-                                   : asgard::time_advance::method::exp;
-    auto const time_str = opts.use_implicit_stepping ? "implicit_time_advance"
-                                                     : "explicit_time_advance";
-    auto const time_id  = asgard::tools::timer.start(time_str);
-    auto const sol      = asgard::time_advance::adaptive_advance(
+    auto const method =
+        opts.use_implicit_stepping
```
mkstoyanov, line 251: We can leave this for now, since this code will be deprecated eventually, but the logic is messy. Effectively, we're creating a hierarchy of flags, `implicit` means implicit, while `explicit` means either explicit or imex.

Instead of boolean flags, since we have three options, we should be reading an enum flag so that:
```
auto const method = opts.stepping_method
```

src/main.cpp
```diff
@@ -236,13 +247,18 @@ int main(int argc, char **argv)
...
+            ? asgard::time_advance::method::imp
+            : (opts.use_imex_stepping ? asgard::time_advance::method::imex
+                                      : asgard::time_advance::method::exp);
+    auto const time_str =
+        opts.use_implicit_stepping
+            ? "implicit_time_advance"
+            : (opts.use_imex_stepping ? "imex_time_advance"
+                                      : "explicit_time_advance");
+    auto const time_id = asgard::tools::timer.start(time_str);
```
mkstoyanov: This whole section overuses `auto`, some of these are floats/doubles, some are integers, some strings or const char* ... 

`auto` is good in templates when the type changes and has to be deduced based on complex relation between template parameters, or in cases where it is both clear what `auto` means and the alternative is messy to write `for(auto x : points)` or `auto ip = points.begin()` as opposed to the iterators.

src/matlab_utilities.cpp
```diff
@@ -471,6 +471,48 @@ fk::matrix<P> reshape(fk::matrix<P, mem> mat, int const nrow, int const ncol)
...
+  fk::vector<P> interpolated(coords.size());
+
+  // perform interpolation at each query point
+  for (int i = 0; i < coords.size(); ++i)
+  {
+    P const query = coords[i];
+    // compute distance from query point to sample points, minimum value is the
+    // nearest neighbor to use
+    fk::vector<P> distance(sample.size());
+    std::transform(sample.begin(), sample.end(), distance.begin(),
```
mkstoyanov: We should start writing these with `for` loops.

With all the complexity that has been added to Asgard-C++ solely for the purpose of accommodating MATLAB conversions, this is the one C++-ism that should not have been included. Math-formulas always use simple indexing, and in Math context `std::transform` should be used only to parallelize or to indicate some operation that isn't a simple math formula (e.g., apply operations to data in some link-lists, or something).

src/matlab_utilities.cpp
```diff
@@ -471,6 +471,48 @@ fk::matrix<P> reshape(fk::matrix<P, mem> mat, int const nrow, int const ncol)
...
+                   [query](P const &x) -> P { return std::fabs(x - query); });
+
+    // find minimum distance
+    P min_distance = MAXFLOAT;
```
mkstoyanov: I always feel uneasy when I have to deal with numbers close to `MAXFLOAT`, how about:
```
P min_distance = distance[0];
int min_ind = 0;
for(int j=1; j<distance.size(); j++)
...
```

Better yet, skip the allocation of distance altogether. Merge the `std::transform()` and the for-loop into one for-loop, compute the distances on the fly and do not store them.

src/matlab_utilities.cpp
```diff
@@ -513,4 +555,11 @@ template fk::matrix<float>
 reshape(fk::matrix<float> mat, int const nrow, int const ncol);
 template fk::matrix<int>
 reshape(fk::matrix<int> mat, int const nrow, int const ncol);
+
+template fk::vector<double> interp1(fk::vector<double> sample,
```
mkstoyanov: I think these should accept parameters as `const &` and not copy the values.

src/matlab_utilities_tests.cpp
```diff
@@ -341,3 +341,20 @@ TEMPLATE_TEST_CASE(
...
   }
 }
+
+TEMPLATE_TEST_CASE("interp1", "[matlab]", float, double)
+{
+  SECTION("basic")
+  {
+    fk::vector<TestType> grid{0.0, 1.0, 2.0, 3.0, 4.0};
+    fk::vector<TestType> vals(grid.size());
+    std::transform(grid.begin(), grid.end(), vals.begin(),
```
mkstoyanov: I would `for-loop` this one too.

src/matlab_utilities_tests.cpp
```diff
@@ -341,3 +341,20 @@ TEMPLATE_TEST_CASE(
...
+                   [](TestType const &x) { return std::cos(x); });
+
+    fk::vector<TestType> queries{1.49, 1.51, -0.1, 4.5};
```
mkstoyanov: Switch the order of definitions and the blank spaces. Think of the logic:
* make the `grid`
* compute the `expected`
    * allocate `vals`
    * fill vals with `cos()` of the `grid`
    * allocate and initialize `expected`
* compute the output
    * allocate and initialize the `queries`
    * compute the `output`
* compare the `expected` and the `output`

src/pde/pde_base.hpp
```diff
@@ -157,6 +157,12 @@ enum class flux_type
   lax_friedrich = 0
 };
 
+enum class imex_flag
```
mkstoyanov, line 86: Don't we already have an enum that handles all stepping types? Implicit, explicit, variations of imex? We can have a `null` type or `none` or `unknown`.

That enum type can be used in the new code also.
```
enum class stepping_flag {
  unspecified,
  implicit,
  explicit,
  imex_implicit,
  imex_explicit,
};
```
I don't like using `exp` here, it leads me to thing of `std::exp` or some other exponential operation as opposed to `explicit`.

src/pde/pde_base.hpp
```diff
@@ -327,10 +333,10 @@ class term
 
 public:
   term(bool const time_dependent_in, std::string const name_in,
-       std::initializer_list<partial_term<P>> const partial_terms)
-      : time_dependent(time_dependent_in), name(name_in),
+       std::initializer_list<partial_term<P>> const partial_terms,
+       imex_flag const flag_in = imex_flag::exp)
```
mkstoyanov: We're assigning a `default` value here, to the `imex_flag`, but how meaningful is that?

Should we not assign a default `none` value, which is obvious to ignore in non-imex cases, then force the user to be specific about `imex_flag::imp` and `imex_flag::exp` when imex is used.

src/pde/pde_base.hpp
```diff
@@ -434,6 +442,67 @@ class source
   scalar_func<P> const time_func;
 };
 
+template<typename P>
+class parameter
```
mkstoyanov: All members are public, make it a `struct` and then you get the constructor for free.

src/pde/pde_base.hpp
```diff
@@ -434,6 +442,67 @@ class source
...
+  std::shared_ptr<parameter<P>> get_parameter(std::string const name)
+  {
+    for (auto p : parameters)
+    {
+      if (p->name == name)
+      {
+        return p;
+      }
+    }
+    std::cerr << " could not find parameter with name '" << name << "'\n";
```
mkstoyanov: I think you want to `throw` exception here.

src/pde/pde_base.hpp
```diff
@@ -434,6 +442,67 @@ class source
...
+      : name(name_in), value(value_in)
+  {}
+
+  std::string const name;
+  g_func_type<P> value;
+};
+
+// Singleton class for storing and receiving PDE parameters
+template<typename P>
+#define param_manager parameter_manager<P>::get_instance()
```
mkstoyanov, line 381: The new code will not use global variables masquerading as proper fancy OOP

It's OK here, since the old code is all build on globals anyway, but switch the `#define` and the `template` directives. It's really confusing to have `#define` between `template` and `class` when it the `template` relates only to the class.

src/pde/pde_base.hpp
```diff
@@ -434,6 +442,67 @@ class source
...
+
+  void add_parameter(parameter<P> const &param)
+  {
+    if (get_parameter(param.name) == nullptr)
+    {
+      parameters.push_back(std::make_shared<parameter<P>>(param));
+    }
+    else
+    {
+      std::cerr << " already has a parameter with name '" << param.name
```
mkstoyanov: `throw` exception here

src/pde/pde_base.hpp
```diff
@@ -598,8 +667,30 @@ class PDE
         expect(md_func.size() == static_cast<unsigned>(num_dims) + 1);
       }
     }
+
+    if (cli_input.using_imex() && moments.size() == 0)
+    {
+      std::cerr << "Invalid PDE choice for IMEX time advance. PDE must have "
```
mkstoyanov: Another `throw`

src/pde/pde_base.hpp
```diff
@@ -598,8 +667,30 @@ class PDE
...
+      exit(-1);
+    }
   }
 
+  // copy constructor to create a 1D version of the PDE
+  // this is used in the IMEX time advance to help define 1D mapping from
+  // wavelet to realspace
+  // TODO: there is likely a better way to do this. Another option is to flatten
+  // element table to 1D (see hash_table_2D_to_1D.m)
+  PDE(const PDE &pde)
```
mkstoyanov: **NO!** 

The signature for a copy-constructor should only ever indicate a copy-constructor. Change the signature, it's OK to pass in a type-tag or a dummy variable.

My approach would be to add a type-tag inside the PDE declaration:
```
class PDE {
...
struct extract_dim0 {};
...
}
```
Then change the constructor to have the signature:
```
PDE(const PDE &pde, extract_dim0) { ... }
```
Call the constructor with:
```
PDE pde_1d = PDE(pde, PDE::extract_dim0());
```

The same idea can  be done with an `int` and then defining `constexp int extract_dim0 = 1;`

Ideally, I would like to have an explicit template type-tag, e.g., `PDE::PDE<extract_dim0>(pde)`, but that's not added until later in the C++ standard.

src/time_advance.cpp
```diff
@@ -392,48 +410,229 @@ implicit_advance(PDE<P> const &pde,
 // current solution vector x. on exit, the next solution vector is stored in fx.
 template<typename P>
 fk::vector<P>
-imex_advance(PDE<P> const &pde, adapt::distributed_grid<P> const &adaptive_grid,
+imex_advance(PDE<P> &pde, adapt::distributed_grid<P> const &adaptive_grid,
```
mkstoyanov, line 407: Let's all take a moment and contemplate the meaning of `const` in a context where >90% of our data is global.

src/time_advance.cpp
```diff
@@ -392,48 +410,229 @@ implicit_advance(PDE<P> const &pde,
...
+
+  // Re-create A using new coefficients
+  auto rebuild_system_matrix = [&]() {
+    if (first_time || update_system)
+    {
+      first_time = false;
 
+      build_system_matrix(pde, table, A, grid);
+      // AA = I - dt*A;
+      for (int j = 0; j < A.ncols(); ++j)
```
mkstoyanov: Can we just do a BLAS `scal()` operation here, there's no stride in `A` and it can be done with a single command.

We can leave the diagonal below, since we don't want to create a dummy vector with just ones, but here we can do better.

src/time_advance.cpp
```diff
@@ -392,48 +410,229 @@ implicit_advance(PDE<P> const &pde,
...
+    return (interp1(nodes, mom2_real, {x_v})[0] /
+            param_manager.get_parameter("n")->value(x_v, t)) -
+           std::pow(u, 2);
+  };
+
+  // Update coeffs
+  generate_all_coefficients<P>(pde, transformer);
+
+  // f2 now
+  P const tolerance  = std::is_same_v<float, P> ? 1e-6 : 1e-12;
```
mkstoyanov, line 547: Has this been tested? This tolerance is way too brutal for `GMRES`, especially without a preconditioner.

This can be achieved only if the problem is very well conditioned, e.g., the problem is small. At the very least we should have a CLI option to overwrite this, otherwise we have to scale by the problem size somehow.

mkstoyanov (changes requested):
I hope I'm not too knit picky, the majority should be easy fixes.

Let me know if you have questions, I'm available on Teams and Slack.

src/main.cpp
mkstoyanov: This needs fixing. First `auto` hides the type of the variable `time_str` but then the type is not `str` or `string` as it is `const char*`. The you have to double-check the code to see that `time_id` is actually `std::string`

Let's not overuse `auto`
```
const char *timer_mode = 
    opts.use_implicit_stepping
            ? "implicit_time_advance"
            : (opts.use_imex_stepping ? "imex_time_advance"
                                      : "explicit_time_advance");
std::string const time_id = asgard::tools::timer.start(timer_mode);
```

src/pde/pde_base.hpp
mkstoyanov: OK let's leave it at that for now.

src/pde/pde_base.hpp
```diff
@@ -434,6 +443,65 @@ class source
   scalar_func<P> const time_func;
 };
 
+template<typename P>
+struct parameter
+{
+  parameter(std::string name_in, g_func_type<P> const &value_in)
```
mkstoyanov: I don't think the constructor is needed, when it just takes the members in the same order as defined, the compiler should automatically add this constructor.

src/pde/pde_base.hpp
mkstoyanov: Returning a `nullptr` will lead to an error elsewhere in the code, should just `throw` here.

src/pde/pde_base.hpp
mkstoyanov: Another `throw`

C++ should never use `exit(-1);` (maybe with some really big exception, i.e., potential security concerns, regular errors should just `throw` exceptions)

src/time_advance.cpp
mkstoyanov: We can accept this in the PR, but will need testing and adjusting.

mkstoyanov (changes requested):
Everything touched has been fixed! Making progress.

src/pde/pde_base.hpp
ckendrick: I could be missing something, but I am getting compiler errors when trying to remove this.

mkstoyanov: When you invoke the compiler, the inputs should be passed in a `{ }`, e.g., in the vlassov pde file:
```
    param_manager.add_parameter(parameter<P>{"n", n});
    param_manager.add_parameter(parameter<P>{"u", u});
    param_manager.add_parameter(parameter<P>{"theta", theta});
```

mkstoyanov (approved):

mkstoyanov:
If @quantumsteve doesn't have anything to add, we can go ahead and merge.

quantumsteve:
:shipit:

PR #442, opened 2022-10-10
Fix remaining -Wshadow warnings

quantumsteve:
Please review the [developer documentation](https://github.com/project-asgard/asgard/wiki/developing)
on the wiki of this project that contains help and requirements.

## Proposed changes

Describe what this PR changes and why.  If it closes an issue, link to it here
with [a supported keyword](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

Added -Wshadow to suggested settings and fixed remaining warnings that appeared.

## What type(s) of changes does this code introduce?
_Put an `x` in the boxes that apply._

- [x] Bugfix
- [ ] New feature
- [x] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

### Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Checklist

_Put an x in the boxes that apply. You can also fill these out after creating
the PR. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- [x] this PR is up to date with current the current state of 'develop'
- [x] code added or changed in the PR has been clang-formatted
- [ ] this PR adds tests to cover any new code, or to catch a bug that is being fixed
- [ ] documentation has been added (if appropriate)

PR #476, opened 2022-12-14
Fix term selection option

ckendrick:

## Proposed changes

This fixes the PDE individual term selection option (`--terms`) after changes in #442. A few leftover shadowed `num_terms` are changed to use `num_terms_in` when parsing the input term string.

Some term selection examples to test:
- `./asgard -p diffusion_2 --terms "1 0"`
- `./asgard -p fokkerplanck_2d_complete_case4 --terms "1 0 1 1 0 0"`

- [x] Bugfix
- [ ] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist

### DIFF
--- a/src/pde/pde_base.hpp
+++ b/src/pde/pde_base.hpp
@@ -639,7 +639,7 @@ public:
   }
 
 private:
-  int get_num_terms(parser const &cli_input, int const num_terms_in) const
+  int get_num_terms(parser const &cli_input, int const max_num_terms) const
   {
     // returns either the number of terms set in the PDE specification, or the
     // number of terms toggled on by the user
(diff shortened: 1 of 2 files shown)


quantumsteve:
May we add a test to reproduce this error?

mkstoyanov (approved):
I'm not sure if we can add a test for those, it's mostly a matter of sync between CLI and expected C++ behavior. The test will have to make a call with parameters and check the output, can be done but I'm not sure if it's worth the trouble.

src/pde/pde_base.hpp
```diff
@@ -469,7 +469,7 @@ class PDE
     expect(num_terms > 0);
 
     expect(dimensions.size() == static_cast<unsigned>(num_dims));
-    expect(terms.size() == static_cast<unsigned>(num_terms));
+    expect(terms.size() == static_cast<unsigned>(num_terms_in));
```
mkstoyanov: Why do we have two `num_terms_in` and `num_terms`?

It may be just my ignorance, but why is one modified by the CLI conditions? It is extremely confusing to have both.

src/pde/pde_base.hpp
```diff
@@ -502,7 +502,7 @@ class PDE
     if (num_active_terms != 0)
     {
       auto const active_terms = cli_input.get_active_terms();
-      for (auto i = num_terms - 1; i >= 0; --i)
+      for (auto i = num_terms_in - 1; i >= 0; --i)
```
mkstoyanov: Are you sure about this? 

In the loop, we're modifying `active_terms(i)` which comes from the `cli_input.get_active_terms()` but you're saying that we should loop from `num_terms_in` which comes from the PDE definition and is not necessarily the CLI value.

mkstoyanov (changes requested):
This is confusing, let's double-check since it doesn't seem we're doing the right thing in all cases.

src/pde/pde_base.hpp
mkstoyanov: Keep the `num_terms` but rename the `num_terms_in` to `max_num_terms`

src/pde/pde_base.hpp
```diff
@@ -650,7 +650,8 @@ class PDE
     if (num_active_terms != 0 && num_active_terms != num_terms_in)
     {
       std::cerr << "failed to parse dimension-many active terms - parsed "
-                << num_active_terms << " terms, expected " << num_terms << "\n";
+                << num_active_terms << " terms, expected " << num_terms_in
```
mkstoyanov: Does this compile? There are several `num_terms_in` left in the code, use an automated search to find them.

mkstoyanov (approved):

mkstoyanov:
Sorry about pushing on another branch, but I couldn't look at the `exit()` statements and the name `num_terms_in` was still confusing in the method.

PR #479, opened 2023-01-10
Convert from wavelet to real space for plotting

quantumsteve:

## Proposed changes

Update asgard_discretization_test to convert from wavelet to real/dense space for plotting. Update functions to use `dimension_description`. 

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/src/asgard_discretization_tests.cpp
+++ b/src/asgard_discretization_tests.cpp
@@ -1,7 +1,7 @@
 #include "asgard_discretization.hpp"
 #include "build_info.hpp"
-#include "transformations.hpp"
 #include "tests_general.hpp"
+#include "transformations.hpp"
 
 using namespace asgard;
 
--- a/src/asgard_discretization_tests.cpp
+++ b/src/asgard_discretization_tests.cpp
@@ -70,7 +70,7 @@ TEMPLATE_TEST_CASE("testing construction of a basic field_discretization",
 
   REQUIRE(init.size() == 32);
 
-  auto const real_space_size = real_solution_size(dims.list);
+  auto const real_space_size = grid.size();
   fk::vector<TestType> real_space(real_space_size);
   // temporary workspaces for the transform
   fk::vector<TestType, mem_type::owner, resource::host> workspace(
@@ -82,7 +82,7 @@ TEMPLATE_TEST_CASE("testing construction of a basic field_discretization",
           fk::vector<TestType, mem_type::view, resource::host>(
               workspace, real_space_size, real_space_size * 2 - 1)};
 
-  static auto const default_workspace_cpu_MB = 187000;
+  auto const default_workspace_cpu_MB = 187000;
 
   // transform initial condition to realspace
   wavelet_to_realspace(dims.list, init, grid.grid->get_table(), transformer,
--- a/src/asgard_discretization_tests.cpp
+++ b/src/asgard_discretization_tests.cpp
@@ -67,17 +67,16 @@ TEMPLATE_TEST_CASE("testing construction of a basic field_discretization",
 
   REQUIRE(init.size() == 32);
 
-  auto const real_space_size = real_solution_size(dims.list);
-  fk::vector<TestType> real_space(real_space_size);
+  auto const dense_size = dense_space_size(dims.list);
+  fk::vector<TestType> real_space(dense_size);
   // temporary workspaces for the transform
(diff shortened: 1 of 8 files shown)


src/asgard_discretization_tests.cpp
```diff
@@ -65,4 +69,40 @@ TEMPLATE_TEST_CASE("testing construction of a basic field_discretization",
   fk::vector<TestType> init = grid.get_initial_conditions(pos_field);
 
   REQUIRE(init.size() == 32);
+
+  auto const real_space_size = real_solution_size(dims.list);
```
mkstoyanov: `real_space_size` should be replaced by just `grid.size()`

Should be the right thing to put here.

src/matlab_plot.cpp
```diff
@@ -236,15 +237,23 @@ template<typename P>
...
+  int const ndims = dims.size();
 
   fk::vector<P> center_coords(ndims * table.size());
 
   // Iterate over dimensions first since matlab needs col-major order
   for (int d = 0; d < ndims; d++)
   {
-    P const max = pde.get_dimensions()[d].domain_max;
-    P const min = pde.get_dimensions()[d].domain_min;
+    P const max = dims[d].domain_max;
```
mkstoyanov: I know you're simply refactoring here, but in general, I don't like to declare a variable that would be used only one time. There is no need for max, we only need to define `min` and `rng`.

Also, both `min` and `rng` are bad names, `domain_min` and `domain_range` should work well on an editor.

src/matlab_plot.cpp
```diff
@@ -299,25 +362,83 @@ void matlab_plot::init_plotting(PDE<P> const &pde, elements::table const &table)
...
+
+  nodes_ = std::vector<matlab::data::Array>(dims.size());
+
+  for (int i = 0; i < dims.size(); i++)
+  {
+    auto const &dim = dims[i];
+    auto const &node_list =
+        generate_nodes(dim.degree, dim.level, dim.d_min, dim.d_max);
+    nodes_[i] =
+        create_array({1, static_cast<size_t>(node_list.size())}, node_list);
```
mkstoyanov: Here we can really use some handy overloads, maybe in this or maybe in the next PR.

Since we have multiple calls to `generate_nodes()` that take the same inputs from one dim, overload so we can just say `generate_nodes(dim[i])`.

Similatly, we can have an overload to `create_array()` that default to the full `node_list` if a range is not provided. That way, the code here becomes:
```
nodes_[i] = create_array(generate_nodes(dim[i]));
```

We can also shortcut the two methods into one. Since this is called multiple times with the same pattern, it will be helpful.

mkstoyanov (changes requested):
Small notes, feel free to push back if you disagree.

src/asgard_discretization_tests.cpp
quantumsteve: Changing this is causing the test to fail. I need to investigate further.

src/matlab_plot.cpp
quantumsteve: The MATLAB syntax sent me down a bit of a rabbit hole, but think think it's a bit cleaner now.

src/asgard_discretization_tests.cpp
mkstoyanov: This is off by a factor of 2, not sure why. I'm investigating too.

quantumsteve: `real_solution_size(dims.list) == 64` while `grid.size() == 32`

mkstoyanov: The `grid.size()` returns the number of degrees of freedom for the sparse grid solution.

The `real_solution_size()` returns the number of degrees of freedom for the "full" or "dense" grid. The plot picture needs the dense variables, which makes sense, but the naming convention is bad.

I want to see the word "dense", e.g., "dense_space_size", it would be much clearer of what it mean.

If `real_solution_size()` is called too many times, then leave the method alone and rename just the variable. If it is a matter of fixing 5-10 lines, let's rename the method `real_solution_size()`.

A larger rename can be done in another PR.

quantumsteve: I made a first attempt at renaming variables.

mkstoyanov (approved):

mkstoyanov:
Merge when ready!

PR #481, opened 2023-01-18
IMEX updates and new moment functions

### LINKED ISSUE #477: Extending moment calculations from 1x-1v to 1x-2v and 1x-3v
@ckendrick's recent PR has the moment calculation, but it's hardcoded to a 1x-1v problem.  This calculation needs to be extended to the 1x-2v and 1x-3v case.  The code to do this is given in https://github.com/project-asgard/DG-SparseGrid/blob/macro-micro/moment_reduced_matrix.m#L56-L159

ckendrick:

## Proposed changes

This fixes an issue in the IMEX time advance where the system matrix was not being updated each time step. The function to create the system matrix was also updated to only include implicit terms when using IMEX.

This PR also adds moment matrix calculations for initial support with running 1x2v and 1x3v problems.

Closes #477.

- [x] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist

### DIFF
--- a/src/batch.cpp
+++ b/src/batch.cpp
@@ -531,7 +531,8 @@ linear_coords_to_indices(PDE<P> const &pde, int const degree,
 // routines with explicit time advance
 template<typename P>
 void build_system_matrix(PDE<P> const &pde, elements::table const &elem_table,
-                         fk::matrix<P> &A, element_subgrid const &grid)
+                         fk::matrix<P> &A, element_subgrid const &grid,
+                         const bool use_imex, imex_flag const imex)
 {
   // assume uniform degree for now
   int const degree    = pde.get_dimensions()[0].get_degree();
@@ -554,6 +555,8 @@ void build_system_matrix(PDE<P> const &pde, elements::table const &elem_table,
                          pde.get_coefficients(k, d).clone_onto_host());
     }
   }
+
+  auto terms = pde.get_terms();
 
   // loop over elements
   for (auto i = grid.row_start; i <= grid.row_stop; ++i)
@@ -593,7 +596,15 @@ void build_system_matrix(PDE<P> const &pde, elements::table const &elem_table,
       {
         std::vector<fk::matrix<P>> kron_vals;
         fk::matrix<P> kron0(1, 1);
-        kron0(0, 0) = 1.0;
+        // if using imex, include only terms that match the flag
+        if (!use_imex || (use_imex && terms[k][0].flag == imex))
+        {
+          kron0(0, 0) = 1.0;
+        }
+        else
+        {
+          kron0(0, 0) = 0.0;
+        }
         kron_vals.push_back(kron0);
         for (int d = 0; d < pde.num_dims; d++)
         {
@@ -700,10 +711,12 @@ template void batched_gemv(batch<double, resource::host> const &a,
 
 template void
 build_system_matrix(PDE<double> const &pde, elements::table const &elem_table,
-                    fk::matrix<double> &A, element_subgrid const &grid);
+                    fk::matrix<double> &A, element_subgrid const &grid,
+                    const bool use_imex, imex_flag const imex);
 template void
 build_system_matrix(PDE<float> const &pde, elements::table const &elem_table,
-                    fk::matrix<float> &A, element_subgrid const &grid);
+                    fk::matrix<float> &A, element_subgrid const &grid,
+                    const bool use_imex, imex_flag const imex);
 
 template class batch_chain<double, resource::device, chain_method::realspace>;
 template class batch_chain<double, resource::host, chain_method::realspace>;
--- a/src/batch.cpp
+++ b/src/batch.cpp
@@ -605,7 +605,7 @@ void build_system_matrix(PDE<P> const &pde, elements::table const &elem_table,
         {
           kron0(0, 0) = 0.0;
         }
-        kron_vals.push_back(kron0);
+        kron_vals.push_back(std::move(kron0));
         for (int d = 0; d < pde.num_dims; d++)
         {
           fk::matrix<P, mem_type::view> op_view = fk::matrix<P, mem_type::view>(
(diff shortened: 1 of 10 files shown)


src/batch.cpp
```diff
@@ -531,7 +531,8 @@ linear_coords_to_indices(PDE<P> const &pde, int const degree,
 // routines with explicit time advance
 template<typename P>
 void build_system_matrix(PDE<P> const &pde, elements::table const &elem_table,
-                         fk::matrix<P> &A, element_subgrid const &grid)
+                         fk::matrix<P> &A, element_subgrid const &grid,
+                         const bool use_imex, imex_flag const imex)
```
mkstoyanov: Why do we need two `imex` flags? When we have the `explicit`, `implicit` and `unspecified` options we can use the `unspecified` option to indicate `no-imex`.

src/matlab_plot.cpp
```diff
@@ -347,14 +358,19 @@ void matlab_plot::init_plotting(std::vector<dimension<P>> const &dims,
 
   nodes_.clear();
 
+  int d = 0;
```
quantumsteve, line 361:
```suggestion
  int d = dims.size();
```

src/matlab_plot.cpp
```diff
@@ -347,14 +358,19 @@ void matlab_plot::init_plotting(std::vector<dimension<P>> const &dims,
...
   for (auto const &dim : dims)
   {
     nodes_.push_back(create_array(generate_nodes(dim)));
+    this->set_var(std::string("nodes_dim" + std::to_string(d)), nodes_.back());
+    d += 1;
```
quantumsteve:
```suggestion
    ++d;
```

src/matlab_plot.cpp
```diff
@@ -367,14 +383,19 @@ void matlab_plot::init_plotting(
...
```
quantumsteve:
```suggestion
    ++d;
```

src/matlab_plot.cpp
```diff
@@ -173,6 +173,17 @@ void matlab_plot::set_var(std::string const &name,
   matlab_inst_->setVariable(name, var, type);
 }
 
+// pushes a vector to a matlab workspace with the given name
+template<typename T>
+void matlab_plot::push(std::string const &name, fk::vector<T> const &data,
+                       ml_wksp_type const type)
+{
+  // create a matlab array object
+  auto ml_array = create_array({static_cast<size_t>(data.size()), 1}, data);
```
quantumsteve:
```suggestion
  auto ml_array = create_array(data);
```

src/moment.cpp
```diff
@@ -130,6 +151,114 @@ void moment<P>::createMomentReducedMatrix(PDE<P> const &pde,
   }
 }
 
+template<typename P>
+void moment<P>::createMomentReducedMatrix_3d(PDE<P> const &pde,
```
mkstoyanov: At a glance, `createMomentReducedMatrix_3d()` and `createMomentReducedMatrix_4d()` have lots of repeated code. There should be a way to either generalize this, or perhaps template on "num-velocity-dims" that will take values 1, 2, 3.

Then:
* `d_dim_3` will be unused for the 2D template and the optimizer should get rid of that
* `g_vec_3` could be initialized to empty vector for the 2D case and never used
* some of the constants (e.g., 2 or 3) can be initialized from the template parameters
* the inner-most loop of the three nested three loops can be put to loop over a single entry (based on the template parameter).
* Note it is Ok to have if-statements inside for-loops when the `if` depends on a parameter that is known at compile time (e.g., template parameter)

mkstoyanov (changes requested):

src/batch.cpp
ckendrick: Both flags are needed since this code was also being used for non-IMEX things - to make sure that all terms still get included when running without IMEX, even if the PDE has a term marked `imex_explicit` or `imex_implicit`. Additionally, `program_opts` was not being passed through, so this was also a quick fix without changing everything to pass `program_opts` (although the latest changes in `develop` have made `program_opts` available).

However, since the matrix-free changes have been merged into `develop` and this branch is now rebased to include those changes, this is no longer being used anymore. I would like to still keep it for now in case we need to temporarily change back to this. I can open an issue to remove this code in the future outside of this PR.

src/moment.cpp
ckendrick: I consolidated these functions down into one to minimize repeated code. I can open an issue to clean it up further if needed, but this code is only meant to be temporary and will be deleted after the new version. For now, it would be good to get the rest of these changes in.

src/batch.cpp
```diff
@@ -593,7 +596,15 @@ void build_system_matrix(PDE<P> const &pde, elements::table const &elem_table,
...
+        // if using imex, include only terms that match the flag
+        if (!use_imex || (use_imex && terms[k][0].flag == imex))
+        {
+          kron0(0, 0) = 1.0;
+        }
+        else
+        {
+          kron0(0, 0) = 0.0;
+        }
         kron_vals.push_back(kron0);
```
quantumsteve:
```suggestion
        kron_vals.push_back(std::move(kron0));
```

quantumsteve: line 616 has the same issue. fk::matrix copied when it should be moved.

src/batch.cpp
quantumsteve:
```suggestion
                         std::optional<imex_flag> const imex)
```
Would std::optional bring us down to a single additional parameter?

src/moment.cpp
```diff
@@ -90,41 +90,108 @@ linear_coords_to_indices(PDE<P> const &pde, int const degree,
...
+  // Define additional g_vecs for higher number of v dimensions
+  fk::vector<P> g_vec_2, g_vec_3;
+  if (nvdim >= 2)
+  {
+    g_vec_2 = this->fList[moment_idx][2];
+  }
+  if (nvdim >= 3)
+  {
+    g_vec_3 = this->fList[moment_idx][3];
+  }
```
quantumsteve:
```suggestion
    if (nvdim >= 3)
    {
      g_vec_3 = this->fList[moment_idx][3];
    }
  }
```

src/batch.cpp
mkstoyanov: The only place where the flag is being used is in the line:
```
if (!use_imex || (use_imex && terms[k][0].flag == imex))
```

With my suggestion, the condition will become:
```
if (imex == imex::unspecified || terms[k][0].flag == imex)
```

What am I missing here?

mkstoyanov (approved):
Great! Let's update in case there's a conflict and we can merge.

quantumsteve (approved):

PR #482, opened 2023-01-19
matrix-free implicit solve

### LINKED ISSUE #472: Matrix-free implicit solve using batched Kronecker gemv.

Currently any implicit solve requires constructing the (assumed dense) global stiffness matrix.   For any run in 3D or higher, this is very memory intensive on low-resolution problems.  Moreover, the matrix construction is serial.

Since we have a parallel implementation of the Kronecker gemv (for the explicit time advance), we should utilize this to build a matrix-free implicit backward Euler solver using either GMRES for BICGSTAB.  

We can ignore preconditioning for ease of implementation.

quantumsteve:

## Proposed changes

This uses `kronmult::execute` instead of building the global stiffness matrix.

fixes #472 

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist

### DIFF
--- a/src/solver.cpp
+++ b/src/solver.cpp
@@ -6,10 +6,10 @@
 namespace asgard::solver
 {
 template<typename precision>
-class MatrixFree
+class matrix_free
 {
 public:
-  MatrixFree(PDE<precision> const &pde, elements::table const &elem_table,
+  matrix_free(PDE<precision> const &pde, elements::table const &elem_table,
              options const &program_options, element_subgrid const &my_subgrid,
              int const workspace_size_MB)
       : pde_{pde}, elem_table_{elem_table}, program_options_{program_options},
@@ -33,10 +33,10 @@ private:
 };
 
 template<typename precision>
-class DenseMatrix
+class dense_matrix
 {
 public:
-  DenseMatrix(fk::matrix<precision> const &A) : A_{A} {}
+  dense_matrix(fk::matrix<precision> const &A) : A_{A} {}
   // template<typename xmem, typename ymem, typename resrc>
   void operator()(fk::vector<precision> const &x, fk::vector<precision> &y,
                   precision const alpha = 1.0, precision const beta = 0.0)
@@ -55,7 +55,7 @@ P simple_gmres(fk::matrix<P> const &A, fk::vector<P> &x, fk::vector<P> const &b,
                fk::matrix<P> const &M, int const restart, int const max_iter,
                P const tolerance)
 {
-  return simple_gmres(DenseMatrix{A}, x, b, M, restart, max_iter, tolerance);
+  return simple_gmres(dense_matrix{A}, x, b, M, restart, max_iter, tolerance);
 }
 
 template<typename P>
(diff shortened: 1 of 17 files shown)


src/solver.cpp
```diff
@@ -1,20 +1,83 @@
...
+              int const workspace_size_MB)
+      : pde_{pde}, elem_table_{elem_table}, program_options_{program_options},
+        my_subgrid_{my_subgrid}, workspace_size_MB_{workspace_size_MB}
+  {}
+  void operator()(fk::vector<precision> const &x, fk::vector<precision> &y,
+                  precision const alpha = 1.0, precision const beta = 0.0)
+  {
+    auto tmp = kronmult::execute(pde_, elem_table_, program_options_,
+                                 my_subgrid_, workspace_size_MB_, x);
+    tmp      = x - tmp * pde_.get_dt();
```
quantumsteve: `matrix_free::operator()` calculates (I - dt * A)x while `dense_matrix::operator()` calculates A*x. While this seems to work, I fear it's not intuitive and could cause issues in the future.

suggestions? I thought about creating a third class that could be passed as `matrix_replacement` but am hesitant to move implementation details into the header.

mkstoyanov: * Do not think of these as matrices, dense and matrix free. Think of them as operators that need to be inverted. 
* Be a bit more explicit when something is just a wrapper.
* In place of `dense_matrix` and `matrix-free` can we just use lambdas?

If I have to give names here, I would use `dense_matrix_wrapper` and `euler_operator`, that would be clear enough that these are different.

I still prefer using nameless lambdas that simply capture what they need.

src/time_advance.cpp
```diff
@@ -616,8 +597,8 @@ imex_advance(PDE<P> &pde, adapt::distributed_grid<P> const &adaptive_grid,
 
   // Final stage f3
   fk::vector<P> f_3(x.size());
-  solver::simple_gmres(A, f_3, x, fk::matrix<P>(), restart, max_iter,
-                       tolerance);
+  solver::simple_gmres(pde, table, program_opts, grid, workspace_size_MB, f_3,
```
mkstoyanov: We are passing `workspace_size_MB` a lot, but I don't see the actual workspace. If the workspace is some hidden static/global variable, then the size should be part of that, or perhaps a member of `pde` `table` or `grid`, whichever is appropriate.

src/solver.cpp
quantumsteve: I changed both to lambdas.

src/time_advance.cpp
quantumsteve: Here's the hidden workspace variable. Unfortunately, it's not constructed until after the number of sub-grids are calculated from  `workspace_size_MB`
https://github.com/project-asgard/asgard/blob/99fe77aba3a6cfba5093874fc8fd903be0459c89/src/kronmult.cpp#L143

I'm leaning towards moving it to the options class where it would be a command-line option. What do you think?

mkstoyanov: Yeah, this is messy.

If we put it in the CLI options, will that tie it as part of the `pde` class? It's OK, so long as we are not stuck to the deprecated `pde` class.

mkstoyanov (approved):

quantumsteve:
I'm going to wait for more builds to finish before merging this branch.

PR #495, opened 2023-02-07
Sum of separable functions for analytical solutions

### LINKED ISSUE #488: Sum of separable functions for analytic solution
The analytic solution needs to be able to accept a sum of separable functions.  This should reuse the code from the initial condition case.  

Assigning @ckendrick

ckendrick:

## Proposed changes

This changes how analytical solutions are specified in the PDE to be a sum of separable functions rather than just a single function.

Closes #488

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist

### DIFF
--- a/src/transformations.hpp
+++ b/src/transformations.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "adapt.hpp"
 #include "asgard_dimension.hpp"
 #include "basis.hpp"
 #include "distribution.hpp"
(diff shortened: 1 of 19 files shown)


src/transformations.hpp
```diff
@@ -175,6 +176,14 @@ fk::vector<P> forward_transform(
   return transformed;
 }
 
+template<typename P>
+fk::vector<P> sum_separable_funcs(
+    std::vector<md_func_type<P>> const &funcs,
```
mkstoyanov, line 184: Instead of changing every single PDE file, why don't we add an overload to this that will accept single-instance parameters and then convert those into vectors with one entry.

mkstoyanov (changes requested):

src/transformations.hpp
ckendrick: I think we would need to also overload the PDE constructor to make this work. This PR is changing the definition of `exact_vector_funcs` in the PDE class to be a vector of vectors of functions, which is how it should have been originally.

mkstoyanov: You're right, we would need a few overloads, but the end-result would be much cleaner. 

The user will make a single function when one function is needed and a vector when it is needed. The overload(s) will handle both cases but simply adding a conversion layer of putting the single function into a vector and passing on.

mkstoyanov (dismissed):

mkstoyanov (approved):

PR #496, opened 2023-02-07
Adjust IO and add element metadata

ckendrick:

## Proposed changes

This adds an alternative function used to save data with HDF5 along with a simple python script to read and plot the data.

The existing IO function saves a single file throughout the simulation but resizes the HDF5 dataset dimensions for each timestep, and does not save any grid metadata needed for plotting.

The approach in this PR is to create an individual file for each timestep with basic metadata (time and element nodes). This probably needs to be rewritten in a cleaner manner, but this was used to quickly write data files and plot them.

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04 + HDF5

## Checklist

### DIFF
--- a/scripts/read_output.py
+++ b/scripts/read_output.py
@@ -18,7 +18,13 @@ def plot_from_file(filename, dataset, ax = plt):
 
     tmp = data_file['soln'][()]
 
-    ax.contourf(nodes, nodes, tmp.reshape((len(nodes), len(nodes))).transpose())
+    try:
+        # reshape to 2D matrix
+        tmp = tmp.reshape((len(nodes), len(nodes))).transpose()
+    except ValueError:
+        raise RuntimeError("Could not reshape solution vector, was it run with full grid option?")
+
+    ax.contourf(nodes, nodes, tmp)
     ax.set_title("t = {}".format(data_file['time'][()]))
 
 
--- a/scripts/read_output.py
+++ b/scripts/read_output.py
@@ -1,0 +1,36 @@
+#!/bin/env python
+
+import os
+import sys
+
+import h5py
+
+import matplotlib.pyplot as plt
+
+def read_data(filename, dataset):
+    data_file = h5py.File(filename, 'r')
+
+    print(data_file)
+    print(data_file.keys())
+
+    nodes = data_file['nodes'][()]
+    soln = data_file['soln'][()]
+
+    tmp = data_file['soln'][()]
+
+    fig, ax = plt.subplots()
+    ax.contourf(nodes, nodes, tmp.reshape((len(nodes), len(nodes))).transpose())
+    ax.set_title("t = {}".format(data_file['time'][()]))
+    plt.show()
+
+
+if __name__ == '__main__':
+    if len(sys.argv) <= 1:
+        raise RuntimeError("Expected a datafile")
+
+    input_fname = sys.argv[1]
+
+    if not os.path.exists(input_fname):
+        raise RuntimeError("File '{}' does not exist".format(input_fname))
+
+    read_data(input_fname, 'asgard')
--- a/scripts/read_output.py
+++ b/scripts/read_output.py
@@ -7,7 +7,7 @@ import h5py
 
 import matplotlib.pyplot as plt
 
-def read_data(filename, dataset):
+def plot_from_file(filename, dataset, ax = plt):
     data_file = h5py.File(filename, 'r')
 
     print(data_file)
@@ -18,10 +18,8 @@ def read_data(filename, dataset):
 
     tmp = data_file['soln'][()]
 
-    fig, ax = plt.subplots()
     ax.contourf(nodes, nodes, tmp.reshape((len(nodes), len(nodes))).transpose())
     ax.set_title("t = {}".format(data_file['time'][()]))
-    plt.show()
 
 
 if __name__ == '__main__':
@@ -33,4 +31,7 @@ if __name__ == '__main__':
     if not os.path.exists(input_fname):
         raise RuntimeError("File '{}' does not exist".format(input_fname))
 
-    read_data(input_fname, 'asgard')
+    fig, ax = plt.subplots()
+    plot_from_file(input_fname, 'asgard', ax)
+
+    plt.show()
--- a/src/io.hpp
+++ b/src/io.hpp
@@ -64,7 +64,6 @@ void write_output(PDE<P> const &pde, fk::vector<P> const &vec, P const time,
 {
   std::string const output_file_name =
       output_dataset_name + "_" + std::to_string(file_index) + ".h5";
-  unsigned int vec_size = (unsigned int)vec.size();
 
   auto const dims = pde.get_dimensions();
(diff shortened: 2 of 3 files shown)


src/io.hpp
```diff
@@ -54,4 +56,33 @@ void update_output_file(HighFive::DataSet &dataset, fk::vector<P> const &vec,
...
 }
+
+template<typename P>
+void write_output(PDE<P> const &pde, fk::vector<P> const &vec, P const time,
+                  int const file_index,
+                  std::string const output_dataset_name = "asgard")
+{
+  std::string const output_file_name =
+      output_dataset_name + "_" + std::to_string(file_index) + ".h5";
+  unsigned int vec_size = (unsigned int)vec.size();
```
```suggestion
  unsigned int vec_size = static cast<unsigned int>(vec.size());
```

quantumsteve: Is `vec_size` unused?

scripts/read_output.py
```diff
@@ -0,0 +1,36 @@
...
+
+    print(data_file)
+    print(data_file.keys())
+
+    nodes = data_file['nodes'][()]
+    soln = data_file['soln'][()]
+
+    tmp = data_file['soln'][()]
+
+    fig, ax = plt.subplots()
```
mkstoyanov: Instead of creating your own `axis` object, take one from the user.

`def read_data(filename, dataset, ax = plt):`

You can call `set_title` inside the `read_data`, but call `show` in `__main__`. That would give us and future users more flexibility when arranging plots and decorating the plots, e.g., we can import `read_output` as a module and make more complex plots.

scripts/read_output.py
```diff
@@ -0,0 +1,36 @@
+#!/bin/env python
+
+import os
+import sys
+
+import h5py
+
+import matplotlib.pyplot as plt
+
+def read_data(filename, dataset):
```
mkstoyanov: Change this to `plot_from_file` or something like that, reading is only one of the operations done here.

src/main.cpp
```diff
@@ -321,12 +321,11 @@ int main(int argc, char **argv)
 #ifdef ASGARD_IO_HIGHFIVE
     if (opts.should_output_wavelet(i))
     {
-      asgard::update_output_file(output_dataset, f_val);
+      asgard::write_output(*pde, f_val, time, i + 1, "asgard_wavelet");
```
mkstoyanov, line 324: Isn't this creating more dependence on the `*pde` variable, we want to avoid passing in the object, especially if all we need is `dims`.

src/main.cpp
```diff
@@ -321,12 +321,11 @@ int main(int argc, char **argv)
...
     }
     if (opts.should_output_realspace(i))
     {
-      asgard::update_output_file(output_dataset_real, real_space,
-                                 realspace_output_name);
+      asgard::write_output(*pde, real_space, time, i + 1, "asgard_real");
```
mkstoyanov, line 328: Another case of using `*pde`

src/main.cpp
```diff
@@ -318,12 +318,11 @@ int main(int argc, char **argv)
...
```
quantumsteve, line 328:
```
[55/90] Building CXX object CMakeFiles/asgard.dir/src/main.cpp.o
/home/svh/Documents/asgard/src/main.cpp: In function ‘int main(int, char**)’:
/home/svh/Documents/asgard/src/main.cpp:218:14: warning: unused variable ‘realspace_output_name’ [-Wunused-variable]
  218 |   auto const realspace_output_name = "asgard_realspace";
      |              ^~~~~~~~~~~~~~~~~~~~~
[90/90] Linking CXX executable tensors-tests
```

src/io.hpp
```diff
@@ -54,4 +56,32 @@ void update_output_file(HighFive::DataSet &dataset, fk::vector<P> const &vec,
...
+void write_output(PDE<P> const &pde, fk::vector<P> const &vec, P const time,
+                  int const file_index,
+                  std::string const output_dataset_name = "asgard")
+{
+  std::string const output_file_name =
+      output_dataset_name + "_" + std::to_string(file_index) + ".h5";
+
+  auto const dims = pde.get_dimensions();
+  // TODO: FIX - assumes levels are const across dims
+  static auto const nodes =
```
quantumsteve: Is `nodes` always consistent throughout the lifetime of the program? 

I'm more concerned about a user writing something we cannot control.

src/main.cpp
quantumsteve: Should we remove these lines so the old files aren't created?

https://github.com/project-asgard/asgard/blob/c9af7ed8a1208068ef7d4c6dfb5f86c461cf78c7/src/main.cpp#LL212-L221C7

scripts/read_output.py
```diff
@@ -0,0 +1,37 @@
...
+    ax.set_title("t = {}".format(data_file['time'][()]))
+
+
+if __name__ == '__main__':
+    if len(sys.argv) <= 1:
+        raise RuntimeError("Expected a datafile")
+
+    input_fname = sys.argv[1]
+
+    if not os.path.exists(input_fname):
```
quantumsteve, line 37: Would you also add an error message if someone opens `asgard_wavelet*.h5` with this script?

src/main.cpp
ckendrick: I had forgot to change those lines to write the initial condition with the "new" function, so I updated those. The warning should be fixed now too.

src/io.hpp
ckendrick: This code is changed in the collisional landau pde PR but I agree `nodes` should not be static here.

scripts/read_output.py
ckendrick: I don't mind adding a check for this, although technically the script will work with both (the wavelet contour will just look like noise).
Updated versions of this script also plot the electric field and moments, which are the same for either realspace or wavelet mode which is why I am hesitant to add this.

quantumsteve:
```
./asgard -p continuity_2 -n 627 -d 3 -l "4 4" -m "4 4" -w 100
python3 ../scripts/read_output.py asgard_wavelet_600.h5
```
```
Invalid MIT-MAGIC-COOKIE-1 key<HDF5 file "asgard_wavelet_600.h5" (mode r)>
<KeysViewHDF5 ['nodes', 'soln', 'time']>
Traceback (most recent call last):
  File "../scripts/read_output.py", line 35, in <module>
    plot_from_file(input_fname, 'asgard', ax)
  File "../scripts/read_output.py", line 21, in plot_from_file
    ax.contourf(nodes, nodes, tmp.reshape((len(nodes), len(nodes))).transpose())
ValueError: cannot reshape array of size 432 into shape (48,48)
```

src/main.cpp
quantumsteve: :+1:

src/io.hpp
quantumsteve: :+1:

mkstoyanov (approved):

PR #511, opened 2023-03-06
Add collisional landau pde

ckendrick:

## Proposed changes

- Adds the collisional landau damping PDE
- Expands on adding additional data (E field, moments) and metadata (nodes for each dimension, levels, degree, pde info) to the HDF5 output files. The electric field is now stored as a member variable in the PDE class as a result.
- Saves GMRES iteration counts (inner, outer) and error to HDF5 files.
- Adds more performance timers to the IMEX advance and across the code. These are mainly meant to be temporary to show areas for optimization later.
- Fixes a bug in the `vlasov_lb_full_f` pde where the wrong imex flag was used.

This should be merged after #496. 

- [x] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist

### DIFF
--- a/src/coefficients.cpp
+++ b/src/coefficients.cpp
@@ -19,6 +19,7 @@ void generate_all_coefficients(
     PDE<P> &pde, basis::wavelet_transform<P, resource::host> const &transformer,
     P const time, bool const rotate)
 {
+  tools::timer.start("gen_coefficients");
   expect(time >= 0.0);
 
   for (auto i = 0; i < pde.num_dims; ++i)
@@ -65,6 +66,7 @@ void generate_all_coefficients(
     }
     pde.rechain_dimension(i);
   }
+  tools::timer.stop("gen_coefficients");
 }
 
 template<typename P>
--- a/src/pde/pde_collisional_landau.hpp
+++ b/src/pde/pde_collisional_landau.hpp
@@ -37,7 +37,9 @@ private:
   static bool constexpr has_analytic_soln_     = false;
   static int constexpr default_degree          = 3;
 
-  static P constexpr nu = 1.0; // collision frequency
+  static P constexpr nu       = 1.0;    // collision frequency
+  static P constexpr A        = 1.0e-4; // amplitude
+  static P constexpr theta_in = 1.0;
 
   static fk::vector<P>
   initial_condition_dim_x_0(fk::vector<P> const &x, P const t = 0)
@@ -45,7 +47,7 @@ private:
     ignore(t);
     fk::vector<P> fx(x.size());
     std::transform(x.begin(), x.end(), fx.begin(), [](P const x_v) -> P {
-      return 1.0 + 1.0e-4 * std::cos(0.5 * x_v);
+      return 1.0 + A * std::cos(0.5 * x_v);
     });
     return fx;
   }
@@ -55,13 +57,14 @@ private:
   {
     ignore(t);
 
-    P const coefficient = 1.0 / std::sqrt(2.0 * PI);
+    P const coefficient = 1.0 / std::sqrt(2.0 * PI * theta_in);
 
     fk::vector<P> fx(x.size());
-    std::transform(x.begin(), x.end(), fx.begin(),
-                   [coefficient](P const x_v) -> P {
-                     return coefficient * std::exp(-0.5 * std::pow(x_v, 2));
-                   });
+    std::transform(
+        x.begin(), x.end(), fx.begin(), [coefficient](P const x_v) -> P {
+          return coefficient *
+                 std::exp(-0.5 * (1.0 / theta_in) * std::pow(x_v, 2));
+        });
     return fx;
   }
 
@@ -126,7 +129,7 @@ private:
   {
     ignore(t);
 
-    return (1.0 + 1.0e-4 * std::cos(0.5 * x));
+    return (1.0 + A * std::cos(0.5 * x));
   }
 
   static P u(P const &x, P const t = 0)
@@ -478,7 +481,7 @@ private:
   inline static term<P> const term_i3v =
       term<P>(false,  // time-dependent
               "I3_v", // name
-              {i3_pterm_v1, i3_pterm_v2}, imex_flag::imex_explicit);
+              {i3_pterm_v1, i3_pterm_v2}, imex_flag::imex_implicit);
 
   inline static std::vector<term<P>> const terms_8 = {term_i3x, term_i3v};
 
--- a/src/pde/pde_collisional_landau.hpp
+++ b/src/pde/pde_collisional_landau.hpp
@@ -188,15 +188,13 @@ private:
     return (x > 0.0) ? x : 0.0;
   }
 
-  inline static const partial_term<P> e1_pterm_x =
-      partial_term<P>(coefficient_type::div, e1_g1, partial_term<P>::null_gfunc,
-                      flux_type::downwind, boundary_condition::periodic,
-                      boundary_condition::periodic);
+  inline static const partial_term<P> e1_pterm_x = partial_term<P>(
+      coefficient_type::div, e1_g1, nullptr, flux_type::downwind,
+      boundary_condition::periodic, boundary_condition::periodic);
 
   inline static const partial_term<P> e1_pterm_v = partial_term<P>(
-      coefficient_type::mass, e1_g2, partial_term<P>::null_gfunc,
-      flux_type::central, boundary_condition::periodic,
-      boundary_condition::periodic);
+      coefficient_type::mass, e1_g2, nullptr, flux_type::central,
+      boundary_condition::periodic, boundary_condition::periodic);
 
   inline static term<P> const term_e1x =
       term<P>(false,  // time-dependent
@@ -226,15 +224,13 @@ private:
     return (x < 0.0) ? x : 0.0;
   }
 
-  inline static const partial_term<P> e2_pterm_x =
-      partial_term<P>(coefficient_type::div, e2_g1, partial_term<P>::null_gfunc,
-                      flux_type::upwind, boundary_condition::periodic,
-                      boundary_condition::periodic);
+  inline static const partial_term<P> e2_pterm_x = partial_term<P>(
+      coefficient_type::div, e2_g1, nullptr, flux_type::upwind,
+      boundary_condition::periodic, boundary_condition::periodic);
 
   inline static const partial_term<P> e2_pterm_v = partial_term<P>(
-      coefficient_type::mass, e2_g2, partial_term<P>::null_gfunc,
-      flux_type::central, boundary_condition::periodic,
-      boundary_condition::periodic);
+      coefficient_type::mass, e2_g2, nullptr, flux_type::central,
+      boundary_condition::periodic, boundary_condition::periodic);
 
   inline static term<P> const term_e2x =
       term<P>(false,  // time-dependent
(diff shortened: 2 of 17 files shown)


src/time_advance.cpp
```diff
@@ -466,15 +467,17 @@ imex_advance(PDE<P> &pde, adapt::distributed_grid<P> const &adaptive_grid,
...
 
   auto do_poisson_update = [&](fk::vector<P> const &f_in) {
+    tools::timer.start("poisson_update");
     // Get 0th moment
     fk::vector<P> mom0(dense_size);
-    fk::vector<P> mom0_real(dense_size);
     fm::gemv(pde.moments[0].get_moment_matrix(), f_in, mom0);
-    wavelet_to_realspace<P>(pde_1d, mom0, adaptive_grid_1d.get_table(),
-                            transformer, tmp_workspace, mom0_real);
+    auto mom0_real = pde.moments[0].create_realspace_moment(
```
quantumsteve: `moment::create_realspace_moment` returns a const reference but using `auto` creates a copy.

src/time_advance.cpp
```diff
@@ -466,15 +467,17 @@ imex_advance(PDE<P> &pde, adapt::distributed_grid<P> const &adaptive_grid,
...
+        pde_1d, mom0, adaptive_grid_1d.get_table(), transformer, tmp_workspace);
     param_manager.get_parameter("n")->value = [mom0_real](P const x_v,
```
quantumsteve: Since we're making another copy in this lambda capture, should create_realspace_moment return by value and we do a move capture here?

src/time_advance.cpp
mkstoyanov: Overall the code uses `auto` rather excessively, even for thing like `int` and `double`.

`auto` is OK when used appropriately and in moderation, but it is relevant for the coders to know when something is a matrix, vector, pointer or reference. It also helps the compiler catch problems.

src/time_advance.cpp
mkstoyanov: If we change the earlier code to return a reference, then we should capture by reference here.

mkstoyanov: This is a pattern in many places.

src/time_advance.cpp
```diff
@@ -626,49 +634,58 @@ imex_advance(PDE<P> &pde, adapt::distributed_grid<P> const &adaptive_grid,
...
-  fm::gemv(pde.moments[0].get_moment_matrix(), f_2, mom0);
-  wavelet_to_realspace<P>(pde_1d, mom0, adaptive_grid_1d.get_table(),
-                          transformer, tmp_workspace, mom0_real);
-  param_manager.get_parameter("n")->value = [mom0_real](P const x_v,
-                                                        P const t = 0) -> P {
+  fm::gemv(pde.moments[0].get_moment_matrix(), x, mom0);
+  mom0_real = pde.moments[0].create_realspace_moment(
+      pde_1d, mom0, adaptive_grid_1d.get_table(), transformer, tmp_workspace);
+  param_manager.get_parameter("n")->value =
+      [&mom0_real = mom0_real](P const x_v, P const t = 0) -> P {
```
mkstoyanov: This creates the illusion of a shadow, just use `[&](P const x_v)` there's no need to say `[&mom0_real = mom0_real]`

It's repeated in several places.

src/pde/pde_collisional_landau.hpp
```diff
@@ -0,0 +1,495 @@
...
+
+  static fk::vector<P>
+  initial_condition_dim_v_0(fk::vector<P> const &x, P const t = 0)
+  {
+    ignore(t);
+
+    P const coefficient = 1.0 / std::sqrt(2.0 * PI * theta_in);
+
+    fk::vector<P> fx(x.size());
+    std::transform(
```
mkstoyanov, line 63: I really think we should stop using `transform` in these, a user has to know about iterators, lambdas and capture to understand the code. Isn't it much easier:
```
for(int i=0; i<x.size(); i++){
    fx[i] = coefficient * std::exp(-0.5 * (1.0 / theta_in) * std::pow(x[i], 2));
}
```

mkstoyanov: Don't have to change it now, but it's probably better for switch the use for the future.

src/pde/pde_collisional_landau.hpp
```diff
@@ -0,0 +1,495 @@
...
+  // source for poisson problem is -d_xx phi = n - 1 = S(n)
+  static P S(P const &y, P const t = 0)
+  {
+    ignore(t);
+    // subtracts quadrature values by one
+    return y - 1.0;
+  }
+
+  // holds the maximum absolute value of E
+  static P MaxAbsE(P const &x, P const t = 0)
```
mkstoyanov, line 166: Is this one used?

src/pde/pde_collisional_landau.hpp
```diff
@@ -0,0 +1,495 @@
...
+  static P get_dt_(dimension<P> const &dim)
+  {
+    ignore(dim);
+    /* return dx; this will be scaled by CFL from command line */
+    // return std::pow(0.25, dim.get_level());
+
+    // TODO: these are constants since we want dt always based on dim 2,
+    //  but there is no way to force a different dim for this function!
+    // (Lmax - Lmin) / 2 ^ LevX * CFL
+    return (6.0 - (-6.0)) / std::pow(2, 3);
```
mkstoyanov: Comment says ` (Lmax - Lmin) / 2 ^ LevX * CFL` but the expression is missing the CFL. I think we should update the comment, replace `std::pow(2, 3)` by `8`, and maybe comment that `2^LevX = 8`.

src/pde/pde_collisional_landau.hpp
```diff
@@ -0,0 +1,495 @@
...
+  {
+    ignore(x);
+    ignore(time);
+    return -1.0;
+  }
+
+  static P e2_g2(P const x, P const time = 0)
+  {
+    ignore(time);
+    return (x < 0.0) ? x : 0.0;
```
mkstoyanov, line 224: Is it better to say `return std::min(x, 0.0);`

Not sure, your call.

mkstoyanov (changes requested):

src/pde/pde_collisional_landau.hpp
quantumsteve: std::min requires `x` and `0.0` to be the same type. Something like the suggestion below should work.
```suggestion
    return std::min(x, P{0.0});
```

ckendrick: This is already done in the next PR: https://github.com/project-asgard/asgard/pull/512/commits/1d913b57a8c08cf8634f6d90586932eb898be361

src/pde/pde_collisional_landau.hpp
ckendrick: It is used in the poisson update in the imex advance: https://github.com/project-asgard/asgard/blob/develop/src/time_advance.cpp#L509

src/pde/pde_collisional_landau.hpp
mkstoyanov: We should not have let these linger for so long.

src/pde/pde_collisional_landau.hpp
mkstoyanov: The parameter with string ID `MaxAbsE` is used in several places, but I cannot find a place where this method is called.

src/pde/pde_collisional_landau.hpp
ckendrick: The comment above this says this will be scaled by the CFL from the command line.
I will update the comment and `pow`. This function is never used at the moment.

src/pde/pde_collisional_landau.hpp
ckendrick: The string ID gets mapped to this function [here](https://github.com/ckendrick/asgard/blob/collisional_landau_io/src/pde/pde_collisional_landau.hpp#L28). The function is wrapped in a gfunc for a partial term [here](https://github.com/ckendrick/asgard/blob/collisional_landau_io/src/pde/pde_collisional_landau.hpp#L290-L295). In the time advance we use the string ID to update the function which then gets used next time the coefficients are generated.

mkstoyanov: Yes, I missed that. Don't want to waste time, but this brings couple of questions:

* if the `MaxAbsE` parameter is never updated, then there's no point to have it as a `parameter` and it can be hard-coded in `MaxAbsE_func`
* if it is updated, then this is just an initial value and we can use a lambda expression in initialization, as opposed to making these static functions that will be used in only one iteration

ckendrick: It does get updated every time step, but yes, this function is just used to set the initial value (same for all the other `parameter` functions defined in the PDE). I agree we can do something better for initialization.

src/coefficients.cpp
```diff
@@ -19,6 +19,7 @@ void generate_all_coefficients(
     PDE<P> &pde, basis::wavelet_transform<P, resource::host> const &transformer,
     P const time, bool const rotate)
 {
+  tools::timer.start("gen_coefficients");
```
mkstoyanov, line 22: Speaking of performance, do we have an option that will turn these into no-op?

Not a big deal, but still something.

src/solver.cpp
```diff
@@ -127,17 +129,17 @@ P simple_gmres(matrix_replacement mat, fk::vector<P> &x, fk::vector<P> const &b,
   };
 
   auto const done = [](P const error, int const outer_iters,
-                       int const inner_iters) {
+                       int const inner_iters) -> gmres_info<P> {
```
mkstoyanov, line 132: Why not override the `operator <<` for the `gmres_info` struct, then we can have this output anywhere that we may need it.

Over the long run, we want ASGard to be mostly silent and we want to give the user the ability to fine-grain the CLI output. Also, `std::cout` is not exactly MPI friendly.

src/pde/pde_collisional_landau.hpp
mkstoyanov: Are we going to convert the initializations to lambdas or are we skipping it for this PR?

mkstoyanov (changes requested):
Actually, my comments are minor, so it's OK to override this and merge.

So long as we keep the show going.

src/pde/pde_collisional_landau.hpp
ckendrick: If it is ok with you, it might be best to skip for this PR. I will open an issue to fix this, but for now I think it is best to get this in

src/coefficients.cpp
ckendrick: That is a good question. As far as I know, I don't believe there is currently a way to do that. We could pull this into a compile flag that disables it, similar to how we handle `expect` statements: https://github.com/project-asgard/asgard/blob/01016aa01129e560a14093b557e2b7f2c4c17ab0/src/tools.hpp#L15-L19

I can open an issue for this.

src/solver.cpp
ckendrick: That is a good point, I will do that in a separate PR. 
I agree `std::cout` needs to either be replaced with `node_out()` or just get removed, and we can report back in `main` when needed now that we are returning all the information.
I believe this GMRES implementation was more of an exploratory work, and was not written with MPI in mind. It may be worth revisiting this entire routine for MPI.

src/coefficients.cpp
mkstoyanov: Yes, let's open an issue. You can assign it to me.

src/solver.cpp
mkstoyanov: Putting MPI in GMRES will be a challenge, but at least we won't be making it any harder with hard-coded `std::cout`.

Another PR!

mkstoyanov (approved):

mkstoyanov:
@ckendrick unless @quantumsteve object, go ahead and merge.

quantumsteve (approved):

PR #529, opened 2023-04-15
Kronmult manual

mkstoyanov:
This includes a reimplementation of the kronmult kernel for the case of 2D + linear elements.

I'm seeing 10x improvement on the reported flop rate and that's after Cole's latest improvements on the number of blocks being used. However, we should be consistent in the benchmarking and the system that we use.

@ckendrick can you test this kernel. You can turn it on and off on line 341 in src/device/kronmult_cuda.cpp


mkstoyanov:
Had to reduce the number of blocks, saves on integer operations.

mkstoyanov:
I'm seeing even better improvement for 3D. Going cubic for 3D will require changes since data now has 64 entries and I have to either change the algorithm in one of two ways:
* one warp will work on a batch entry and process it in two-three cycles
* 64 threads can work on the batch entry but will need to sync the threads across warps

ckendrick (approved):

mkstoyanov:
@quantumsteve any comments or can we merge?

quantumsteve:
Is the kronmult repo still being used? If not, I'd like to remove it here or in a subsequent PR.

mkstoyanov:
> Is the kronmult repo still being used?

Yes, the new code is not general enough to cover every N and every number of dimensions. We can keep using kronmult and keep replacing methods as we go along, but for now, we want to debug and test with the new code. It does cover important cases, e.g., 2D linear, quadratic and cubic. 3D linear and quadratic, and so on.

I want to merge this in stages, as opposed to some monstrous PR.

quantumsteve (approved):

PR #530, opened 2023-04-18
Kernel improvements

mkstoyanov:
* enabled more than 32 threads when using the kernels for quadratic basis
* cleanup for the block-threads logic and some of the variable names

PS: 3D continuity equation, 8-9 levels and quadratic-cubic basis hit the double precision flop limit of my RTX 2080Ti.

PSS: 2080Ti and 3080 have reduced double-precision capabilities, the Volta gets a teraflop from max of 5.

### DIFF
--- a/src/device/asgard_kronmult.hpp
+++ b/src/device/asgard_kronmult.hpp
@@ -66,10 +66,10 @@ void gpu2d(T const *const Aarray_[], int const lda, T const *const pX_[],
   }
   else if constexpr (n == 3)
   {
-    int constexpr num_threads = 32;
+    int constexpr num_threads = 1024;
     int num_blocks            = std::min(
-        max_blocks, (num_batch + num_threads / 9 - 1) /
-                        (num_threads / 9)); // one operation takes two threads
+        max_blocks, (num_batch + 3 * (num_threads / 32) - 1) /
+                        (3 * (num_threads / 32))); // one operation takes two threads
     kernel::gpu2d<T, num_threads, 3>
         <<<num_blocks, num_threads>>>(Aarray_, lda, pX_, pY_, num_batch);
   }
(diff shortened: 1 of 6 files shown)


ckendrick (changes requested):
This looks good. Running the two-stream full-grid level 7 case `./asgard -p two_stream -d 3 -l "7 7" -t 1.5625e-3 -f -x -n 1` on a RTX 3080 shows about a factor of 2x on the avg gflops.
On `develop`: 
```
kronmult - avg: 45.3317 min: 30.436 max: 56.404 med: 48.3250  avg gflops: 102.576722 calls: 50
```

This PR:
```
kronmult - avg: 21.4836 min: 12.449 max: 31.578 med: 22.5220  avg gflops: 219.198144 calls: 50
```

My only comment would be if we could change some of the constant integers into a name, i.e, all of the "32" into a `#define WARPSIZE 32` or something similar, just to help readability. 
Another example would be in the `gpu3d_n4` kernel, it might be helpful if the "16" constant could be named something (`MAT_SIZE`?). This might help better clarify what some numbers are vs strides and so on.

mkstoyanov:
Yes, this needs better documentation overall. One of the tricks is that I don't want threads working on the same `i` to cross the warp boundaries, otherwise I will need to synchronize.

The constants can be named better, but for the most part they correspond to lengths of cycles more so than size of matrices. I will try to explain better.

mkstoyanov:
Added some better names, but if you are trying to figure out the algorithm from the code, don't. This requires a separate document with charts and explanations. I will do that eventually, but let's do this one at a time.

I actually have multiple ideas of potential improvements, which will affect the write up, but I cannot test these until we remove the major bottleneck of using `lda`. We'll discuss this on the meeting today.

ckendrick (approved):

PR #533, opened 2023-04-21
Manual kernel4

mkstoyanov:
Many changes, here is a short list:
* consolidated the many manual kernels into a single master kernel that instantiates up to 6D quadratic basis
    * lower dimensions allow for even more basis functions, 5D can go up to cubic and so on
* added testing module for the new kronmult capabilities
* added benchmark tool, not ideal but useful
* added cleaner logic to handle special cases of thread padding and such

Overall the new kronmult module is more standalone and the code associated with the old kronmult can be carved out with relative ease. However, the CPU kernels are not done yet.


mkstoyanov:
I think the problem with the CI is that the smaller GPUs don't have enough RAM for the ridiculous amounts that kronmult is hogging.

I need to do some CPU kernels for completeness and we can switch off the old kronmult and the ridiculous workspace. We can leave it as a compile time option, but it's not necessary.

Meanwhile we can ignore the new GPUs, after @quantumsteve confirms that the issue with the available RAM.

quantumsteve:
> I think the problem with the CI is that the smaller GPUs don't have enough RAM for the ridiculous amounts that kronmult is hogging.
> 
> I need to do some CPU kernels for completeness and we can switch off the old kronmult and the ridiculous workspace. We can leave it as a compile time option, but it's not necessary.
> 
> Meanwhile we can ignore the new GPUs, after @quantumsteve confirms that the issue with the available RAM.

The smaller GPU has 24GB, which is more than Cole or I have in our office computers.

Does the `--memory` option work with the new kernels?

mkstoyanov:
> 
> The smaller GPU has 24GB, which is more than Cole or I have in our office computers.
> 
> Does the `--memory` option work with the new kernels?

The new kernels don't use any additional memory, it's only `kronmult` that's hogging data, the allocation of the workspace for the old kernels has not been turned off. The `--memory` option should work, but with 24GB that's should not be the issue.

I already fixed a bug in `fk::vector` when in Debug mode it was checking the size that had nothing to do with the new kernels, it was just about me using `fk::vector`

And it is infuriating that I don't see the full log and can only guess at the error.

src/device/asgard_kronmult_common.hpp
```diff
@@ -25,6 +30,127 @@ inline int blocks(int work_size, int work_per_block, int max_blocks)
...
+    }
+  }
+  return result;
+}
+
+/*!
+ * \brief Reference implementation of gemv, compared to BLAS alpha = beta = 1.
+ */
+template<typename T>
+void reference_gemv(int n, T const A[], T const x[], T y[])
```
quantumsteve: We already have a reference gemv implementation. 
https://github.com/project-asgard/asgard/blob/d2ccd7809b6d849378cfea59400ff2a73c6fe3b4/src/lib_dispatch.cpp#L518

mkstoyanov: I was thinking of this as a "stand-alone" package. In fact, we could even use the optimized BLAS thing here.

Let me take a look at that.

src/device/asgard_kronmult_cycle1.hpp
```diff
@@ -0,0 +1,324 @@
...
+  else if constexpr (power == 6)
+  {
+    return n * n * n * n * n * n;
+  }
+  else
+  {
+    static_assert(power >= 1 and power <= 6,
+                  "int_pow() does not works with specified power");
+    return 0;
+  }
```
quantumsteve: I _think_ this will silence some of the warnings from nvcc :crossed_fingers: 
```suggestion
  static_assert(power >= 1 and power <= 6,
                "int_pow() does not works with specified power");
  return 0;
```

mkstoyanov: I'll give it a try, but only enough CUDA 11.8 and newer does not generate the warnings.

src/device/asgard_kronmult_common.hpp
mkstoyanov: Switched to lib-dispatch

src/device/asgard_kronmult_cycle1.hpp
mkstoyanov: Fixed that and another one.

ckendrick (approved):

mkstoyanov:
@quantumsteve Can I merge?

quantumsteve:
Yes

PR #534, opened 2023-04-24
Cpu kernels

mkstoyanov:
* CPU kernels for kronmult

### DIFF
--- a/src/device/asgard_kronmult_cpu.cpp
+++ b/src/device/asgard_kronmult_cpu.cpp
@@ -45,7 +45,7 @@ void run_cpu_variant(T const *const pA[], int const lda, T const *const pX[],
   {
     if constexpr (dimensions == 1)
     {
-      T Y[n] = {0};
+      T Y[n] = {{0}};
       for (int j = 0; j < n; j++)
       {
         for (int k = 0; k < n; k++)
@@ -82,7 +82,7 @@ void run_cpu_variant(T const *const pA[], int const lda, T const *const pX[],
       }
       else if constexpr (n >= 3)
       {
-        T W[n][n] = {0}, Y[n][n] = {0};
+        T W[n][n] = {{{0}}}, Y[n][n] = {{{0}}};
         for (int j = 0; j < n; j++)
           for (int k = 0; k < n; k++)
             for (int s = 0; s < n; s++)
@@ -103,7 +103,7 @@ void run_cpu_variant(T const *const pA[], int const lda, T const *const pX[],
     }
     else if constexpr (dimensions == 3)
     {
-      T W[n][n][n] = {0}, Y[n][n][n] = {0};
+      T W[n][n][n] = {{{{0}}}}, Y[n][n][n] = {{{{0}}}};
       for (int j = 0; j < n; j++)
         for (int l = 0; l < n; l++)
           for (int k = 0; k < n; k++)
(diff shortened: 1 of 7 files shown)


src/device/asgard_kronmult_cpu.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+// always use one thread per kron-product
+#pragma omp parallel for
+  for (int i = 0; i < num_batch; i++)
+  {
+    if constexpr (dimensions == 1)
+    {
+      T Y[n];
+      for (int j = 0; j < n; j++)
+        Y[j] = 0;
```
quantumsteve:
```suggestion
      T Y[n]= {0};
```

src/device/asgard_kronmult_cpu.cpp
```diff
@@ -0,0 +1,479 @@
...
+                  pX[i][n * n * j + n * l + k] * pA[3 * i][j * lda + s];
+      for (int j = 0; j < n; j++)
+        for (int l = 0; l < n; l++)
+          for (int k = 0; k < n; k++)
+            for (int s = 0; s < n; s++)
+              W[l][s][k] += Y[l][j][k] * pA[3 * i + 1][j * lda + s];
+      for (int j = 0; j < n; j++)
+        for (int l = 0; l < n; l++)
+          for (int k = 0; k < n; k++)
+            Y[j][l][k] = 0;
```
quantumsteve:
```suggestion
      std::fill( &Y[0][0][0], &Y[0][0][0] + sizeof(Y), T{0.});)
```

src/device/asgard_kronmult_cpu_general.cpp
```diff
@@ -0,0 +1,366 @@
...
+                               ? n4
+                               : ((dimensions == 4)
+                                      ? n3
+                                      : ((dimensions == 3)
+                                             ? n2
+                                             : ((dimensions == 2) ? n : 1))))))
+  {
+    static_assert(1 <= dimensions and dimensions <= 6);
+  }
+  T &operator()(int i, int j, int k, int l, int m, int p)
```
quantumsteve, line 37: Which operator() depends on the value of `dimensions`?

src/device/asgard_kronmult_cpu_general.cpp
```diff
@@ -0,0 +1,366 @@
...
+  T &operator()(int m, int p) { return serialized[m * n + p]; }
+  T &operator()(int p) { return serialized[p]; }
+  void zero()
+  {
+    for (auto &s : serialized)
+      s = 0;
+  }
+
+private:
+  int const n5, n4, n3, n2, n;
```
quantumsteve, line 74: Do we want Tensor to be movable, copyable?

mkstoyanov: We don't care, this is a helper method inside the kernels and not a general structure that we would use elsewhere. I just need something that is allocated on the stack and can be indexed with multiple integers.

src/device/asgard_kronmult_cpu_general.cpp
mkstoyanov: Do you want to put a static assert in each operator? It's kind of obvious that 6 indexes mean 6D.

src/device/asgard_kronmult_cpu.cpp
mkstoyanov: Actually `sizeof(Y) / sizeof(T)` but a much better idea than the loop.

src/device/asgard_kronmult_cpu_general.cpp
mkstoyanov: Put in the static asserts.

src/device/asgard_kronmult_cpu_general.cpp
mkstoyanov: Sorry, allocatable on the heap.

This struct is not for a general use, that's why it's defined in a .cpp file.

quantumsteve (approved):

PR #537, opened 2023-04-27
Remove old kronmult library. Fall back to CPU implementation.

quantumsteve:

## Proposed changes

This removes the older, slower kronmult library. In the (rare?) case that a GPU kernel is not available it falls back to a CPU kernel.

- [ ] Bugfix
- [ ] New feature
- [x] Build related changes
- [x] Other (please describe): remove rarely used library  

- [x] Yes
- [ ] No

In rare cases an input that ran on the GPU may now fall back to the CPU.

my laptop, Ubuntu 20.04

## Checklist

### DIFF
--- a/src/device/kronmult_cuda.cpp
+++ b/src/device/kronmult_cuda.cpp
@@ -202,9 +202,9 @@ prepare_kronmult_kernel(int const *const flattened_table,
 
     int constexpr max_dims = 6;
     expect(num_dims <= max_dims);
-
-    // calculate and store operator row indices for this element
-    int operator_row[max_dims];
+    src / solver_tests.cpp
+          // calculate and store operator row indices for this element
+          int operator_row[max_dims];
     int const *const row_coords = flattened_table + coord_size * row;
     get_indices(row_coords, operator_row, degree, num_dims);
 
@@ -300,11 +300,6 @@ void call_kronmult(int const n, P *x_ptrs[], P *output_ptrs[],
 {
 #ifdef ASGARD_USE_CUDA
   // if a GPU kernel is not available, thow an error
-  if (!kronmult::is_implemented::gpu(num_dims, n))
-    throw std::runtime_error("Kronmult GPU kernel not implemented for degree " +
-                             std::to_string(n) + " and dimension " +
-                             std::to_string(num_dims));
-
   kronmult::execute_gpu<P>(num_dims, n, operator_ptrs, lda, x_ptrs, output_ptrs,
                            num_krons);
   auto const stat = cudaDeviceSynchronize();
--- a/src/device/kronmult_cuda.cpp
+++ b/src/device/kronmult_cuda.cpp
@@ -17,18 +17,14 @@
 #define HOST_FUNCTION
 #endif
 
-#include "kronmult1_xbatched.hpp"
-#include "kronmult2_xbatched.hpp"
-#include "kronmult3_xbatched.hpp"
-#include "kronmult4_xbatched.hpp"
-#include "kronmult5_xbatched.hpp"
-#include "kronmult6_xbatched.hpp"
-
 #include "asgard_kronmult.hpp"
 
 #ifdef ASGARD_USE_OPENMP
 #include <omp.h>
 #endif
+
+#include <cassert>
+#include <cmath>
 
 #ifdef expect
 #undef expect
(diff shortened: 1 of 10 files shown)


src/device/kronmult_cuda.cpp
```diff
@@ -303,101 +294,24 @@ void prepare_kronmult(int const *const flattened_table,
 // note  the input memory referenced by x_ptrs will be over-written
 // --------------------------------------------
 template<typename P>
-void call_kronmult(int const n, P *x_ptrs[], P *output_ptrs[], P *work_ptrs[],
+void call_kronmult(int const n, P *x_ptrs[], P *output_ptrs[],
                    P const *const operator_ptrs[], int const lda,
                    int const num_krons, int const num_dims)
 {
 #ifdef ASGARD_USE_CUDA
+  if (kronmult::is_implemented::gpu(num_dims, n))
```
mkstoyanov: I was using the old implementation as a fallback for unimplemented `n` I think we should just remove the `is_implemented` struct in `asgard_kronmult.hpp` and here we should just call `execute_gpu()`

The `execute_gpu()` will throw and error if called with incorrect  size.

If you want to use the CPU as a fallback you run into the problem that the data is already loaded on the GPU and will have to be moved back. I think it's OK to throw and in this PR or another we can have a runtime option (command line) to skip the GPU.

What do you think?

mkstoyanov:
Somewhere in ASGarD we are also reporting `workspace` size. We should remove that one too, as it is confusing. (I didn't see it removed in the PR)

quantumsteve:
> Somewhere in ASGarD we are also reporting `workspace` size. We should remove that one too, as it is confusing. (I didn't see it removed in the PR)

I removed the factor of 2. The `element_x` array still uses the workspace size. If we can remove that array, it should be in a separate PR.

src/device/kronmult_cuda.cpp
quantumsteve: :facepalm: The code now checks and throws and exception.

src/device/kronmult_cuda.cpp
```diff
@@ -303,102 +294,25 @@ void prepare_kronmult(int const *const flattened_table,
...
-
-    // -------------------------------------------
-    // note important to wait for kernel to finish
-    // -------------------------------------------
-    auto const stat = cudaDeviceSynchronize();
-    expect(stat == cudaSuccess);
-  }
+  // if a GPU kernel is not available, thow an error
+  if (!kronmult::is_implemented::gpu(num_dims, n))
+    throw std::runtime_error("Kronmult GPU kernel not implemented for degree " +
```
mkstoyanov: I wasn't clear before, `execute_gpu()` already does the throw. This block should have just the call to execute and the device synch.

The only reason why `is_implemented` was there was to send the unimplemented cases to the old kronmult so the tests wouldn't be failing while I was building the cases.

src/kronmult_tests.cpp
```diff
@@ -119,7 +119,7 @@ TEMPLATE_TEST_CASE("test kronmult", "[kronmult]", float, double)
   SECTION("6d - non-uniform level")
   {
     auto const pde_choice        = PDE_opts::continuity_6;
-    auto const degree            = 1;
+    auto const degree            = 2;
```
mkstoyanov: Why the change to the test?

src/solver_tests.cpp
```diff
@@ -231,7 +231,7 @@ TEMPLATE_TEST_CASE("test kronmult", "[kronmult]", float, double)
   SECTION("6d - non-uniform level")
   {
     auto const pde_choice        = PDE_opts::continuity_6;
-    auto const degree            = 1;
+    auto const degree            = 2;
```
mkstoyanov: Same here, why?

mkstoyanov (changes requested):

mkstoyanov:
So, ignoring the non-RAII memory management in `kronmult_tests.cpp`, do you know why they were reading the workspace back to the host?

src/kronmult_tests.cpp
quantumsteve: An exception was being thrown (no n=1 gpu implementation?)

mkstoyanov: `asgard_kronmult.cpp` has a d=6 n=1 case implemented (line 391) and there is a test in `asgard_kronmult_tests.cpp` on line 143. What's missing is the `asgard_kronmult.hpp` available in the set (number 106).

The whole `is_implemented` was an ugly hack and should go away (see the earlier comment). That should also fix the test.

src/device/kronmult_cuda.cpp
quantumsteve: removed

src/kronmult_tests.cpp
quantumsteve: `is_implemented` has been removed

quantumsteve:
> So, ignoring the non-RAII memory management in `kronmult_tests.cpp`, do you know why they were reading the workspace back to the host?

It looks like someone wanted to use an existing lambda expression instead of writing a CUDA kernel :man_shrugging:

mkstoyanov (approved):

mkstoyanov:
I want the old kronmult tests gone too, but you can leave the pleasure for me for when I come back.

PR #538, opened 2023-05-05
Reduce atomics

mkstoyanov:
* Removes the need for atomic operations on the CPU side
* Updates the benchmark to reflect the ordering of `pY`

GPU needs more work, in progress and for another PR.


src/device/asgard_kronmult_cpu.cpp
```diff
@@ -9,346 +9,352 @@ namespace asgard::kronmult
...
+          pY[i][1] += y1;
+          pY[i][2] += y2;
+          pY[i][3] += y3;
+        }
+        else if constexpr (n >= 3)
+        {
+          T W[n][n] = {{{0}}}, Y[n][n] = {{{0}}};
+          for (int j = 0; j < n; j++)
+            for (int k = 0; k < n; k++)
+              for (int s = 0; s < n; s++)
```
quantumsteve, line 94: Should these inner two loops be reversed? `W[s][k]` and `pX[i][n * j + k]` would then be accessed sequentially and `pA[2 * i][j * lda + s]` is constant for the innermost loop.

I don't see any difference in `asgard-kronmult-benchmark` :confused:

mkstoyanov: The counter argument is that hitting the column of `A` will keep it hot in cache, but the reality is that here `n` is either 3 or 4. I suspect the loops are unrolled and maybe reordered. Unless the change helps with better `avx` vectorization, it won't matter.

A better benchmark here is the `continuity_3` PDE, but I doubt you'll see anything there either.

The main advantage of the PR is the removal of the atomics and even if you're right about the order of the loops, any advantage is removed from the disconnect due the large `lda`. I'm trying to take out the big fish first, before min-maxing. Don't get me wrong, I would fix something if I could measure the difference right now or the answer is obvious one way or another, otherwise I'll wait until we do more cleaning and then do lower level testing (if we even have time).

quantumsteve (approved):

ckendrick (approved):

PR #541, opened 2023-05-14
Kronmult struct

mkstoyanov:
Another monstrous push, but this fixes most of the glaring kronmult issues:
* matrices are now stored without lda (lda is n)
    * there is a copy right now but it can be changed without kronmult changes in the future
* there is persistent kronmult-matrix object that avoids repeatedly sending the same data over
* the location of the matrices is stored in an index array, no more pointers
* X and Y are referenced by index, as opposed to a pointer
    * one matrix can be applied to multiple X and Y without reindexing/repointering
* the time stepping schemes now use persistent matrices


src/asgard_kronmult_matrix.cpp
```diff
@@ -0,0 +1,155 @@
...
+    }
+  }
+
+  fk::vector<int, mem_type::owner, resource::host> iA(
+      num_rows * num_cols * num_terms * num_dimensions);
+  fk::vector<precision, mem_type::owner, resource::host> vA(osize);
+
+#ifdef ASGARD_USE_CUDA
+  std::cout << "  kronmult dense matrix allocation (MB): "
+            << get_MB<int>(iA.size()) + get_MB<precision>(vA.size()) << "\n";
```
quantumsteve, line 56: Why is one type `int` and the other type `precision`?

src/asgard_kronmult_matrix.cpp
```diff
@@ -0,0 +1,155 @@
...
+        for (int ocol = 0; ocol < num_ops; ocol++)
+          for (int orow = 0; orow < num_ops; orow++)
+            for (int i = 0; i < kron_size; i++)
+              pA = std::copy_n(ops.data() + kron_size * orow +
+                                   lda * (kron_size * ocol + i),
+                               kron_size, pA);
+      }
+      else
+      {
+        pA = std::fill_n(pA, pde.get_coefficients(t, d).size(), 0);
```
```suggestion
        pA = std::fill_n(pA, pde.get_coefficients(t, d).size(), precision{0});
```

src/asgard_kronmult_matrix.cpp
mkstoyanov: Because `iA` stands for indexes of the `A` matrices and consists of `int` data while `vA` are the values of the `A` matrices and consists of floating point data (of the given precision).

src/device/asgard_kronmult_cpu.cpp
```diff
@@ -5,347 +5,768 @@
...
-          T Y[n][n] = {{{0}}};
+          T const *A1 = &(vA[iA[ma++]]);
+          T W[n][n] = {{{0}}}, Y[n][n] = {{{0}}};
 #pragma omp simd collapse(3)
           for (int j = 0; j < n; j++)
             for (int s = 0; s < n; s++)
               for (int k = 0; k < n; k++)
-                W[s][k] += pX[i][n * j + k] * pA[2 * i][j * lda + s];
+                W[s][k] += x[tj + n * j + k] * A1[j * n + s];
+          T const *A0 = &(vA[iA[ma++]]);
```
quantumsteve:
```suggestion
          T const * const A0 = &(vA[iA[ma++]]);
```

src/device/asgard_kronmult_cpu.cpp
```diff
@@ -5,347 +5,768 @@
...
+              y[ti + j] -= Y[j];
+            else
+              y[ti + j] += alpha * Y[j];
         }
-        else if constexpr (n >= 3)
+        else if constexpr (dimensions == 2)
         {
-          T W[n][n] = {{{0}}};
-          T Y[n][n] = {{{0}}};
+          T const *A1 = &(vA[iA[ma++]]);
```
quantumsteve:
```suggestion
          T const * const A1 = &(vA[iA[ma++]]);
```

src/device/asgard_kronmult_cpu.cpp
```diff
@@ -5,347 +5,768 @@
...
-#pragma omp simd collapse(4)
-        for (int j = 0; j < n; j++)
-          for (int s = 0; s < n; s++)
-            for (int l = 0; l < n; l++)
-              for (int k = 0; k < n; k++)
-                Y[s][l][k] +=
-                    pX[i][n * n * j + n * l + k] * pA[3 * i][j * lda + s];
+        else if constexpr (dimensions == 3)
+        {
+          T const *A2  = &(vA[iA[ma++]]);
```
quantumsteve:
```suggestion
          T const * const A2  = &(vA[iA[ma++]]);
```
and so forth...

src/device/asgard_kronmult_cpu.cpp
```diff
@@ -5,347 +5,768 @@
...
+
+/*!
+ * \brief Helper method that instantiates correct kernel based on alpha and beta.
+ */
+template<typename T>
+void cpu_n0(int const d, int const rows, int cols, int const terms,
+            int const iA[], T const vA[], T const alpha, T const x[],
+            T const beta, T y[])
+{
+  if (beta == 0)
```
quantumsteve, line 438: How safe are these floating point comparisons? Do we use the general case anywhere?

src/device/asgard_kronmult_cpu_general.hpp
```diff
@@ -0,0 +1,416 @@
...
+      for (int jx = 0; jx < num_cols; jx++)
+      {
+        // tensor i (ti) is the first index of this tensor in x
+        int const tj = jx * Y.size();
+        for (int t = 0; t < num_terms; t++)
+        {
+          if constexpr (dimensions == 1)
+          {
+            Y.zero();
+            T const *A = &(vA[iA[ma++]]);
```
quantumsteve:
```suggestion
            T const * const A = &(vA[iA[ma++]]);
```
and similar throughout this file

src/device/asgard_kronmult_cyclex.hpp
```diff
@@ -178,154 +206,193 @@ __global__ void cyclex(T const *const pA[], int const lda, T const *const pX[],
...
-          yinc += A[threadIdx.y][ia + k * n] * X[threadIdx.y][ix + k];
-        atomicAdd(&pY[i][threadIdx.x + team_size], yinc);
-      }
+      atomicAdd(&y[yoffset + threadIdx.x], yinc0);
+
+      if constexpr (num_cycles >= 2)
+        if constexpr (num_cycles == 2 and num_last_cycle == team_size)
+          atomicAdd(&y[yoffset + threadIdx.x + team_size], yinc1);
+        else if (threadIdx.x < num_last_cycle)
+          atomicAdd(&y[yoffset + threadIdx.x + team_size], yinc1);
```
quantumsteve, line 337: Is there any advantage to combining the two if conditions with an `or` vs two identical calls to `atomicAdd(...)`?

src/elements.hpp
```diff
@@ -121,7 +117,7 @@ class table
   std::unordered_map<int64_t, fk::vector<int>> id_to_coords_;
 
   // table of active elements staged for on-device kron list building
-  fk::vector<int, mem_type::owner, resource::device> active_table_d_;
+  fk::vector<int> active_table_d_;
```
quantumsteve: should this still have the `d` (device?) postfix?

src/time_advance.cpp
```diff
@@ -433,7 +456,7 @@ imex_advance(PDE<P> &pde, adapt::distributed_grid<P> const &adaptive_grid,
                        fk::vector<P, mem_type::view, resource::host>(
                            workspace, dense_size, dense_size * 2 - 1)};
 
-  auto const &table    = adaptive_grid.get_table();
+  // auto const &table    = adaptive_grid.get_table();
```
quantumsteve, line 459: commented out code?

src/time_advance.cpp
```diff
@@ -600,9 +624,13 @@ imex_advance(PDE<P> &pde, adapt::distributed_grid<P> const &adaptive_grid,
   if (pde.do_collision_operator)
   {
     // f2 now
-    pde.gmres_outputs[0] = solver::simple_gmres(
-        pde, table, program_opts, grid, f_2, x, fk::matrix<P>(), restart,
-        max_iter, tolerance, imex_flag::imex_implicit);
+    // pde.gmres_outputs[0] = solver::simple_gmres(
+    //    pde, table, program_opts, grid, f_2, x, fk::matrix<P>(), restart,
+    //    max_iter, tolerance, imex_flag::imex_implicit);
```
quantumsteve: commented out code?

src/time_advance.cpp
```diff
@@ -681,9 +708,15 @@ imex_advance(PDE<P> &pde, adapt::distributed_grid<P> const &adaptive_grid,
     // Final stage f3
     tools::timer.start("implicit_2_solve");
     fk::vector<P> f_3(x);
-    pde.gmres_outputs[1] = solver::simple_gmres(
-        pde, table, program_opts, grid, f_3, x, fk::matrix<P>(), restart,
-        max_iter, tolerance, imex_flag::imex_implicit, static_cast<P>(0.5));
+    // pde.gmres_outputs[1] = solver::simple_gmres(
```
quantumsteve, line 711: commented out code?

src/time_advance.cpp
```diff
@@ -681,9 +708,15 @@ imex_advance(PDE<P> &pde, adapt::distributed_grid<P> const &adaptive_grid,
...
+    //    pde, table, program_opts, grid, f_3, x, fk::matrix<P>(), restart,
+    //    max_iter, tolerance, imex_flag::imex_implicit, static_cast<P>(0.5));
```
quantumsteve: commented out code?

src/device/asgard_kronmult_cpu.cpp
mkstoyanov: Removed the `A2, A1, A0` variables and left one variable `A` that changes as needed.

src/device/asgard_kronmult_cpu.cpp
mkstoyanov: If there is any rounding in beta, I want to switch to the general case, i.e., if we set `beta` as in `apply(..., 0, y)` then we can zero out `y`. But if we have a `beta` that comes from other computations and is very close to `0`, then I want to use that `beta` otherwise I may introduce rounding when not needed.

As for the usage, right now we are only using `0` and `1`, but I want to change that in the future. We are doing too much repeated work and moving data back and forth, I want to have the generality to incorporate `alpha` and `beta` which may save us on calling an explicit scale method.

src/device/asgard_kronmult_cyclex.hpp
mkstoyanov: One is `if constexpr` and the other one is runtime `if`. I want to make sure that if the `constexpr` statement hold, then I never call the runtime if which is particularly bad on the GPU.

If I replace the whole thing with a single runtime `if` I would be at the mercy of the optimizer to analyze the if-statement and realize what can be done at runtime and compile time. Maybe it's OK, but I don't want to count on it.

src/elements.hpp
mkstoyanov: I should look at removing this, but there is sometimes logic that mandates a copy and the use of different variables.

Good catch!

mkstoyanov: Fixed!

src/time_advance.cpp
mkstoyanov: Removed

quantumsteve (dismissed):

mkstoyanov:
@ckendrick can you check now, perhaps this fixed it

ckendrick (changes requested):
The latest changes have fixed the issue with the Landau problem.

Everything looks good, but I'm getting tons of `-Wunused-but-set-parameter` warnings (compiled with `-DASGARD_RECOMMENDED_DEFAULTS=ON`):
```
/home/u0k/asgard/src/device/asgard_kronmult_cpu.cpp: In instantiation of ‘void asgard::kronmult::cpu_dense(int, int, int, const int*, const T*, T, const T*, T, T*) [with T = double; int dimensions = 1; int n = 2; asgard::kronmult::scalar_case alpha_case = asgard::kronmult::scalar_case::neg_one; asgard::kronmult::scalar_case beta_case = asgard::kronmult::scalar_case::zero]’:
/home/u0k/asgard/src/device/asgard_kronmult_cpu.cpp:498:66:   required from ‘void asgard::kronmult::cpu_dense(int, int, int, const int*, const T*, T, const T*, T, T*) [with T = double; int d = 1; int n = 2]’
/home/u0k/asgard/src/device/asgard_kronmult_cpu.cpp:615:25:   required from ‘void asgard::kronmult::cpu_dense(int, int, int, int, int, const int*, const T*, T, const T*, T, T*) [with T = double]’
/home/u0k/asgard/src/device/asgard_kronmult_cpu.cpp:770:41:   required from here
/home/u0k/asgard/src/device/asgard_kronmult_cpu.cpp:83:54: warning: parameter ‘alpha’ set but not used [-Wunused-but-set-parameter]
   83 |                int const iA[], T const vA[], T const alpha, T const x[],
      |                                              ~~~~~~~~^~~~~
/home/u0k/asgard/src/device/asgard_kronmult_cpu.cpp:84:24: warning: parameter ‘beta’ set but not used [-Wunused-but-set-parameter]
   84 |                T const beta, T y[])
```
Is it possible to fix this warning?

mkstoyanov:
I think I can work around it, but it is a non-warning. `alpha` is used in conjunction with a template parameter and for some values it is indeed ignored. The variable should not be considered `unused` unless it is unused in all cases but the compiler doesn't check all cases is warning preemptively. We can probably just add `ignore(alpha);` and it will be fine.

mkstoyanov:
Actually, there are no warnings if you use a newer compiler. I think I got all cases but @ckendrick will have to double-check.

src/device/asgard_kronmult_cpu.cpp
```diff
@@ -5,347 +5,770 @@
...
+ */
+template<typename T, int dimensions, int n, scalar_case alpha_case,
+         scalar_case beta_case>
+void cpu_dense(int const num_rows, int num_cols, int const num_terms,
+               int const iA[], T const vA[], T const alpha, T const x[],
+               T const beta, T y[])
 {
   static_assert(1 <= dimensions and dimensions <= 6);
   static_assert(n > 1, "n must be positive and n==1 is a special case handled "
                        "by another method");
```
ckendrick: Here is one more spot to suppress the warning.
```suggestion
  (void)alpha;
  (void)beta;
```

ckendrick (approved):

PR #546, opened 2023-05-18
different approach to lib-dispatch

mkstoyanov:
* a different approach to lib_dispatch, uses sane cblas interface and template variables to switch between device/host

### DIFF
--- a/src/lib_dispatch.cpp
+++ b/src/lib_dispatch.cpp
@@ -227,54 +227,40 @@ void rotg(P *a, P *b, P *c, P *s, resource const resrc)
   }
 }
 
-template<typename P>
-P nrm2(int *n, P *x, int *incx, resource const resrc)
-{
+template<resource resrc = resource::host, typename P>
+P nrm2(int n, P const x[], int incx){
   expect(x);
-  expect(incx && *incx >= 0);
-  expect(n && *n >= 0);
+  expect(incx >= 0);
+  expect(n >= 0);
+  static_assert(std::is_same_v<P, double> or std::is_same_v<P, float>);
 
-  if (resrc == resource::device)
-  {
-    // device-specific specialization if needed
 #ifdef ASGARD_USE_CUDA
-    // no non-fp blas on device
-    expect(std::is_floating_point_v<P>);
+  if constexpr (resrc == resource::device)
+  {
     P norm;
-    // function instantiated for these two fp types
-    if constexpr (std::is_same<P, double>::value)
+    if constexpr (std::is_same_v<P, double>)
     {
       auto const success =
-          cublasDnrm2(device.get_handle(), *n, x, *incx, &norm);
+          cublasDnrm2(device.get_handle(), n, x, incx, &norm);
       expect(success == 0);
     }
-    else if constexpr (std::is_same<P, float>::value)
+    else if constexpr (std::is_same_v<P, float>)
     {
       auto const success =
-          cublasSnrm2(device.get_handle(), *n, x, *incx, &norm);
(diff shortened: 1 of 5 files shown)


src/lib_dispatch.cpp
```diff
@@ -227,54 +227,39 @@ void rotg(P *a, P *b, P *c, P *s, resource const resrc)
...
-    else if constexpr (std::is_same<P, float>::value)
+    else if constexpr (std::is_same_v<P, float>)
     {
-      auto const success =
-          cublasSnrm2(device.get_handle(), *n, x, *incx, &norm);
+      auto const success = cublasSnrm2(device.get_handle(), n, x, incx, &norm);
       expect(success == 0);
     }
     return norm;
-#endif
```
quantumsteve, line 258: If `resrc == resource::device` and `ASGARD_USE_CUDA` is NOT defined, one should get an error (ideally at compile-time) instead of falling back to the CPU implementation.

mkstoyanov: I agree with you but the general approach of ASGarD seems to be to default to the CPU, e.g.,

https://github.com/project-asgard/asgard/blob/4cc8431a1228f5d19f2f330780bc2df76c68e31a/src/tensors.hpp#L891

I would like to change that over time.

Here we can have a `statis_assert` in fast-math and in lib-dispatch we can instantiate only when CUDA is enabled (maybe with another assert to be sure).

How does that sound?

quantumsteve: I think it would be smart to enclose the CPU code in `else if constexpr (resrc == resource::host)` so the compiler knows when it is unreachable code.

quantumsteve: Without the backup CPU implementation, several tests in `fast_math_tests.cpp` should fail.

mkstoyanov: After the latest change, `nrm2` is instantiated for the host only if CUDA is enabled and fast-math has a static-assert. Has to put `#ifdef` guards in the fast-math testing, which would not compile due to the static asserts.

I think this is the direction we should be heading for now. We don't need to go for a sweeping change, but small increments will improve this over time.

src/lib_dispatch.cpp
```diff
@@ -1213,8 +1198,11 @@ template void rotg(float *, float *, float *, float *, resource const resrc);
 template void
 rotg(double *, double *, double *, double *, resource const resrc);
 
-template float nrm2(int *n, float *x, int *incx, resource const resrc);
-template double nrm2(int *n, double *x, int *incx, resource const resrc);
+template float nrm2<resource::host, float>(int n, float const x[], int incx);
+template double nrm2<resource::host, double>(int n, double const x[], int incx);
+template float nrm2<resource::device, float>(int n, float const x[], int incx);
+template double
+nrm2<resource::device, double>(int n, double const x[], int incx);
```
quantumsteve:
```suggestion
#ifdef ASGARD_USE_CUDA
template float nrm2<resource::device, float>(int n, float const x[], int incx);
template double
nrm2<resource::device, double>(int n, double const x[], int incx);
#endif
```

src/lib_dispatch.cpp
mkstoyanov: @quantumsteve Ops, I'm a bit distracted today, I forgot to push the latest changes.

quantumsteve (approved):
clever use of `constexpr if` to avoid two functions and `std::enable_if`

mkstoyanov:
There is a way to do this without `constexpr if` but it requires `reinterpret_cast` to strip away type safety right as you're making the call to BLAS. The `constexpr if` way is naturally much better.

PR #547, opened 2023-05-23
Update lib_dispatch::rotg and lib_dispatch::copy api

quantumsteve:

## Proposed changes

Update lib_dispatch::rotg and lib_dispatch::copy following the style of #546.

- [ ] Bugfix
- [ ] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/src/lib_dispatch.cpp
+++ b/src/lib_dispatch.cpp
@@ -241,7 +241,6 @@ P nrm2(int n, P const x[], int incx)
   if constexpr (resrc == resource::device)
   {
 #ifdef ASGARD_USE_CUDA
-
     P norm;
     if constexpr (std::is_same_v<P, double>)
     {
--- a/src/lib_dispatch.cpp
+++ b/src/lib_dispatch.cpp
@@ -189,15 +189,16 @@ void initialize_libraries(int const local_rank)
 #endif
 }
 
-template<typename P>
-void rotg(P *a, P *b, P *c, P *s, resource const resrc)
+template<resource resrc, typename P>
+void rotg(P *a, P *b, P *c, P *s)
 {
-  expect(a && b && c && s);
+  expect(a);
+  expect(b);
+  expect(c);
+  expect(s);
+  static_assert(std::is_same_v<P, double> or std::is_same_v<P, float>);
 
-  // function doesn't make sense outside of FP context
-  expect(std::is_floating_point_v<P>);
-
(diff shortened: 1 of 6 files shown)


src/lib_dispatch.cpp
```diff
@@ -265,53 +268,54 @@ P nrm2(int n, P const x[], int incx)
...
+    }
+    else if constexpr (std::is_same<P, float>::value)
+    {
+      cblas_scopy(n, x, incx, y, incy);
+    }
+    else
+    {
+      for (int i = 0; i < n; ++i)
+      {
+        y[i * incy] = x[i * incx];
```
mkstoyanov, line 317: Are we ever using this? It's probably best to keep integer and floating point tricks separate, especially since one is calling accelerated backend and the other is naive implementation (i.e., no openmp or simd).

quantumsteve: I tried removing that code but found that we are using it in `fk::vector` and `fk::matrix`.

quantumsteve: https://github.com/search?q=repo%3Aproject-asgard%2Fasgard%20matrix%3Cint&type=code

mkstoyanov: OK, leave it for now.

Strided copies of integers should be avoided, contiguous copies should just use `std::copy()`. If this is doing what I think, the contiguous copy of `fk::vector` should be replaced with `std::copy()` and we should not be using `fk::matrix` to store the multi-indexes and definitely should not be doing strided copies of integers. I don't want to waste time on this right now, we'll check again once the multi-indexes are updated.

src/lib_dispatch_tests.cpp
```diff
@@ -609,14 +612,16 @@ TEMPLATE_TEST_CASE(
       int n   = x.size();
       int inc = 1;
 
-      lib_dispatch::copy(&n, x_test.data(), &inc, y_test.data(), &inc,
-                         resource::device);
+      lib_dispatch::copy<resource::device>(n, x_test.data(), inc, y_test.data(),
```
mkstoyanov, line 615: `int` on the `device` can pass only when CUDA is disabled and the device is the `cpu`

quantumsteve: It's protected by the if statement on line 607.

mkstoyanov (approved):

PR #548, opened 2023-05-23
Fix vector overflow

ckendrick:

## Proposed changes

Running a large problem such as `./asgard -p two_stream -d 3 -l "7 7" -t 1.5625e-3 -f -x -n 1` after #541 causes an overflow when allocating the kronmult `iA` vector.

This changes `fk::vector` to use `int64_t` for the vector size to prevent overflow for large vectors.

The `update_row` and `update_col` functions also need to be updated to use int64_t, but these require updating `lib_dispatch::copy` to use int64_t as well. For now, a check was added to make sure the vector size is less than INT_MAX.

- [x] Bugfix
- [ ] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist

### DIFF
--- a/src/asgard_kronmult_matrix.cpp
+++ b/src/asgard_kronmult_matrix.cpp
@@ -48,7 +48,7 @@ make_kronmult_dense(PDE<precision> const &pde,
   }
 
   fk::vector<int, mem_type::owner, resource::host> iA(
-      num_rows * num_cols * num_terms * num_dimensions);
+      int64_t{num_rows} * num_cols * num_terms * num_dimensions);
   fk::vector<precision, mem_type::owner, resource::host> vA(osize);
 
 #ifdef ASGARD_USE_CUDA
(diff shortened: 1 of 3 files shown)


mkstoyanov:
How is this possible!? A two dimensional problem with level 7 has 2^7 * 2^7 = 2^14 = 4096 cells?

The total size of `iA` is number of cells times number of dimensions (2) times number of terms.  In order to overflow you need to have 2^16 terms in the equation!

What's going on here?

ckendrick:
For this problem in gdb it is showing`num_rows` and `num_cols` = 16384, then with 4 terms and 2 dimensions, it was allocating iA to be `2^14 * 2^14 * 2^2 * 2^1 = 2^31`

mkstoyanov:
Ugh! I guess you're right, I missed that the number of cells is squared.

Nevertheless, switching to `int64_t` is probably much harder than this, the size if used all over the code. And it won't help in the end, the problem is too big for the CPU and you will run out of GPU memory.

This must be done with a sparse method instead of wasting time multiplying and adding zeros. We'll discuss in a bit.

mkstoyanov:
This particular problem is fixed in #552 by switching to a sparse vector.

The size of `iA` is a potential issue and something to consider, but even if we switch to `int64_t` the change is much more involved than the few lines in the PR. Perhaps we should start an issue.

PR #550, opened 2023-05-23
Avoid static_cast with timers

quantumsteve:

## Proposed changes

https://en.cppreference.com/w/cpp/chrono/duration/duration_cast has an example of using a floating-point duration, thus avoiding static_cast.

This actually causes `tools_test` to fail and I had to increase the precision in the report std::string to verify that there is no error.

- [ ] Bugfix
- [ ] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/src/tools.hpp
+++ b/src/tools.hpp
@@ -48,9 +48,7 @@ public:
       auto const gflops = flops / 1e9;
       expect(dur >= 0.0);
 
-      auto const gflops_per_sec = dur == 0.0
-                                      ? std::numeric_limits<double>::infinity()
-                                      : gflops / (dur * 1e-3); // to seconds
+      auto const gflops_per_sec = gflops / (dur * 1e-3); // to seconds
 
       insert(id_to_flops_, identifier, gflops_per_sec);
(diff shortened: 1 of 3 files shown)


src/tools.hpp
```diff
@@ -36,22 +36,21 @@ class simple_timer
...
     {
       expect(flops > 0);
       auto const gflops = flops / 1e9;
       expect(dur >= 0.0);
 
-      auto const gflops_per_sec =
-          dur == 0.0 ? std::numeric_limits<double>::infinity()
-                     : gflops / (static_cast<double>(dur) * 1e-6); // to seconds
+      auto const gflops_per_sec = dur == 0.0
+                                      ? std::numeric_limits<double>::infinity()
```
mkstoyanov: I don't like the `infinity` here. This is later used to compute weighted average and `infinity` is multiplied by 0, which is why it works, but it is still a mathematical error (even if it is well defined by the IEEE floating point standard). If it will be multiplied by 0, then we can have 0 in place of infinity.

mkstoyanov (approved):

PR #552, opened 2023-05-26
Sparse matrices

mkstoyanov:
* added kronmult sparse matrix format with testing and all supported sizes
* added command line switch for sparse/dense `--kron-mode sparse` or `--kron-mode dense` (default is dense)
* added quite a bit of code for detecting connectivity between nodes
* now compiling in only one precision as defined by `ASGARD_USE_DOUBLE_PREC` (keep compile time low)

As before @ckendrick could you please run some tests on the problems that are not covered in the CI.


quantumsteve:
How do we want to update CI to cover single and double precision?

mkstoyanov:
> How do we want to update CI to cover single and double precision?

I don't. I want to have the option to build with float, double or both, and the CI will do both.

The goal is to have a library with kronmult, operators, sparse grids, etc., and all in different precisions. Then each PDE will be a separate executable file, which will allow for a lot more flexibility. A user will build asgard once and then implement their own PDEs in addition to the ones that we provide.

In development, I want to build only one precision, especially since the PDEs use only one precision.

quantumsteve:
> I don't. I want to have the option to build with float, double or both, and the CI will do both.

Make a second copy of each build, one with `ASGARD_USE_DOUBLE_PREC=TRUE`, another with ``ASGARD_USE_DOUBLE_PREC=FALSE`?

mkstoyanov:
> > I don't. I want to have the option to build with float, double or both, and the CI will do both.
> 
> Make a second copy of each build, one with `ASGARD_USE_DOUBLE_PREC=TRUE`, another with ``ASGARD_USE_DOUBLE_PREC=FALSE`?

No, do it in CMake. Have a new option `ASGARD_ENABLE_PRECISION="float;double"`, it will default to both but it can be set to work with only float or only double. The CI will test the default (both) and locally we can do whatever. I want to use only one precision, since I have no need to recompile multiple precisions over and over. Kronmult takes a while to compile.

PS: This is related to the whole idea of asgard as a library as opposed to an executable with a fixed set of PDEs.

src/device/asgard_spkronmult_cyclex.hpp
```diff
@@ -0,0 +1,400 @@
...
+#include "asgard_kronmult_common.hpp"
+
+namespace asgard::kronmult::kernel
+{
+/*!
+ * \brief Variant of int_pow() that accepts the power as input, but the method is still constexpr.
+ */
+template<int n>
+__device__ constexpr int int_pow(int p)
+{
```
quantumsteve, line 12:
```suggestion
{
  static_assert(p > 1);
  static_assert(p < 7);
```

mkstoyanov: ?? p isn't know at compile time. I don't want runtime assertions in CUDA kernels.

quantumsteve: sorry, I meant `n`

quantumsteve: Oops, I saw the constexpr and assumed it was only being used at compile time...

mkstoyanov: The hope is that it will be only at compile time, the loop where this is used has fixed start/end depending on the template inputs. The hope is that the loop will be unrolled and then the method will be evaluated at compile time for each of the stages ... maybe just a hope, but this is only for the 6D cubic case and kronmult kernels are getting lower on the priority list.

src/asgard_kronmult_matrix_tests.cpp
```diff
@@ -20,39 +29,86 @@ void test_kronmult(int dimensions, int n, int num_rows, int num_terms,
...
+  }
+
 #else
-  asgard::kronmult_matrix<T> kmat(dimensions, n, num_rows, num_rows, num_terms,
-                                  std::move(iA), std::move(vA));
+
+  if (matrix_mode == sparse_mode)
+  {
+    asgard::fk::vector<int> pntr(num_rows + 1);
+    asgard::fk::vector<int> indx(num_rows * num_rows);
```
quantumsteve, line 84: (for another day)
`fk::vector<int>` currently leads to adding our own implementation of various BLAS functions...

mkstoyanov: Yes, I want to split floating point and integer logic in BLAS, but we still need a vector to store integer data on the GPU (we need the data abstraction but not the compute one).

I'll make sure to never use BLAS methods on `int` vectors.

quantumsteve (changes requested):
I recommend moving ahead with sparse matrices and moving discussion of `ASGARD_USE_DOUBLE_PREC` to a different PR. While I share your goal of reducing compilation time, more work needs to be done to maintain testing coverage.

mkstoyanov:
> I recommend moving ahead with sparse matrices and moving discussion of `ASGARD_USE_DOUBLE_PREC` to a different PR. While I share your goal of reducing compilation time, more work needs to be done to maintain testing coverage.

This is the main offending PR, adding both sparse and dense variants to the kernels effectively doubles the compilation time. Respecting the existing `ASGARD_USE_DOUBLE_PREC` option is the short term mitigation strategy.

We have not lost coverage in the line-by-line sense of the word, everything is templated and even if it is tested in only one configuration, it is still tested.

Worst case scenario, since we are testing in double precision, we temporarily lose the ability to work in single precision. It's not a big loss and it will be fixed soon.

ckendrick (approved):

PR #553, opened 2023-05-27
removed reference counts from matrix and vector

mkstoyanov:
The ref-counts with `std::shared_ptr` inside the matrix and vector class are used only in an assert statement to make sure the `view` objects have been destroyed before the memory of the `owner` is released. This can be done with a sanitizer and even then the use case is obscure as I have not seen it happen. Nevertheless, the ref-counts were enabled in `Release` mode even when the assertions are disabled, which comes at a cost with absolutely no benefit.

Removing the ref-counts speeds things by about 10% in the `continuity_3` simulation. The number of created and destroyed vectors is still huge, but his helps a tiny bit.

### DIFF
--- a/src/tensors.hpp
+++ b/src/tensors.hpp
@@ -500,9 +500,9 @@ private:
   // to this private constructor, also with a dummy variable
   template<mem_type omem, mem_type m_ = mem,
            typename = enable_for_all_views<m_>>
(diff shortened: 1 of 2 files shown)


quantumsteve (approved):
`std::shared_ptr<T>` is deceptively expensive

PR #559, opened 2023-05-31
Flexible precision

mkstoyanov:
Split the ability to build `float` and `double` precision, only lib-dispatch hasn't been adjusted but we can do that when it won't interfere with other people pushing updates there.

The PR cleans up some of the type-safety issues and makes the code faster to compile, which is handy when we grow in capabilities.


src/distribution.cpp
```diff
@@ -756,31 +756,40 @@ gather_results(fk::vector<P> const &my_results, distribution_plan const &plan,
...
-  MPI_Datatype const mpi_type =
-      std::is_same<P, double>::value ? MPI_DOUBLE : MPI_FLOAT;
+  MPI_Datatype const mpi_type = []()->MPI_Datatype
+    {
+      if constexpr (std::is_same<P, double>::value)
+        return MPI_DOUBLE;
+      else if constexpr (std::is_same<P, float>::value)
+        return MPI_FLOAT;
+      else if constexpr (std::is_same<P, int>::value)
+        return MPI_INT;
```
quantumsteve:
```suggestion
      else if constexpr (std::is_same<P, bool>::value)
        return MPI_CXX_BOOL;
```

src/distribution.cpp
```diff
@@ -1219,71 +1228,88 @@ void scatter_matrix(P *A, int *descA, P *A_distr, int *descA_distr)
...
-gather_matrix<double>(double *A, int *descA, double *A_distr, int *descA_distr);
-template void
 scatter_matrix<float>(float *A, int *descA, float *A_distr, int *descA_distr);
-template void scatter_matrix<double>(double *A, int *descA, double *A_distr,
-                                     int *descA_distr);
 #endif
+#endif
+
+template int
+get_global_max(int const my_max, distribution_plan const &plan);
```
quantumsteve:
```suggestion
template bool
get_global_max(bool const my_max, distribution_plan const &plan);
```

src/time_advance.cpp
```diff
@@ -123,11 +123,11 @@ adaptive_advance(method const step_method, PDE<P> &pde,
...
     old_size            = adaptive_grid.size();
-    auto const y_refined =
+    fk::vector<P> const y_refined =
         adaptive_grid.refine_solution(pde, y_stepped, program_opts);
-    refining = static_cast<bool>(
-        get_global_max(static_cast<float>(y_stepped.size() != y_refined.size()),
-                       adaptive_grid.get_distrib_plan()));
+    // if either one of the ranks reports 1, i.e., y_stepped.size() changed
+    refining = (get_global_max<int>((y_stepped.size() != y_refined.size()) ? 1 : 0,
+                            adaptive_grid.get_distrib_plan()) == 1);
```
```suggestion
    refining = get_global_max<bool>(y_stepped.size() != y_refined.size(), adaptive_grid.get_distrib_plan());
```

src/distribution.hpp
```diff
@@ -274,7 +274,7 @@ double get_MB(int64_t const num_elems)
 
 // find maximum element in plan using each rank's local max
 template<typename P>
-P get_global_max(P const my_max, distribution_plan const &plan);
+P get_global_max(no_deduce<P> const my_max, distribution_plan const &plan);
```
quantumsteve: With a boolean specialization, I think we can go back to template argument deduction.

mkstoyanov: Yes, we have to constrain the type to the ones we have linked to the MPI types, but we can auto-deduce here. I'll fix it.

src/adapt_tests.cpp
```diff
@@ -79,9 +79,9 @@ void test_adapt(parser const &problem, std::filesystem::path gold_base)
   };
 
   auto const gold_coarse_table =
-      fk::matrix<int>(read_matrix_from_txt_file(table_coarse_path));
+      fk::matrix<int>(read_matrix_from_txt_file<int>(table_coarse_path));
```
quantumsteve:
```suggestion
      read_matrix_from_txt_file<int>(table_coarse_path);
```
We should skip the additional move constructor.

mkstoyanov: I missed a spot?

src/adapt_tests.cpp
```diff
@@ -49,11 +49,11 @@ void test_adapt(parser const &problem, std::filesystem::path gold_base)
   auto const table_coarse_path = gold_base / (prefix + "coarse_table.dat");
 
   auto const fval_orig =
-      fk::vector<P>(read_vector_from_txt_file(fval_orig_path));
+      fk::vector<P>(read_vector_from_txt_file<P>(fval_orig_path));
   auto const gold_coarse =
-      fk::vector<P>(read_vector_from_txt_file(fval_coarse_path));
+      fk::vector<P>(read_vector_from_txt_file<P>(fval_coarse_path));
```
quantumsteve:
```suggestion
      read_vector_from_txt_file<P>(fval_coarse_path);
```
`read_vector_from_txt_file` also has an extra unnecessary move constructor.

src/adapt_tests.cpp
```diff
@@ -79,9 +79,9 @@ void test_adapt(parser const &problem, std::filesystem::path gold_base)
   };
 
   auto const gold_coarse_table =
-      fk::matrix<int>(read_matrix_from_txt_file(table_coarse_path));
+      read_matrix_from_txt_file<int>(table_coarse_path);
   auto const gold_refine_table =
-      fk::matrix<int>(read_matrix_from_txt_file(table_refine_path));
+      fk::matrix<int>(read_matrix_from_txt_file<int>(table_refine_path));
```
quantumsteve, line 82: another extra move constructor

src/adapt_tests.cpp
quantumsteve: We don't need to call the extra move constructor
https://github.com/project-asgard/asgard/blob/2d8a205ce2a63b145bba81d31f934373471ebed7/src/tensors.hpp#L2066

src/adapt_tests.cpp
mkstoyanov: I'll have to take another look at these.

src/distribution.cpp
```diff
@@ -756,31 +756,40 @@ gather_results(fk::vector<P> const &my_results, distribution_plan const &plan,
...
+      else if constexpr (std::is_same<P, float>::value)
+        return MPI_FLOAT;
+      else if constexpr (std::is_same<P, bool>::value)
+        return MPI_CXX_BOOL;
+      else
+        static_assert(std::is_same<P, double>::value, "The value of P must be double, float, or int");
+    }();
 
   P global_max;
   success = MPI_Allreduce(&my_max, &global_max, 1, mpi_type, MPI_MAX,
```
quantumsteve: Since `MPI_MAX` isn't defined for `MPI_CXX_BOOL` I think we'd have to use `MPI_LOR` to get the same effect.

Using an integer type and converting back to bool may be cleaner :man_shrugging:

src/adapt_tests.cpp
mkstoyanov: I think I got them all this time, used some global search.

src/distribution.hpp
mkstoyanov: Reverted to auto-deduce.

src/distribution.cpp
mkstoyanov: The can be handled with a `constexpr if`, it's clean enough except the method is badly named. In fact, it was badly named to begin, as it does not search for the global but for the max within the ranks that are associated with this set of rows ... MPI needs cleaning and it is wrong to create-delete comms every time this method is called.

Leave it for now, we'll deal with it once we get to MPI.

mkstoyanov:
@quantumsteve Are we ready to merge this one?

quantumsteve (approved):

PR #564, opened 2023-06-01
Fix tensor const correct

### LINKED ISSUE #467: fk::matrix<P> const and fk::vector<P> return mutable pointer.
**Describe the bug**

fk::matrix<P> const and fk::vector<P> return P* which can be modified. 

https://github.com/project-asgard/asgard/blob/410ca628d5d88da3989bac261f0edf9afde0176c/src/tensors.hpp#L408
https://github.com/project-asgard/asgard/blob/410ca628d5d88da3989bac261f0edf9afde0176c/src/tensors.hpp#L408

**To Reproduce**
1. git commit hash being built
2. cmake command
3. full program/test invocation command
4. additional steps 

A_copy is declared const [here](https://github.com/project-asgard/asgard/blob/develop/src/fast_math_tests.cpp#L802) but is modified by fm::gesv and therefore different on [line 810](https://github.com/project-asgard/asgard/blob/develop/src/fast_math_tests.cpp#L810)

**Expected behavior**

Same behavior as std::vector::data()
https://en.cppreference.com/w/cpp/container/vector/data

**System:**
 - system name [e.g. fusiont5, summit]
 - modules loaded [e.g. output of `module list`]
 - other systems where this is reproducible [e.g. "my laptop", "none"]

**Additional context**

mkstoyanov:

## Proposed changes

Closes #467

Makes the tensors const-correct by properly overloading the `.data()` method.

Removed unnecessary code: `batch_gemv` and the raw-pointers were not used.

- [x] Bugfix
- [ ] New feature
- [ ] Build related changes

- [x] Yes
- [ ] No

Ubuntu 22.04

## Checklist

- [x] documentation has been added (if appropriate)


mkstoyanov:
@quantumsteve finally we can fix this one, it should make your life a bit easier when fixing lib-dispatch and it will feed well into some of the ideas that I have of  improving the creation/deletion of transients.

src/lib_dispatch.cpp
```diff
@@ -1052,13 +951,17 @@ void getrs(char *trans, int *n, int *nrhs, P *A, int *lda, int *ipiv, P *b,
   expect(*ldb >= 1);
   expect(*lda >= 1);
   expect(*n >= 0);
+
+  // the const_cast below is needed due to bad header under OSX
```
quantumsteve, line 955: We might try defining `ACCELERATE_NEW_LAPACK` in another PR and see what happens on MacOS.

https://developer.apple.com/documentation/accelerate/blas

quantumsteve (approved):
:tada:

src/lib_dispatch.cpp
mkstoyanov: Maybe worth a try, but it reads more like they just have 64 bit indexing, which can be a problem if we pass in the sizes as pointers (compiler will automatically convert if we pass by value)

PR #568, opened 2023-06-08
reading wrong terms in dense matrices and anisotropic grids

mkstoyanov:
If the total number of coefficients is not the same for all terms, the dense matrix will yield a bug.


quantumsteve (approved):
Let's try creating a quickly running test that catches this oversight.

mkstoyanov:
> Let's try creating a quickly running test that catches oversight.

Let's discuss that on the next big meeting, make sure you bring it up.

This problem will appear when the max-level for different terms is different, which may be something that never happens.

PR #571, opened 2023-06-15
Remove device fallback to CPU

quantumsteve:

## Proposed changes

This PR removes a confusing and unused feature that would allocate "device" vectors and matrices on the host CPU if CUDA isn't enabled. This feature is only used in a handful of tests.

- [ ] Bugfix
- [ ] New feature
- [ ] Build related changes

- [x] Yes
- [ ] No

This removes a CPU fallback that appears to only be used in test code. 

my laptop, Ubuntu 20.04

## Checklist

### DIFF
--- a/src/tensors.hpp
+++ b/src/tensors.hpp
@@ -1009,12 +1009,12 @@ fk::vector<P, mem, resrc>::vector(std::initializer_list<P> list)
 {
   if constexpr (resrc == resource::host)
   {
-    data_ = new P[size_]();
+    data_ = new P[size_];
     std::copy(list.begin(), list.end(), data_);
   }
   else
   {
-    allocate_device(data_, size_);
+    allocate_device(data_, size_, false);
     copy_to_device(data_, list.begin(), size_);
   }
 }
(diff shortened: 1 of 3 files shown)


src/tensors.hpp
```diff
@@ -822,14 +822,10 @@ allocate_device(P *&ptr, int64_t const num_elems, bool const initialize = true)
...
-    ptr = new P[num_elems]();
-  }
-  else
-  {
-    ptr = new P[num_elems];
-  }
+  ignore(ptr);
+  ignore(num_elems);
+  ignore(initialize);
+  expect(false);
```
mkstoyanov: I think `throw` with an explanation is better than the non-informative `excpect(false);`

In fact, why don't we break the memory operations into two blocks:
```
#ifdef ASGARD_USE_CUDA
template<typename P>
inline void
allocate_device( ... )
#else
template<typename P>
inline void
allocate_device(P *&, int64_t const, bool const = true){
  throw std::runtime_error("calling allocate_device without CUDA");
}
#endif
```

We can do that for the few methods and later (in another PR) we can move those to a separate file. That will make it easier to move to ROCM by simply including a different file.

mkstoyanov (approved):

PR #572, opened 2023-06-16
Kronmult limited memory

mkstoyanov:
Kronmult now automatically splits the problem into chunks to avoid overflow of the 32-bit integer. This is done on the CPU and GPU, sparse and dense cases. This will not limit the total amount of GPU memory being used, but it will allow handling larger problems, should the limitation be the index size as opposed to the memory size.

There is also a CMake option to enable the limit for GPU memory, but the method is currently too slow. Doing a proper out-of-core algorithm requires overlapping communication and computation the kronmult module, which in turn requires dabbling with pinned memory. Even then, for smaller dimensions and degree, the cost of data movement is much higher than the gains from fast GPU.

Improvements will follow in further PRs.


CMakeLists.txt
```diff
@@ -98,6 +98,7 @@ option (ASGARD_PROFILE_PERF "enable profiling support for using linux perf" "")
 option (ASGARD_PROFILE_VALGRIND "enable profiling support for using valgrind" "")
 option (ASGARD_GRAPHVIZ_PATH "optional location of bin/ containing dot executable" "")
 option (ASGARD_USE_CUDA "Optional CUDA support for asgard" OFF)
+option (ASGARD_USE_GPU_MEM_LIMIT "Allow the ability to limit the GPU memory used by kronmult (can hurt performance)" OFF)
```
quantumsteve: What's the best way to add this to CI? Create a new CI build with `ASGARD_USE_GPU_MEM_LIMIT=ON`?

mkstoyanov: Yes, it must use CUDA and `-DASGARD_USE_GPU_MEM_LIMIT=ON`

By default, we should test with off.

mkstoyanov:
@quantumsteve BTW we should move on this PR before we start merging the rest of the two-stream and Cole's work. That's because this PR makes the workaround for the 32-bit indexing limitation and potential GPU memory limit.

The limit itself is not so important, currently it imposes a huge performance penalty and more improvements will be needed over time. But the 32-bit indexing is important to fix.

CMakeLists.txt
```diff
@@ -119,6 +120,10 @@ if (ASGARD_RECOMMENDED_DEFAULTS)
   endif ()
 endif ()
 
+if (ASGARD_USE_GPU_MEM_LIMIT AND NOT ASGARD_USE_CUDA)
```
quantumsteve, line 123: `cmake_dependent_option` can makes these checks.

src/asgard_grid_1d.hpp
```diff
@@ -0,0 +1,198 @@
...
+      {
+        send = current - 1;
+      }
+      else
+      {
+        return current;
+      };
+      current = (sstart + send) / 2;
+    }
+    return -1;
```
quantumsteve, line 176: Is this expected to occur?

src/asgard_kronmult_matrix.hpp
```diff
@@ -169,23 +313,186 @@ class kronmult_matrix
...
+      {
+#ifdef ASGARD_USE_GPU_MEM_LIMIT
+        // multiple calls, need to move data, call kronmult, then move next data
+        // data loading is done asynchronously using the load_stream
+        int *load_buffer    = worka.data();
+        int *compute_buffer = workb.data();
+        auto stats          = cudaMemcpyAsync(load_buffer, list_iA[0].data(),
+                                     sizeof(int) * list_iA[0].size(),
+                                     cudaMemcpyHostToDevice, load_stream);
+        assert(stats == cudaSuccess);
```
quantumsteve:
```suggestion
        expect(stats == cudaSuccess);
```

src/asgard_kronmult_matrix.hpp
```diff
@@ -169,23 +313,186 @@ class kronmult_matrix
...
+            cudaStreamSynchronize(nullptr);
+          std::swap(load_buffer, compute_buffer);
+
+          if (i + 1 < list_iA.size())
+          {
+            // begin loading the next chunk of data
+            stats = cudaMemcpyAsync(load_buffer, list_iA[i + 1].data(),
+                                    sizeof(int) * list_iA[i + 1].size(),
+                                    cudaMemcpyHostToDevice, load_stream);
+            assert(stats == cudaSuccess);
```
quantumsteve:
```suggestion
            expect(stats == cudaSuccess);
```

src/asgard_kronmult_matrix.hpp
quantumsteve: I think we need a dedicated check for compiler warnings with CUDA enabled.

src/asgard_kronmult_matrix.hpp
```diff
@@ -169,23 +313,186 @@ class kronmult_matrix
...
+#endif
+      }
+    }
     fk::copy_to_host(y, ydev.data(), ydev.size());
 #else
     if (is_dense())
-      kronmult::cpu_dense(num_dimensions_, kron_size_, num_rows_, num_cols_,
-                          num_terms_, iA.data(), vA.data(), alpha, x, beta, y);
+    {
+      if (iA.size() > 0)
```
quantumsteve, line 465: NOT for this PR, but `fk::vector` and `fk::matrix` would benefit from an `empty()` member function.

src/asgard_kronmult_matrix.hpp
```diff
@@ -244,7 +554,67 @@ class kronmult_matrix
...
+
+    expect((row_indx_.size() == 0 and col_indx_.size() == 0) or
+           (row_indx_.size() > 0 and col_indx_.size() > 0));
+
+    tensor_size_ = compute_tensor_size(num_dimensions_, kron_size_);
+
+    flops_ = 0;
+    for (auto const &a : list_iA)
+      flops_ += static_cast<int64_t>(a.size());
+    flops_ *= int64_t(tensor_size_) * kron_size_;
```
quantumsteve:
```suggestion
    flops_ *= int64_t{tensor_size_} * kron_size_;
```

quantumsteve (changes requested):
I just updated `AWS CodeBuild us-east-1 (asgard-unit-cuda)` to set `-DSGARD_USE_GPU_MEM_LIMIT=1`. I reran the job and it passes :tada:

Just a couple small changes to avoid unused variable warnings in Release builds. 

`cmake_dependent_option` is a nice way to handle options that depend on other options being set.

src/asgard_grid_1d.hpp
mkstoyanov: If two cells are connected, this will return the location of the non-zero within the sparse 1D matrix. If the cells are not connected, this will return invalid index `-1`. Usage is:
```
int idx = rule1d.get_offset(row, col);
if (idx == -1)
  // process disconnected cells
else
  // process connected cells
```

src/asgard_kronmult_matrix.hpp
mkstoyanov: Agree!

src/asgard_kronmult_matrix.hpp
mkstoyanov: You mean runtime check?

CMakeLists.txt
mkstoyanov: `cmake_dependent_option` does not raise error, it only hides option (probably just in the GUI).

Switched to the `cmake_dependent_option` but the if-statement must remain.

quantumsteve: :disappointed: Wonder if it's a CMake bug...

src/asgard_kronmult_matrix.hpp
quantumsteve: The `expect` macro should expand to (void)stats; so that stats doesn't look like an unused variable.

quantumsteve (approved):

PR #576, opened 2023-06-28
Add twostream PDE test

ckendrick:

## Proposed changes

This adds a test for the twostream PDE that runs a FG level 5 for 15 time steps. The moments and total relative energy is checked, which is sufficient to check for implementation issues.

- [ ] Bugfix
- [ ] New feature
- [ ] Build related changes
- [x] Other (please describe): test for existing PDE

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/src/time_advance.cpp
+++ b/src/time_advance.cpp
@@ -439,18 +439,17 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
   static bool first_time = true;
 
   // create 1D version of PDE and element table for wavelet->realspace mappings
-  static PDE pde_1d = PDE(pde, PDE<P>::extract_dim0);
-  static adapt::distributed_grid adaptive_grid_1d(pde_1d, program_opts);
+  PDE pde_1d = PDE(pde, PDE<P>::extract_dim0);
+  adapt::distributed_grid adaptive_grid_1d(pde_1d, program_opts);
 
   // Create workspace for wavelet transform
-  static auto const dense_size = dense_space_size(pde_1d);
-  static fk::vector<P, mem_type::owner, resource::host> workspace(dense_size *
-                                                                  2);
-  static std::array<fk::vector<P, mem_type::view, resource::host>, 2>
-      tmp_workspace = {fk::vector<P, mem_type::view, resource::host>(
-                           workspace, 0, dense_size - 1),
-                       fk::vector<P, mem_type::view, resource::host>(
-                           workspace, dense_size, dense_size * 2 - 1)};
+  auto const dense_size = dense_space_size(pde_1d);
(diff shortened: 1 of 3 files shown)


mkstoyanov:
The test seems to be failing in single precision. So, the first question is whether the test even makes sense in single precision or if we should restrict it to double. Maybe it's as simple as adjusting the tolerances. 

Also, do we need to do the 600 steps, before the bug-fix, the test was failing on the second step.

ckendrick:
Yes it looks like single precision is the problem now. The tolerances I chose for single precision are just a guess - I haven't tried to run the two-stream PDE in single precision mode. Whether we want to support this, I'm not sure.

I'm perfectly happy with just changing the IMEX tests to be `double` only if no one has an issue with that.

As far as running to 600 steps, I am conflicted. Even before the bug fix, the solutions looked fine visually until ~step 600. The energy checks added caught the issue early-on, but I feel like it may be helpful to have a test that runs for a lot of iterations to help catch a future false positive. For instance, when approving your PRs, I ran this test for about 100 steps and visually it looked fine to me - which was why we didn't detect the bug until later.
This test also checks the moment integrals which detected problems starting around 500-600 steps, so I feel like it is another good indicator in addition to energy.

All of our other PDE tests are run for 10 steps or less, so it might be beneficial to have "longer" runs. Then again, we don't want to add unnecessary time to all of the CI. I believe this runs fairly quickly, but we will (very soon) be adding more of these tests for all of the 1x2v, 1x3v PDEs added with and without adaptivity. Most of those can be run for ~10 iterations, but they will take on the order of 10 minutes per test.

@mkstoyanov @quantumsteve One possibility: we could pull the "body" of these IMEX tests into a function and define two tests - a short one that is permanently enabled and runs for 10-20 steps. The second test could be the full 600 step test that is conditionally enabled behind a CMake build option (`ASGARD_LONG_TESTS` or similar). We can add a new CI pipeline that enables only the longer tests.

mkstoyanov:
A question for @stefan-schnake and @endeve is whether we can cook up a forcing term to the two-stream problem so that we would know an exact solution. The forcing term can be enabled only for testing, but it will allow us to catch problems early.

Visually the difference did not appear until step 500, but even after the first step, the solution was off by 1.E-6, which is too much for double precision and even more if we look at the time-step. The delay in visual breakup was simply because of the small time-step. Other PDEs (the ones that use 10 steps) actually compare to a "gold" solution. I don't want to generate a "gold" file for this PDE, I'm not fan of the approach, but an analytic/exact solution will work.

ckendrick:
@mkstoyanov I temporarily decreased the number of steps this test uses from 600 down to 20 so this doesn't keep holding everything else up.

I opened issue #587 where we can discuss what the best solution for handling longer CI tests would be. We can come back and modify this test later on.

mkstoyanov (approved):

PR #581, opened 2023-07-11
Add matrix-free GMRES solver test

quantumsteve:

## Proposed changes

This adds a matrix-free gmres solver test for use as we move the vectors `x` and `b` to the GPU and avoid unnecessary copies.

- [ ] Bugfix
- [ ] New feature
- [ ] Build related changes
- [x] Other (please describe): testing improvement

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/src/solver.cpp
+++ b/src/solver.cpp
@@ -24,22 +24,23 @@ simple_gmres(fk::matrix<P> const &A, fk::vector<P> &x, fk::vector<P> const &b,
                       tolerance);
 }
 
-template<typename P>
+template<typename P, resource resrc>
 gmres_info<P>
-simple_gmres_euler(const P dt, kronmult_matrix<P> const &mat, fk::vector<P> &x,
-                   fk::vector<P> const &b, int const restart,
-                   int const max_iter, P const tolerance)
+simple_gmres_euler(const P dt, kronmult_matrix<P> const &mat,
+                   fk::vector<P, mem_type::owner, resrc> &x,
+                   fk::vector<P, mem_type::owner, resrc> const &b,
+                   int const restart, int const max_iter, P const tolerance)
 {
   return simple_gmres(
-      [&](fk::vector<P> const &x_in, fk::vector<P> &y, P const alpha,
+      [&](fk::vector<P, mem_type::owner, resrc> const &x_in,
+          fk::vector<P, mem_type::owner, resrc> &y, P const alpha,
           P const beta) -> void {
-        mat.apply(-dt * alpha, x_in.data(), beta, y.data());
+        mat.template apply<resrc>(-dt * alpha, x_in.data(), beta, y.data());
         int one = 1, n = y.size();
-        lib_dispatch::axpy(n, alpha, x_in.data(), one, y.data(), one);
+        lib_dispatch::axpy<resrc>(n, alpha, x_in.data(), one, y.data(), one);
       },
       x, b, fk::matrix<P>(), restart, max_iter, tolerance);
 }
-
 /*! Generates a default number inner iterations when no use input is given
  * \param num_cols Number of columns in the A matrix.
  * \returns default number of iterations before restart
@@ -56,10 +57,13 @@ int default_gmres_restarts(int num_cols)
                     maximum);
 }
 
+static int pos_from_indices(int i, int j) { return i + j * (j + 1) / 2; }
+
 // simple, node-local test version
-template<typename P, typename matrix_replacement>
+template<typename P, typename matrix_replacement, resource resrc>
 gmres_info<P>
-simple_gmres(matrix_replacement mat, fk::vector<P> &x, fk::vector<P> const &b,
+simple_gmres(matrix_replacement mat, fk::vector<P, mem_type::owner, resrc> &x,
+             fk::vector<P, mem_type::owner, resrc> const &b,
              fk::matrix<P> const &M, int restart, int max_iter, P tolerance)
 {
   if (tolerance == parser::NO_USER_VALUE_FP)
@@ -109,7 +113,7 @@ simple_gmres(matrix_replacement mat, fk::vector<P> &x, fk::vector<P> const &b,
     return (norm == 0.0) ? static_cast<P>(1.0) : norm;
   }();
 
-  fk::vector<P> residual(b);
+  fk::vector<P, mem_type::owner, resrc> residual(b);
   auto const compute_residual = [&]() {
     P const alpha = -1.0;
     P const beta  = 1.0;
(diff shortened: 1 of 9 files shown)


mkstoyanov:
Kronmult doesn’t do gpu to gpu work yet, everything in and out is on the cpu. It’s easy to make, but it’s not there yet. This may be jumping the gun a bit.

Also, lib-dispatch no longer needs the definition of constants one or n.

quantumsteve:
> Kronmult doesn’t do gpu to gpu work yet, everything in and out is on the cpu. It’s easy to make, but it’s not there yet. This may be jumping the gun a bit.

ColeGarD  seems to do gpu to gpu work. 
https://github.com/ckendrick/asgard/blob/273ce79264497fcde1077fd1ac655f789fcd97d9/src/asgard_kronmult_matrix.hpp#L309

src/asgard_kronmult_matrix.hpp
```diff
@@ -304,13 +304,26 @@ class kronmult_matrix
    *
    * This method is not thread-safe!
    */
+  template<resource input_resrc  = resource::host,
```
mkstoyanov: Make one template parameters, mix-and-match gets messy.
```
template<resource resrc  = resource::host>
```

src/asgard_kronmult_matrix.hpp
```diff
@@ -304,13 +304,26 @@ class kronmult_matrix
...
+           resource output_resrc = resource::host>
   void apply(precision alpha, precision const x[], precision beta,
              precision y[]) const
   {
 #ifdef ASGARD_USE_CUDA
-    if (beta != 0)
-      fk::copy_to_device(ydev.data(), y, ydev.size());
-    fk::copy_to_device(xdev.data(), x, xdev.size());
+    if constexpr (input_resrc == resource::host)
```
mkstoyanov: My expectation here is that this will work fine:
```
  void apply(precision alpha, precision const x[], precision beta,
             precision y[]) 
{
if constexpr (resrc == resource::host){
#ifdef ASGARD_USE_CUDA
      if (beta != 0)
        fk::copy_to_device(ydev.data(), y, ydev.size());
      fk::copy_to_device(xdev.data(), x, xdev.size());
      apply<resource::device>(alpha, xdev.data(), beta, ydev.data());
      fk::copy_to_host(y, ydev.data(), ydev.size());
#else 
    ... CPU kronmult code ...
#endif
} else {
#ifdef ASGARD_USE_CUDA
    ... CPU kronmult code ...
#else 
    static_assert(resrc == resource::host, "kronmult cannot work on device if CUDA is not enabled ...");
#endif
}
}

```

src/solver.cpp
```diff
@@ -109,17 +112,29 @@ simple_gmres(matrix_replacement mat, fk::vector<P> &x, fk::vector<P> const &b,
...
-      precond_factored = true;
+      if constexpr (resrc == resource::device)
+      {
+#ifdef ASGARD_USE_CUDA
+        auto res = residual.clone_onto_host();
+        precond_factored ? fm::getrs(precond, res, precond_pivots)
+                         : fm::gesv(precond, res, precond_pivots);
+        fk::copy_vector(residual, res);
+        precond_factored = true;
+#endif
```
mkstoyanov, line 133: Throw some sort of error or static assert if this is called on the device without CUDA

mkstoyanov: There is a bunch of these

src/solver.cpp
```diff
@@ -149,32 +164,46 @@ simple_gmres(matrix_replacement mat, fk::vector<P> &x, fk::vector<P> const &b,
...
+        {
+          fm::getrs(precond, new_basis, precond_pivots);
+        }
       }
 
       for (int k = 0; k <= i; ++k)
       {
-        fk::vector<P, mem_type::const_view> const basis_vect(basis, k, 0,
-                                                             basis.nrows() - 1);
+        fk::vector<P, mem_type::const_view, resrc> const basis_vect(
```
mkstoyanov: This is very inefficient, I must have missed it in the original PR. We can change it later, but new-basis should be just a view of the i+1-st column of basis, and the for-loop should be replaced with `gemv` operations.

Look here: (it's not ideal, but it is better)
https://github.com/mkstoyanov/hala/blob/develop/hex/solvers/hala_solvers_gmres.hpp

src/solver.cpp
```diff
@@ -217,9 +249,13 @@ simple_gmres(matrix_replacement mat, fk::vector<P> &x, fk::vector<P> const &b,
...
     std::vector<int> pivots(restart);
     fm::gesv(proj, s_view, pivots);
-    x = x + (fk::matrix<P, mem_type::view>(basis, 0, basis.nrows() - 1, 0,
-                                           restart - 1) *
-             s_view);
+
+    fk::matrix<P, mem_type::view, resrc> m(basis, 0, basis.nrows() - 1, 0,
+                                           restart - 1);
+    if constexpr (resrc == resource::device)
+      fm::gemv(m, s_view.clone_onto_device(), x, false, P{1.0}, P{1.0});
```
mkstoyanov, line 263: The signature of the fast-math methods is messed up .... for another PR.

mkstoyanov (changes requested):

src/solver.cpp
quantumsteve: I made some progress changing from dot products to gemv. I'd like to keep additional changes to separate pull requests.

src/solver.cpp
quantumsteve: done

src/asgard_kronmult_matrix.hpp
quantumsteve: Think I found a solution using lambdas that required minimal changes to apply

src/asgard_kronmult_matrix.hpp
```diff
@@ -353,22 +378,22 @@ class kronmult_matrix
...
-          kronmult::gpu_dense(
-              num_dimensions_, kron_size_, output_size(),
-              list_iA[i].size() / (num_dimensions_ * num_terms_), num_cols_,
-              num_terms_, list_iA[i].data(), vA.data(), alpha, xdev.data(),
-              (i == 0) ? beta : 1,
-              ydev.data() + i * list_row_stride_ * tensor_size_);
+          kronmult::gpu_dense(num_dimensions_, kron_size_, output_size(),
+                              list_iA[i].size() /
+                                  (num_dimensions_ * num_terms_),
+                              num_cols_, num_terms_, list_iA[i].data(),
```
mkstoyanov: In out-of-core mode (e.g., if we limit the memory), the entries of `list_iA` sit on the CPU.

Perhaps we should wait for the GPU-to-GPU PR first and not break use-cases.

mkstoyanov (approved):

PR #588, opened 2023-07-18
Add option --max_adapt_level

### LINKED ISSUE #583: Add different max levels for adaptivity
When adaptivity is enabled, only a single maximum level can be specified that applies to all dimensions.  However, in many cases the maximum level for adaptivity is dependent on the dimension (e.g. x vs v) for CFL constraints.  

The ability to pass a `--max_adapt_level "2 4 4"` that only refines each dimension up to the given level.  This should(TM) be a quick fix.

**Additional context**
I would not change `max_level` to a vector.  `max_level` is used all throughout the code (ex: defining the hierarchy index). 

Relevant changes should be made [here](https://github.com/project-asgard/asgard/blob/63ba4be868bf889b4ad1dda4698b2340783b97a5/src/pde/pde_base.hpp#L820-L846) to set `max_level` from the command line options and @ckendrick has a commit [here](https://github.com/ckendrick/asgard/commit/d5a6680ed9a244fac906f0f0cde71de4a830cbd5) hacks in a specific max adaptivity level vector for a particular problem.

quantumsteve:

## Proposed changes

This adds the option `--max_adapt_level` which provided finer control than the current `-max_level`

#fixed #583

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/src/elements.cpp
+++ b/src/elements.cpp
@@ -222,20 +222,22 @@ table::get_child_elements(int64_t const index, options const &opts) const
   std::list<int64_t> daughter_ids;
   for (auto i = 0; i < num_dims; ++i)
   {
-    int level = max_adapt_levels.empty() ? opts.max_level : max_adapt_levels[i];
     // first daughter in this dimension
+    int level = max_adapt_levels.empty() ? opts.max_level : max_adapt_levels[i];
     if (coords(i) + 1 < level)
     {
       auto daughter_coords          = coords;
       daughter_coords(i)            = coords(i) + 1;
       daughter_coords(i + num_dims) = coords(i + num_dims) * 2;
-      daughter_ids.push_back(map_to_id(daughter_coords, level, num_dims));
+      daughter_ids.push_back(
+          map_to_id(daughter_coords, opts.max_level, num_dims));
 
       // second daughter
       if (coords(i) >= 1)
       {
         daughter_coords(i + num_dims) = coords(i + num_dims) * 2 + 1;
-        daughter_ids.push_back(map_to_id(daughter_coords, level, num_dims));
+        daughter_ids.push_back(
+            map_to_id(daughter_coords, opts.max_level, num_dims));
       }
     }
   }
--- a/src/elements.cpp
+++ b/src/elements.cpp
@@ -218,24 +218,24 @@ table::get_child_elements(int64_t const index, options const &opts) const
   // all coordinates have 2 entries (lev, cell) per dimension
   auto const num_dims = coords.size() / 2;
 
+  auto const max_adapt_levels = opts.max_adapt_levels;
   std::list<int64_t> daughter_ids;
   for (auto i = 0; i < num_dims; ++i)
   {
+    int level = max_adapt_levels.empty() ? opts.max_level : max_adapt_levels[i];
     // first daughter in this dimension
-    if (coords(i) + 1 < opts.max_level)
+    if (coords(i) + 1 < level)
     {
       auto daughter_coords          = coords;
       daughter_coords(i)            = coords(i) + 1;
       daughter_coords(i + num_dims) = coords(i + num_dims) * 2;
-      daughter_ids.push_back(
-          map_to_id(daughter_coords, opts.max_level, num_dims));
+      daughter_ids.push_back(map_to_id(daughter_coords, level, num_dims));
 
       // second daughter
       if (coords(i) >= 1)
       {
         daughter_coords(i + num_dims) = coords(i + num_dims) * 2 + 1;
-        daughter_ids.push_back(
-            map_to_id(daughter_coords, opts.max_level, num_dims));
+        daughter_ids.push_back(map_to_id(daughter_coords, level, num_dims));
       }
     }
   }
--- a/src/pde/pde_base.hpp
+++ b/src/pde/pde_base.hpp
@@ -823,20 +823,21 @@ private:
     // set maximum level to generate term coefficients
     if (cli_input.do_adapt_levels())
     {
-      return cli_input.get_max_level();
+      if(auto const & levels = cli_input.get_max_adapt_levels(); !levels.empty())
+        return *std::max_element(levels.begin(), levels.end());
+      else
+        return cli_input.get_max_level();
     }
     else
     {
       // if adaptivity is not used, only generate to the highest dim level
       auto const levels = cli_input.get_starting_levels();
-      return levels.size() > 0
-                 ? *std::max_element(levels.begin(), levels.end())
-                 : std::max_element(
+      return levels.empty() ? std::max_element(
                        dims.begin(), dims.end(),
                        [](dimension<P> const &a, dimension<P> const &b) {
                          return a.get_level() < b.get_level();
                        })
-                       ->get_level();
+                       ->get_level() : *std::max_element(levels.begin(), levels.end());
     }
   }
 
--- a/src/pde/pde_base.hpp
+++ b/src/pde/pde_base.hpp
@@ -823,7 +823,8 @@ private:
     // set maximum level to generate term coefficients
     if (cli_input.do_adapt_levels())
     {
-      if(auto const & levels = cli_input.get_max_adapt_levels(); !levels.empty())
+      if (auto const &levels = cli_input.get_max_adapt_levels();
+          !levels.empty())
         return *std::max_element(levels.begin(), levels.end());
       else
         return cli_input.get_max_level();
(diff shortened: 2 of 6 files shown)


src/elements.cpp
```diff
@@ -218,24 +218,24 @@ table::get_child_elements(int64_t const index, options const &opts) const
...
     // first daughter in this dimension
-    if (coords(i) + 1 < opts.max_level)
+    if (coords(i) + 1 < level)
     {
       auto daughter_coords          = coords;
       daughter_coords(i)            = coords(i) + 1;
       daughter_coords(i + num_dims) = coords(i + num_dims) * 2;
-      daughter_ids.push_back(
-          map_to_id(daughter_coords, opts.max_level, num_dims));
+      daughter_ids.push_back(map_to_id(daughter_coords, level, num_dims));
```
stefan-schnake: This line needs to be reverted.  `opts.max_level` defines the indicies for the hierarchy across all dimensions.  If you change this here, then a lot needs to be changed to compensate (for no gain).

src/elements.cpp
```diff
@@ -218,24 +218,24 @@ table::get_child_elements(int64_t const index, options const &opts) const
...
 
       // second daughter
       if (coords(i) >= 1)
       {
         daughter_coords(i + num_dims) = coords(i + num_dims) * 2 + 1;
-        daughter_ids.push_back(
-            map_to_id(daughter_coords, opts.max_level, num_dims));
+        daughter_ids.push_back(map_to_id(daughter_coords, level, num_dims));
```
stefan-schnake: Same as Line 232

src/program_options_tests.cpp
```diff
@@ -124,145 +126,172 @@ TEST_CASE("parser constructor/getters", "[program_options]")
...
+  SECTION("set --max_adapt_level below starting levels")
   {
     std::cerr.setstate(std::ios_base::failbit);
     parser const p =
-        make_parser({"asgard", "--solver=gmres", "--outer_it=-10"});
+        make_parser({"--adapt", R"(-l "5 5")", R"(--max_adapt_level="4 4")"});
     std::cerr.clear();
     REQUIRE(!p.is_valid());
   }
+  //--adapt
```
stefan-schnake: Add a test with `max_adapt_level` with different values for each dimension.

src/pde/pde_base.hpp
```diff
@@ -823,20 +823,24 @@ class PDE
...
+        return cli_input.get_max_level();
     }
     else
     {
       // if adaptivity is not used, only generate to the highest dim level
       auto const levels = cli_input.get_starting_levels();
-      return levels.size() > 0
-                 ? *std::max_element(levels.begin(), levels.end())
-                 : std::max_element(
+      return levels.empty()
```
stefan-schnake, line 836: Is this just reordering the logic?

stefan-schnake (changes requested):

src/elements.cpp
ckendrick: I think this and the similar change below should still use `opts.max_level` here since that maximum level is used in the hash table. We want to always use the highest level when doing these `map_to_id` calls.

If this is run with level 4 6 6 6, we want this to be 6 instead of the current dimension level. We need to double check with @stefan-schnake though.

src/pde/pde_base.hpp
quantumsteve: Yes, I switched `level.size() > 0` to `levels.empty()` and swapped true and false branches.

src/program_options_tests.cpp
quantumsteve: done in `time_advance_tests.cpp`

src/elements.cpp
quantumsteve: changed

src/elements.cpp
quantumsteve: updated

src/elements.cpp
quantumsteve: changed back to `opts.max_level`

src/program_options.cpp
```diff
@@ -375,6 +376,53 @@ parser::parser(int argc, char const *const *argv)
...
+          << '\n';
+      valid = false;
+    }
+    if (!(max_adapt_lev.empty() && do_adapt))
+    {
+      std::cerr
+          << "input maximum adaptivity levels without enabling adaptivity..."
+          << '\n';
+      valid = false;
+    }
```
ckendrick: Is this if statement duplicated?

ckendrick (changes requested):

mkstoyanov:
What exactly is the difference between `max_adapt_level` and `max_level`? Finer control is not clear to me.

src/program_options.cpp
quantumsteve: :man_facepalming:

quantumsteve:
> What exactly is the difference between `max_adapt_level` and `max_level`? Finer control is not clear to me.

See https://github.com/project-asgard/asgard/issues/583

stefan-schnake:
> What exactly is the difference between max_adapt_level and max_level? Finer control is not clear to me.

An original assumption of this code is when adaptively is enabled, you should let the adaptivity determine what the proper level is in each dimension.  `max_level` was used to preload the coefficients matrices (which were time independent), define the hierarchy, and was expected to be set very high.

However, with the problems that we are doing now, that involve different timescales in the x vs v directions, having a uniform max_level is not practical.  Often the algorithm will adapt too far, violate the CFL, and then blow up.

Because `max_level` is used to define the hierarchy, changing it could break a lot of things.  I'm fine if we remove the option to change `max_level` from the command line.

mkstoyanov:
So you want anisotropic max-level (which is a better word than adapt-max-level since all cases refer to adaptivity).

We should either use a time-step that satisfies the CFL for all levels or have a way to adapt the time-step based on the CFL. How do we do that now? Don't we have something that computes CFL?

stefan-schnake:
> So you want anisotropic max-level (which is a better word than adapt-max-level since all cases refer to adaptivity).

Yes

> We should either use a time-step that satisfies the CFL for all levels or have a way to adapt the time-step based on the CFL. How do we do that now?

There is an ability to set a CFL based of one of the dimensions, but there is no way to set a final time command line input so it's pretty useless.

mkstoyanov (approved):

src/program_options.cpp
```diff
@@ -81,7 +81,9 @@ parser::parser(int argc, char const *const *argv)
       clara::detail::Opt(gmres_inner_iterations, "inner_it > 0")["--inner_it"](
           "Number of inner iterations in gmres solver") |
       clara::detail::Opt(gmres_outer_iterations, "outer_it > 0")["--outer_it"](
-          "Number of outer iterations in gmres solver");
+          "Number of outer iterations in gmres solver") |
+      clara::detail::Opt(aniso_adapt_levels_str, "")["--aniso_adapt_levels"](
```
ckendrick: Can we keep the user-facing option as `--max_adapt_levels` but keep the help text as-is (anisotropic hierarchical levels)? 
I feel like `--max_adapt_levels` is more clear for the user given we have options for `--max_level` and `--adapt`. The user might look for something like max adapt level instead of having to know it is `--aniso_adapt_levels`.

mkstoyanov: The issue is that both `max_level` and `max_adapt_levels` are list of numbers, one for each dimension. Think of the following:
```
--max_adapt_levels "3 4 4" --max_level "4 3 4"
```
How confusing is that?

If we want to keep `max_adapt_levels`, we should change the other one to accept a single number. That way it is clear that one option refers to all levels, while the other provides per-level granularity.

ckendrick: `--max_level` is only a single number. Currently `--max_level 6` is interpreted as `6 6 6` when adapting.

`--max_adapt_levels` will let us override that per-dimension by doing something like `--max_adapt_levels "3 4 6" --max_level 6`

mkstoyanov: Wow, I guess you can pass in just a single number. I've been passing in lists and it works just fine.

Then it's Ok to have the option.

mkstoyanov (dismissed):

src/program_options.cpp
ckendrick: Hmm.. that seems like a CLI parsing bug if it is accepting a list for `--max_level`. It is defined as an int in the `parser` class: https://github.com/project-asgard/asgard/blob/170da165cb60b2cc344cf9ebc8ccd5546d065649/src/program_options.hpp#L352

mkstoyanov: Try it, it's perfectly happy to accept a list, just like the `-l` option. Anyway, that's why I was confused on why you would want two lists.

ckendrick: You're right, it does accept a list! This is unintentional with the current implementation. It looks like it is using only the first element to set the value. I guess the clara parser is reducing this list to an integer automatically? This is definitely an edge case not caught in the current CLI parsing validation.

If @quantumsteve can rename `--aniso_adapt_levels` to `--max_adapt_levels` then this PR looks good to me.

mkstoyanov (approved):

ckendrick (approved):

PR #589, opened 2023-07-18
Use refined solution as initial guess to GMRES

### LINKED ISSUE #584: Use refined solution as initial guess in GMRES
Adaptivity is currently slow because if the solution is refined, then a GMRES solve is done from scratch with an initial condition of the previous timestep prolongated on the current grid.  Additionally, the current adaptive update is checked for possible refinement, and if refinement is needed and determined, then this update is thrown away.

If adaptivity is enabled and the grid is refined, we should use the (prolongated) update as the initial condition to the GMRES solve.  Hopefully, this would result in fewer iterations for convergence on the refined grids as you are searching only on the modes added in the refinement.  GMRES is a bad smoother so there is still some improvement to be done after this, but its a good start.

ckendrick:

## Proposed changes

This adds the first refinement solution to be the input to GMRES in subsequent refinement steps in IMEX.

Closes #584

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist


src/time_advance.cpp
```diff
@@ -138,6 +139,19 @@ adaptive_advance(method const step_method, PDE<P> &pde,
       auto const y1 =
           adaptive_grid.redistribute_solution(y, old_plan, old_size);
       y.resize(y1.size()) = y1;
+
+      // after first refinement, save the refined vector to use as initial
+      // "guess" to GMRES
+      if (y_first_refine.empty())
+      {
+        y_first_refine = fk::vector<P>(y_refined);
```
```suggestion
        y_first_refine = std::move(y_refined);
```
requires making `y_refined` mutable on line 120/121.

src/time_advance.cpp
```diff
@@ -138,6 +139,19 @@ adaptive_advance(method const step_method, PDE<P> &pde,
       auto const y1 =
           adaptive_grid.redistribute_solution(y, old_plan, old_size);
       y.resize(y1.size()) = y1;
```
```suggestion
      y = adaptive_grid.redistribute_solution(y, old_plan, old_size);
```
Another copy we can avoid.

mkstoyanov:
Generic comments, not linked to a specific lines but the algorithms in general:
* there are inconsistent names here, the new variables is sometimes called `x_something` sometimes `y_something` and sometimes it interacts with `f`. Instead of of `x` and `y`, can we name it something like `state` or whatever is appropriate
* the variable is set in the `imex` method and held in the refinement method, but the variable is not held in `main`, can we not save something by keeping information across time-steps?

src/time_advance.hpp
```diff
@@ -63,7 +63,8 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
              options const &program_opts,
              std::array<boundary_conditions::unscaled_bc_parts<P>, 2> const
                  &unscaled_parts,
-             fk::vector<P> const &x_orig, P const time, solve_opts const solver,
+             fk::vector<P> const &x_orig, fk::vector<P> const &x_prev,
```
quantumsteve, line 66:
```suggestion
             fk::vector<P> const &x_orig, std::optional<fk::vector<P>> const &x_prev,
```

quantumsteve: Using an empty vector for an optional argument works, but I wounder if std::optional is more explicit :thinking:

src/time_advance.cpp
```diff
@@ -148,6 +149,19 @@ adaptive_advance(method const step_method, PDE<P> &pde,
...
+
+      // after first refinement, save the refined vector to use as initial
+      // "guess" to GMRES
+      if (y_first_refine.empty())
+      {
+        y_first_refine = std::move(y_refined);
+      }
+
+      // pad with zeros if more elements were added
+      if (y_first_refine.size() < y.size())
```
quantumsteve: Would it be safer to always resize if different?

ckendrick:
> * there are inconsistent names here, the new variables is sometimes called `x_something` sometimes `y_something` and sometimes it interacts with `f`. Instead of of `x` and `y`, can we name it something like `state` or whatever is appropriate

I agree, the names are inconsistent throughout. We are working on renaming IMEX to make it more clear, and as part of that we will rename all of these `x_orig` `y`, etc to be consistent across all the methods in `time_advance`. This is a larger refactor and I didn't want to clutter this PR with those changes. I can certainly include those changes in this PR if you think now is a good time to do it.

> * the variable is set in the `imex` method and held in the refinement method, but the variable is not held in `main`, can we not save something by keeping information across time-steps?

I can make `y_first_refine` defined in `main` and passed in through the `adaptive_advance` function, but it is only used during the adaptivity portion of the current time step and the next `adaptive_advance` call does not use `y_first_refine`. Storing it in main would help reduce recreating this every time step though.

In general, there probably needs to be a better way to pass things from `main` into the time advance routines to avoid a lot of vector/matrix creations and reallocations. Maybe some sort of "workspace" struct that can hold the memory allocations for a lot of temporaries created during the time advance. This is probably work for a future PR.

src/time_advance.hpp
ckendrick: I'm fine with using `std::optional` but does it support references? I think we would want something like `std::optional<fk::vector<P> &>` to avoid copying `x_prev`.

quantumsteve: `boost::optional` supports references but `std::optional` does not. (No, I'm not suggesting we add boost as a dependency)

quantumsteve (approved):

PR #597, opened 2023-07-26
Add file checkpoint restart

ckendrick:

## Proposed changes

Adds initial implementation of a `--restart` CLI option that accepts an HDF5 file allowing simulations to be resumed.

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist

### DIFF
--- a/src/adapt.hpp
+++ b/src/adapt.hpp
@@ -99,6 +99,16 @@ public:
     return plan_.at(rank);
   }
 
+  void
+  recreate_table(std::vector<int64_t> const &element_ids, int const max_level)
+  {
+    table_.recreate_from_elements(element_ids, max_level);
+
+    auto const new_plan = get_plan(get_num_ranks(), table_);
+    plan_.clear();
+    plan_.insert(new_plan.cbegin(), new_plan.cend());
+  }
+
   elements::table const &get_table() const { return table_; }
   int64_t size() const { return table_.size(); }
 
--- a/src/adapt.hpp
+++ b/src/adapt.hpp
@@ -104,7 +104,7 @@ public:
   {
     table_.recreate_from_elements(element_ids, max_level);
 
-    plan_ = std::move(get_plan(get_num_ranks(), table_));
+    plan_ = get_plan(get_num_ranks(), table_);
   }
 
   elements::table const &get_table() const { return table_; }
--- a/src/adapt.hpp
+++ b/src/adapt.hpp
@@ -104,9 +104,7 @@ public:
   {
     table_.recreate_from_elements(element_ids, max_level);
 
-    auto const new_plan = get_plan(get_num_ranks(), table_);
-    plan_.clear();
-    plan_.insert(new_plan.cbegin(), new_plan.cend());
+    plan_ = std::move(get_plan(get_num_ranks(), table_));
   }
 
   elements::table const &get_table() const { return table_; }
--- a/src/elements.cpp
+++ b/src/elements.cpp
@@ -330,7 +330,6 @@ void table::recreate_from_elements(std::vector<int64_t> const &element_ids,
   int const new_table_size = static_cast<int>(element_ids.size() / coord_size);
 
   std::cout << "Recreating element table:\n";
-  std::cout << "  - original elements: " << original_ids.size() << "\n";
   std::cout << "  - elements from restart: " << new_table_size << "\n";
 
   // clear the existing hash table
--- a/src/elements.cpp
+++ b/src/elements.cpp
@@ -359,8 +359,7 @@ void table::recreate_from_elements(std::vector<int64_t> const &element_ids,
   }
 
   expect(active_element_ids_.size() == id_to_coords_.size());
-  active_table_d_.resize(dev_table_builder.size())
-      .transfer_from(dev_table_builder);
+  active_table_.resize(dev_table_builder.size()) = dev_table_builder;
 
   std::cout << "  - after recreation: " << size() << "\n";
   expect(size() == new_table_size);
--- a/src/elements.cpp
+++ b/src/elements.cpp
@@ -312,6 +312,60 @@ table::table(options const &opts, std::vector<dimension<P>> const &dims)
   active_table_.resize(dev_table_builder.size()) = dev_table_builder;
 }
 
+void table::recreate_from_elements(std::vector<int64_t> const &element_ids,
+                                   int const max_level)
+{
+  // For restarting, we want the element table to contain only the active ids
+  // from the restart file. The active ids saved in the restart file is the
+  // flattened device table, so we need to recreate the element key from the
+  // coordinates.
+  std::vector<int64_t> original_ids(active_element_ids_);
+
+  int const coord_size = get_coords(0).size();
+  expect(coord_size % 2 == 0);
+  int const num_dims = coord_size / 2;
+
+  // calculate the new table size based on the size of each element
+  expect(element_ids.size() % coord_size == 0);
+  int const new_table_size = static_cast<int>(element_ids.size() / coord_size);
+
+  std::cout << "Recreating element table:\n";
+  std::cout << "  - original elements: " << original_ids.size() << "\n";
+  std::cout << "  - elements from restart: " << new_table_size << "\n";
+
+  // clear the existing hash table
+  active_element_ids_.clear();
+  id_to_coords_.clear();
+
+  fk::vector<int> dev_table_builder;
+  for (int i = 0; i < new_table_size; i++)
+  {
+    // build a coord set out of the flattened device table
+    fk::vector<int> coords(coord_size);
+    for (int j = 0; j < coord_size; j++)
+    {
+      coords(j) = element_ids[j + i * coord_size];
+    }
+
+    // get full linear id as key to active element id
+    int64_t id = map_to_id(coords, max_level, num_dims);
+
+    // add this element to the hash table
+    active_element_ids_.push_back(id);
+    id_to_coords_[id].resize(coords.size()) = coords;
+
+    // add the element coords to the flattened device table
+    dev_table_builder.concat(coords);
+  }
+
+  expect(active_element_ids_.size() == id_to_coords_.size());
+  active_table_d_.resize(dev_table_builder.size())
+      .transfer_from(dev_table_builder);
+
+  std::cout << "  - after recreation: " << size() << "\n";
+  expect(size() == new_table_size);
+}
+
 // static construction helper
 // return the cell indices, given a level tuple
 // each row in the returned matrix is the cell portion of an element's
--- a/src/elements.cpp
+++ b/src/elements.cpp
@@ -349,16 +349,16 @@ void table::recreate_from_elements(std::vector<int64_t> const &element_ids,
     // get full linear id as key to active element id
     int64_t id = map_to_id(coords, max_level, num_dims);
 
-    // add this element to the hash table
-    active_element_ids_.push_back(id);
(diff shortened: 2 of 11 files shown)


src/elements.cpp
```diff
@@ -312,6 +312,59 @@ table::table(options const &opts, std::vector<dimension<P>> const &dims)
...
+  int const coord_size = get_coords(0).size();
+  expect(coord_size % 2 == 0);
+  int const num_dims = coord_size / 2;
+
+  // calculate the new table size based on the size of each element
+  expect(element_ids.size() % coord_size == 0);
+  int const new_table_size = static_cast<int>(element_ids.size() / coord_size);
+
+  std::cout << "Recreating element table:\n";
+  std::cout << "  - original elements: " << original_ids.size() << "\n";
```
stefan-schnake: What's the purpose of displaying this?  If we're adapting, then these two numbers could be different.

ckendrick: That's true, this was more just to display feedback for info/debugging purposes if the numbers were different but not expected to be. This is before more metadata is saved (such as whether FG or ASG was used in the original run). If the user ran `--restart` but forgot to add `--adapt` or `-f`, then these could be different or the same.

We could definitely remove this completely if that's better, or rework to have better logic for different cases.

src/adapt.hpp
```diff
@@ -99,6 +99,16 @@ class distributed_grid
     return plan_.at(rank);
   }
 
+  void
+  recreate_table(std::vector<int64_t> const &element_ids, int const max_level)
+  {
+    table_.recreate_from_elements(element_ids, max_level);
+
+    auto const new_plan = get_plan(get_num_ranks(), table_);
+    plan_.clear();
```
mkstoyanov: Clear followed by insert? Don't we have move assignment for this data structutre:
```
plan_ = get_plan(get_num_ranks(), table_);
```

mkstoyanov (changes requested):

src/adapt.hpp
cianciosa: This may be related to this comment.

https://github.com/project-asgard/asgard/blob/b8a9b9b950200fd2f3047ff87387059f0bfa5b72/src/adapt.cpp#L362C1-L362C1

ckendrick: Yes @cianciosa is correct, I don't know if we have found a workaround for that issue yet or not.

src/elements.cpp
stefan-schnake: I would remove the original elements # but keep the other.

src/adapt.hpp
mkstoyanov: Just tried on clang 14 and it works fine. Give it a try, there's no reason why this wouldn't work.

PS: also works on clang 12, whatever the real problem has been must have been fixed.

ckendrick: I swapped this to use `std::move` following the changes added by #600.

src/elements.cpp
```diff
@@ -312,6 +312,58 @@ table::table(options const &opts, std::vector<dimension<P>> const &dims)
...
+    {
+      coords(j) = element_ids[j + i * coord_size];
+    }
+
+    // get full linear id as key to active element id
+    int64_t id = map_to_id(coords, max_level, num_dims);
+
+    // add this element to the hash table
+    active_element_ids_.push_back(id);
+    id_to_coords_[id].resize(coords.size()) = coords;
```
quantumsteve:
```suggestion
    id_to_coords_[id] = std::move(coords);
```

src/adapt.hpp
```diff
@@ -99,6 +99,14 @@ class distributed_grid
     return plan_.at(rank);
   }
 
+  void
+  recreate_table(std::vector<int64_t> const &element_ids, int const max_level)
+  {
+    table_.recreate_from_elements(element_ids, max_level);
+
+    plan_ = std::move(get_plan(get_num_ranks(), table_));
```
quantumsteve:
```suggestion
    plan_ = get_plan(get_num_ranks(), table_);
```

std::move should have no effect.

src/elements.cpp
```diff
@@ -312,6 +312,58 @@ table::table(options const &opts, std::vector<dimension<P>> const &dims)
...
+
+    // add the element coords to the flattened device table
+    dev_table_builder.concat(coords);
+  }
+
+  expect(active_element_ids_.size() == id_to_coords_.size());
+  active_table_.resize(dev_table_builder.size()) = dev_table_builder;
```
quantumsteve:
```suggestion
  active_table_ = std::move(dev_table_builder);
```

src/io.hpp
```diff
@@ -158,7 +163,119 @@ void write_output(PDE<P> const &pde, parser const &cli_input,
     H5Easy::dump(file, "gmres" + std::to_string(i) + "_num_inner",
                  pde.gmres_outputs[i].inner_iter, opts);
   }
+
+  file.flush();
   tools::timer.stop("write_output");
 }
 
+template<typename P>
+void read_restart_metadata(parser *user_vals, std::string restart_file)
```
```suggestion
void read_restart_metadata(parser &user_vals, std::string const &restart_file)
```

src/io.hpp
```diff
@@ -158,7 +163,119 @@ void write_output(PDE<P> const &pde, parser const &cli_input,
...
+  HighFive::File file(restart_file, HighFive::File::ReadOnly);
+
+  std::string const pde_string =
+      H5Easy::load<std::string>(file, std::string("pde"));
+  int const degree = H5Easy::load<int>(file, std::string("degree"));
+  P const dt       = H5Easy::load<P>(file, std::string("dt"));
+  P const time     = H5Easy::load<P>(file, std::string("time"));
+
+  int const ndims    = H5Easy::load<int>(file, std::string("ndims"));
+  std::string levels = "";
```
```suggestion
  std::string levels;
```

src/io.hpp
quantumsteve: Since `user_vals` is never NULL, use a reference.

src/io.hpp
```diff
@@ -158,7 +163,119 @@ void write_output(PDE<P> const &pde, parser const &cli_input,
...
+  int const max_level = H5Easy::load<int>(file, std::string("max_level"));
+  int const dof       = H5Easy::load<int>(file, std::string("dof"));
+  // TODO: this will be used for validation in the future
+  ignore(dof);
+
+  parser_mod::set(*user_vals, parser_mod::pde_str, pde_string);
+  parser_mod::set(*user_vals, parser_mod::degree, degree);
+  parser_mod::set(*user_vals, parser_mod::dt, dt);
+  parser_mod::set(*user_vals, parser_mod::starting_levels_str, levels);
+  parser_mod::set(*user_vals, parser_mod::max_level, max_level);
```
```suggestion
  parser_mod::set(user_vals, parser_mod::pde_str, pde_string);
  parser_mod::set(user_vals, parser_mod::degree, degree);
  parser_mod::set(user_vals, parser_mod::dt, dt);
  parser_mod::set(user_vals, parser_mod::starting_levels_str, levels);
  parser_mod::set(user_vals, parser_mod::max_level, max_level);
```

src/io.hpp
```diff
@@ -158,7 +163,119 @@ void write_output(PDE<P> const &pde, parser const &cli_input,
...
+  expect(static_cast<int>(pde.moments.size()) == num_moments);
+  for (int i = 0; i < num_moments; ++i)
+  {
+    pde.moments[i].createMomentReducedMatrix(pde, hash_table);
+    pde.moments[i].set_realspace_moment(
+        fk::vector<P>(H5Easy::load<std::vector<P>>(
+            file, std::string("moment" + std::to_string(i)))));
+  }
+
+  int step_index = (int)(time / dt);
```
```suggestion
  int step_index = static_cast<int>(time / dt);
```

src/pde/pde_base.hpp
```diff
@@ -710,6 +710,8 @@ class PDE
   fk::vector<P> poisson_off_diag;
 
   fk::vector<P> E_field;
+  fk::vector<P> phi;
+  fk::vector<P> E_source;
```
quantumsteve, line 714: Are these connected to checkpoint restart?

src/program_options.cpp
```diff
@@ -422,6 +452,7 @@ bool parser::using_imex() const { return use_imex_stepping; }
 bool parser::using_full_grid() const { return use_full_grid; }
 bool parser::do_poisson_solve() const { return do_poisson; }
 bool parser::do_adapt_levels() const { return do_adapt; }
+bool parser::do_restart() const { return restart_file != NO_USER_VALUE_STR; }
```
quantumsteve, line 455: (not for this PR) it looks like clara supports std::optional.

src/time_advance.cpp
```diff
@@ -548,7 +550,9 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
       return interp1(nodes, poisson_E, {x_v})[0];
     };
 
-    pde.E_field = poisson_E;
+    pde.E_field  = poisson_E;
```
quantumsteve, line 553: Should changes to time_advance be part of this PR?

quantumsteve (changes requested):
see comments

src/elements.cpp
ckendrick: `coords` is still used below this line so `move` here causes an issue: https://github.com/project-asgard/asgard/blob/ce77cfe4aee97a1dce1b7dcf79507c0c8d117bd7/src/elements.cpp#L357

This method needs to be re-written since it has the same `concat`/realloc issue as the regular element table creation in #598.

This could also be removed once we change how the element table is stored since this is taking the flattened indices saved to file and expanding them back into the [lev0, lev1 ..., levn, cell0, cell1, ..., celln] format.

quantumsteve: Can you swap lines 354 and 357. `fk::vector::concat` only takes const reference.

src/pde/pde_base.hpp
ckendrick: I don't believe these are explicitly required for restarting since they get recreated in IMEX, but this should be confirmed. These were mostly added for info purposes to plot. Not sure if they are still needed or required though.

src/program_options.cpp
ckendrick: That is good to know. I think a lot of CLI options could make use of that.

src/pde/pde_base.hpp
stefan-schnake: They are not connected to the restart as IMEX rebuilds all of the moment data and derivative objects.

ckendrick: Should these get removed entirely or are they still useful to store?

stefan-schnake: They're only useful in a debugging capacity.   Otherwise we don't need them

quantumsteve:
We should have a test running `n` timesteps and another that restarts sometime before then. Do you want to do that in a separate PR?

mkstoyanov (approved):

quantumsteve (approved):

mkstoyanov:
> We should have a test running `n` timesteps and another that restarts sometime before then. Do you want to do that in a separate PR?

I don't want to have hanging PRs and hold things back (divergence is a problem). Let's merge now and we can keep improving over time. (adding missing tests should be very high priority)

quantumsteve:
> > We should have a test running `n` timesteps and another that restarts sometime before then. Do you want to do that in a separate PR?
> 
> I don't want to have hanging PRs and hold things back (divergence is a problem). Let's merge now and we can keep improving over time. (adding missing tests should be very high priority)

I agree. :shipit:

PR #598, opened 2023-07-26
Reduce memory resizes in element table creation

ckendrick:

## Proposed changes

This reduces use of `concat` when creating the element table. For large problems with millions of DOF, `concat` was slow due to freeing/reallocating memory for each coordinate pair (~3-4 hrs for ~230 million elements). Instead, this allocates size for the element table upfront and updates portions of the table rather than reallocating for each coordinate. The size allocated initially assumes the FG worst-case, but shrinks afterwards if needed for smaller adaptive grid problems.

This can still be improved, especially with the planned refactor of the element table, but for now this provides a quick workaround for larger high-dimensional problems.

- [ ] Bugfix
- [ ] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04, wombat

## Checklist

### DIFF
--- a/src/elements.cpp
+++ b/src/elements.cpp
@@ -282,17 +282,14 @@ table::table(options const &opts, std::vector<dimension<P>> const &dims)
   }();
 
   fk::vector<int> dev_table_builder;
-  if (opts.use_full_grid)
+  int64_t dof = 1;
+  for (int lev = 0; lev < dims.size(); lev++)
   {
-    int64_t dof = 1;
-    for (int lev = 0; lev < dims.size(); lev++)
-    {
-      dof *= dims[0].get_degree() * fm::two_raised_to(dims[lev].get_level());
-    }
-
-    // reserve element table data up front
-    dev_table_builder.resize(dof);
+    dof *= dims[0].get_degree() * fm::two_raised_to(dims[lev].get_level());
   }
+
+  // reserve element table data up front
+  dev_table_builder.resize(dof);
 
   int64_t pos = 0;
   for (int row = 0; row < perm_table.nrows(); ++row)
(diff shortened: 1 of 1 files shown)


mkstoyanov (approved):

mkstoyanov (dismissed):

src/elements.cpp
```diff
@@ -282,6 +282,16 @@ table::table(options const &opts, std::vector<dimension<P>> const &dims)
   }();
 
   fk::vector<int> dev_table_builder;
+  int64_t dof = 1;
+  for (size_t lev = 0; lev < dims.size(); lev++)
+  {
+    dof *= dims[0].get_degree() * fm::two_raised_to(dims[lev].get_level());
```
quantumsteve: Should this be `dims[lev].get_degree()`? Otherwise, we can pull `std::pow(dims[0].get_degree(), dims.size())`  out of the loop.

mkstoyanov: The degree is the same in all directions, you can argue that it shouldn't even be part of dims.

I'm fine with implementing `std::pow(dims[0].get_degree(), dims.size())` inside the for-loop, it's probably cheaper than the logarithms that `std::pow` is probably using.

src/elements.cpp
```diff
@@ -304,10 +314,24 @@ table::table(options const &opts, std::vector<dimension<P>> const &dims)
...
     }
   }
 
+  if (pos < dev_table_builder.size())
+  {
+    dev_table_builder = dev_table_builder.extract(0, pos - 1);
+  }
+
   expect(active_element_ids_.size() == id_to_coords_.size());
   active_table_.resize(dev_table_builder.size()) = dev_table_builder;
```
quantumsteve:
```suggestion
  active_table_ = std:move(dev_table_builder);
```

mkstoyanov: You're right, but I don't think this line is part of this PR.

src/elements.cpp
quantumsteve: Just wanted to ensure the index was correct.

mkstoyanov (approved):

PR #599, opened 2023-07-27
Add initial sparse class for sparse moment matrices

ckendrick:

## Proposed changes

This adds a `sparse` class for basic storage of sparse matrices in CSR format. The class currently makes use of the existing vector/matrix classes for reuse of storage/functionality and compatibility with existing code. Functionality is limited and should be extended more in future PRs.

Currently this is only being used in the `moment` class to store the moment matrices for use in IMEX. This is necessary for PDEs beyond 1x1v. `gemv` routines wrapping cuSPARSE are added to `fast_math` and `lib_dispatch` to calculate moments in IMEX. A very basic CPU implementation is also added for support with non-CUDA builds but is not optimized. Test coverage is limited and needs to be extended as well.

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist

### DIFF
--- a/src/fast_math.hpp
+++ b/src/fast_math.hpp
@@ -460,12 +460,11 @@ void getrs(fk::matrix<P, amem> const &A, fk::scalapack_matrix_info &ainfo,
 #endif
 
 // sparse gemv - sparse matrix dense vector multiplication
-template<typename P, mem_type amem, mem_type xmem, mem_type ymem,
-         resource resrc>
+template<typename P, mem_type xmem, mem_type ymem, resource resrc>
 fk::vector<P, ymem, resrc> &
-sparse_gemv(fk::sparse<P, amem, resrc> const &A,
-            fk::vector<P, xmem, resrc> const &x, fk::vector<P, ymem, resrc> &y,
-            bool const trans_A = false, P const alpha = 1.0, P const beta = 0.0)
+sparse_gemv(fk::sparse<P, resrc> const &A, fk::vector<P, xmem, resrc> const &x,
+            fk::vector<P, ymem, resrc> &y, bool const trans_A = false,
+            P const alpha = 1.0, P const beta = 0.0)
 {
   int const rows_opA = trans_A ? A.ncols() : A.nrows();
   int const cols_A   = trans_A ? A.nrows() : A.ncols();
--- a/src/fast_math.hpp
+++ b/src/fast_math.hpp
@@ -480,9 +480,8 @@ sparse_gemv(fk::sparse<P, amem, resrc> const &A,
   int n             = A.ncols();
   int nnz           = A.nnz();
 
-  lib_dispatch::sparse_gemv(&transa, &m, &n, &nnz, A.offsets(), A.columns(),
-                            A.data(), &alpha_, x.data(), &beta_, y.data(),
-                            resrc);
+  lib_dispatch::sparse_gemv<resrc>(&transa, &m, &n, &nnz, A.offsets(), A.columns(),
+                            A.data(), &alpha_, x.data(), &beta_, y.data());
 
   return y;
 }
--- a/src/fast_math.hpp
+++ b/src/fast_math.hpp
@@ -480,8 +480,9 @@ sparse_gemv(fk::sparse<P, amem, resrc> const &A,
   int n             = A.ncols();
   int nnz           = A.nnz();
 
-  lib_dispatch::sparse_gemv<resrc>(&transa, &m, &n, &nnz, A.offsets(), A.columns(),
-                            A.data(), &alpha_, x.data(), &beta_, y.data());
+  lib_dispatch::sparse_gemv<resrc>(&transa, &m, &n, &nnz, A.offsets(),
+                                   A.columns(), A.data(), &alpha_, x.data(),
+                                   &beta_, y.data());
 
   return y;
 }
--- a/src/fast_math.hpp
+++ b/src/fast_math.hpp
@@ -1,6 +1,7 @@
 #pragma once
 #include "lib_dispatch.hpp"
 #include "program_options.hpp"
+#include "sparse.hpp"
 #include "tensors.hpp"
 #include "tools.hpp"
 #ifdef ASGARD_USE_SCALAPACK
@@ -457,5 +458,33 @@ void getrs(fk::matrix<P, amem> const &A, fk::scalapack_matrix_info &ainfo,
   }
 }
 #endif
+
+// sparse gemv - sparse matrix dense vector multiplication
+template<typename P, mem_type amem, mem_type xmem, mem_type ymem,
+         resource resrc>
+fk::vector<P, ymem, resrc> &
+sparse_gemv(fk::sparse<P, amem, resrc> const &A,
+            fk::vector<P, xmem, resrc> const &x, fk::vector<P, ymem, resrc> &y,
+            bool const trans_A = false, P const alpha = 1.0, P const beta = 0.0)
+{
+  int const rows_A = trans_A ? A.ncols() : A.nrows();
+  int const cols_A = trans_A ? A.nrows() : A.ncols();
+
+  expect(rows_A == y.size());
+  expect(cols_A == x.size());
+
+  P alpha_          = alpha;
+  P beta_           = beta;
+  char const transa = trans_A ? 't' : 'n';
+  int m             = A.nrows();
+  int n             = A.ncols();
+  int nnz           = A.nnz();
+
+  lib_dispatch::sparse_gemv(&transa, &m, &n, &nnz, A.offsets(), A.columns(),
+                            A.data(), &alpha_, x.data(), &beta_, y.data(),
+                            resrc);
+
+  return y;
+}
 
 } // namespace asgard::fm
--- a/src/fast_math.hpp
+++ b/src/fast_math.hpp
@@ -467,22 +467,16 @@ sparse_gemv(fk::sparse<P, amem, resrc> const &A,
             fk::vector<P, xmem, resrc> const &x, fk::vector<P, ymem, resrc> &y,
             bool const trans_A = false, P const alpha = 1.0, P const beta = 0.0)
 {
-  int const rows_A = trans_A ? A.ncols() : A.nrows();
-  int const cols_A = trans_A ? A.nrows() : A.ncols();
+  int const rows_opA = trans_A ? A.ncols() : A.nrows();
+  int const cols_A   = trans_A ? A.nrows() : A.ncols();
 
-  expect(rows_A == y.size());
+  expect(rows_opA == y.size());
   expect(cols_A == x.size());
 
-  P alpha_          = alpha;
-  P beta_           = beta;
   char const transa = trans_A ? 't' : 'n';
-  int m             = A.nrows();
-  int n             = A.ncols();
-  int nnz           = A.nnz();
-
-  lib_dispatch::sparse_gemv<resrc>(&transa, &m, &n, &nnz, A.offsets(),
-                                   A.columns(), A.data(), &alpha_, x.data(),
-                                   &beta_, y.data());
+  lib_dispatch::sparse_gemv<resrc>(transa, A.nrows(), A.ncols(), A.nnz(),
+                                   A.offsets(), A.columns(), A.data(), alpha,
+                                   x.data(), beta, y.data());
 
   return y;
 }
--- a/src/lib_dispatch.cpp
+++ b/src/lib_dispatch.cpp
@@ -1254,24 +1254,24 @@ void sparse_gemv(char const trans, int rows, int cols, int nnz,
           const_cast<int *>(col_indices), const_cast<P *>(vals),
           CUSPARSE_INDEX_32I, CUSPARSE_INDEX_32I, CUSPARSE_INDEX_BASE_ZERO,
           CUDA_R_64F);
-      expect(status == 0);
+      expect(status == CUSPARSE_STATUS_SUCCESS);
 
       status = cusparseCreateDnVec(&vecX, cols, const_cast<P *>(x), CUDA_R_64F);
-      expect(status == 0);
+      expect(status == CUSPARSE_STATUS_SUCCESS);
 
       status = cusparseCreateDnVec(&vecY, rows, const_cast<P *>(y), CUDA_R_64F);
-      expect(status == 0);
+      expect(status == CUSPARSE_STATUS_SUCCESS);
 
       // find tmp buffer size if needed
       size_t buffer_size = 0;
       status             = cusparseSpMV_bufferSize(
           device.get_sp_handle(), cusparse_trans(trans), &alpha, mat, vecX,
           &beta, vecY, CUDA_R_64F, CUSPARSE_SPMV_CSR_ALG2, &buffer_size);
-      expect(status == 0);
+      expect(status == CUSPARSE_STATUS_SUCCESS);
 
       void *sp_buffer = NULL;
       auto success    = cudaMalloc(&sp_buffer, buffer_size);
-      expect(success == 0);
+      expect(success == cudaSuccess);
 
       // call the sparse grid dense vector multiply
       // using CSR_ALG2 since it provides deterministic bit-wise results for
@@ -1279,10 +1279,20 @@ void sparse_gemv(char const trans, int rows, int cols, int nnz,
       status = cusparseSpMV(device.get_sp_handle(), cusparse_trans(trans),
                             &alpha, mat, vecX, &beta, vecY, CUDA_R_64F,
                             CUSPARSE_SPMV_CSR_ALG2, sp_buffer);
-      expect(status == 0);
+      expect(status == CUSPARSE_STATUS_SUCCESS);
 
       success = cudaFree(sp_buffer);
-      expect(success == 0);
+      expect(success == cudaSuccess);
+
+      // destroy the cuSparse desriptors
+      status = cusparseDestroyDnVec(vecX);
+      expect(status == CUSPARSE_STATUS_SUCCESS);
+
+      status = cusparseDestroyDnVec(vecY);
+      expect(status == CUSPARSE_STATUS_SUCCESS);
+
+      status = cusparseDestroySpMat(mat);
+      expect(status == CUSPARSE_STATUS_SUCCESS);
     }
     else if constexpr (std::is_same<P, float>::value)
     {
--- a/src/lib_dispatch.cpp
+++ b/src/lib_dispatch.cpp
@@ -1,6 +1,7 @@
 #include "lib_dispatch.hpp"
 #include "build_info.hpp"
 #include "distribution.hpp"
+#include "sparse.hpp"
 #include "tensors.hpp"
 #include "tools.hpp"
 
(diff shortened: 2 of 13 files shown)


src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+        col_indices_{col_indices}, values_{values}
+  {}
+
+  ~sparse() {}
+
+  // create sparse matrix from dense matrix
+  template<resource r_ = resrc, typename = enable_for_host<r_>>
+  sparse(fk::matrix<P, mem, resrc> const &m)
+  {
+    P constexpr tol = 1.0e-10;
```
stefan-schnake: Why is this bumped up so high over machine epsilon?  Is this due to the error from the wavelet transform?

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+
+    row_offsets_[0] = 0;
+    for (int i = 0; i < nnz_; i++)
+    {
+      col_indices_[i]     = i;
+      row_offsets_[i + 1] = i + 1;
+    }
+  }
+
+  // template<mem_type m_ = mem, typename = enable_for_owner<m_>,
```
stefan-schnake: Delete

src/lib_dispatch.cpp
```diff
@@ -1075,6 +1100,102 @@ void scalapack_getrs(char *trans, int *n, int *nrhs, P const *A, int *descA,
 }
 #endif
 
+template<resource resrc, typename P>
+void sparse_gemv(char const *trans, int *rows, int *cols, int *nnz,
```
mkstoyanov: WHY ARE THESE BEING PASSES IN AS POINTERS?

trans, rows, cols, nnz, etc, should be passed in by value

src/lib_dispatch.cpp
```diff
@@ -1075,6 +1100,102 @@ void scalapack_getrs(char *trans, int *n, int *nrhs, P const *A, int *descA,
...
+                 const int *offsets, const int *columns, const P *A, P *alpha,
```
mkstoyanov: Better names here are `row_offsets, `col_indexes` and `vals` (for A).

src/lib_dispatch.cpp
```diff
@@ -1075,6 +1100,102 @@ void scalapack_getrs(char *trans, int *n, int *nrhs, P const *A, int *descA,
...
+  {
+    // TODO: only non-transpose case implemented
+    expect(*trans == 'n');
+
+    int col_index_offset = 0;
+    for (int i = 0; i < *rows; i++)
+    {
+      P sum = 0.0;
+      // sparse case, iterate over number of column entries in the current row
+      int num_in_row = offsets[i + 1] - offsets[i];
```
mkstoyanov: You're needlessly offsetting back and forth and you forgot alpha and beta:
```
for(int r=0; r<rows; i++)
{
  y[r] *= beta;
  for(int c=offsets[i]; c < offsets[i+1]; c++)
  {
    y[i] += alpha * A[c] * x[columns[c]];
  }
}
```

mkstoyanov: Also, the current implementation is not using `columns` and therefore it is wrong.

src/moment.cpp
```diff
@@ -189,17 +194,31 @@ void moment<P>::createMomentReducedMatrix_nd(PDE<P> const &pde,
...
           }
         }
       }
     }
   }
+
+  std::cout << " -- moment_mat map size = " << moment_mat.size() << "\n";
+
+  // create a sparse version of this matrix and put it on the GPU
+  // this->sparse_mat =
```
mkstoyanov: Delete

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+struct dense_item
+{
+  int row;
+  int col;
+  P data;
+};
+
+namespace fk
+{
+// forward declarations
```
mkstoyanov: Why? You are declaring the class right after the forward declare.

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+template<typename P, mem_type mem = mem_type::owner,
+         resource resrc = resource::host> // default to be an owner only on host
+class sparse;
+
+template<typename P, mem_type mem, resource resrc>
```
mkstoyanov: Getting a view of a sparse matrix is incredibly messy. Why do we need this capability and how are we even going to do it without a copy, given that cuSparse doesn't really support it.

Skip the `mem_type` unless we really need it (and we can discuss if it is needed).

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+
+  fk::vector<int, mem, resrc> get_offsets() const { return row_offsets_; }
+  fk::vector<int, mem, resrc> get_columns() const { return col_indices_; }
+  fk::vector<P, mem, resrc> get_values() const { return values_; }
+
+private:
+  int ncols_;
+  int nrows_;
+  int nnz_;
+  int nz_;
```
mkstoyanov: Confusing, what's the difference between `nnz_` and `nz_`?

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+  //
+  // basic queries to private data
+  //
+  int nrows() const { return nrows_; }
+  int ncols() const { return ncols_; }
+  int nnz() const { return nnz_; }
+  int64_t size() const { return int64_t{nrows()} * ncols(); }
+  int64_t sp_size() const { return int64_t{values_.size()}; }
```
mkstoyanov, line 337: How is this different from `nnz()`?

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+  fk::matrix<P, mem, resrc> to_dense() const
+  {
+    // create dense, filled with 0 initially
+    fk::matrix<P, mem, resrc> dense(nrows_, ncols_);
+
+    // populate entries of the dense matrix
+    int col_index_offset = 0;
+    for (int row = 0; row < nrows_; row++)
+    {
+      int num_in_row = row_offsets_[row + 1] - row_offsets_[row];
```
mkstoyanov: Adding and subtracting the row-indexes here, see my earlier comment for the sparse mat-vec multiply.

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+
+template<typename P, mem_type mem, resource resrc>
+class sparse
+{
+  template<typename, mem_type, resource>
+  friend class sparse;
+
+public:
+  template<mem_type m_ = mem, typename = enable_for_owner<m_>>
+  sparse()
```
mkstoyanov: Lets initialize the constants too, so that rows/cols/nnz are set to zero too.

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+  {}
+
+  explicit sparse(int nrows, int ncols, int nnz,
```
mkstoyanov, line 41: For the owner case, we need to add a constructor that accepts `r-val-refs` for all the vectors.

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+    values_ = fk::vector<P>(values);
+    nnz_    = values_.size();
+
+    expect(m.size() == values_.size() + nz_);
+  }
+
+  // create sparse matrix from multimap
+  sparse(std::multimap<int, dense_item<P>> const &items, int ncols, int nrows)
+  {
+    P constexpr tol = 1.0e-10;
```
mkstoyanov: The tolerance should depend on the precision, should it not? Also, what's the right number here? Maybe we need to discuss.

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+    nrows_ = nrows;
+    nz_    = 0;
+
+    row_offsets_.resize(nrows_ + 1);
+    row_offsets_[0] = 0;
+
+    std::vector<P> values;
+    std::vector<int> col_indices_tmp;
+
+    values.reserve(nrows_);
```
mkstoyanov, line 135: Can we run a quick loop and figure out the number of non-zeros first?

This reserves too little, `col_indices_tmp` are not reserved either. Then we can avoid relocation in `push` and the copy to `fk::vector`.

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+        values_{a.values_}
+  {}
+  */
+
+  // copy constructor
+  sparse(fk::sparse<P, mem, resrc> const &other)
+      : ncols_{other.ncols_}, nrows_{other.nrows_}, nnz_{other.nnz_},
+        nz_{other.nz_}
+  {
+    row_offsets_ = fk::vector<int, mem, resrc>(other.get_offsets());
```
mkstoyanov: Double initialization of `row_offsets_`, we can initialize directly with the ints above.

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+
+  // transfer functions
+  // host->dev, new matrix
+  template<resource r_ = resrc, typename = enable_for_host<r_>>
+  sparse<P, mem_type::owner, resource::device> clone_onto_device() const
+  {
+    auto offsets_dev = row_offsets_.clone_onto_device();
+    auto col_dev     = col_indices_.clone_onto_device();
+    auto val_dev     = values_.clone_onto_device();
+    return sparse<P, mem_type::owner, resource::device>(
```
mkstoyanov: This creates additional copy for the vectors. Can be avoided with a constructor accepting `r-val-refs`, then we don't need the temporary variables.

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+    else
+      expect(false); // above cases cover all implemented mem types
+
+    //  Print these out as row major even though stored in memory as column
+    //  major.
+    for (auto i = 0; i < nrows(); ++i)
+    {
+      for (auto j = 0; j < ncols(); ++j)
+      {
+        if (exists(i, j))
```
mkstoyanov: Is this the only place where `exists()` is used? It is not needed if you only loop over the values with indexes, there is no need to brute force the check over all entries.

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+        {
+          std::cout << "(" << std::setw(2) << i << ", " << std::setw(2) << j
+                    << ") ";
+          P const val = values_[col_indices_[row_offsets_[i]]];
+          if constexpr (std::is_floating_point<P>::value)
+          {
+            std::cout << std::setw(12) << std::setprecision(4)
```
mkstoyanov: Do this once, outside of the loops.

src/fast_math.hpp
```diff
@@ -458,4 +459,32 @@ void getrs(fk::matrix<P, amem> const &A, fk::scalapack_matrix_info &ainfo,
...
 
+// sparse gemv - sparse matrix dense vector multiplication
+template<typename P, mem_type amem, mem_type xmem, mem_type ymem,
+         resource resrc>
+fk::vector<P, ymem, resrc> &
+sparse_gemv(fk::sparse<P, amem, resrc> const &A,
+            fk::vector<P, xmem, resrc> const &x, fk::vector<P, ymem, resrc> &y,
+            bool const trans_A = false, P const alpha = 1.0, P const beta = 0.0)
+{
+  int const rows_A = trans_A ? A.ncols() : A.nrows();
```
mkstoyanov: This isn't `rows_A`, this is `rows_opA` or `expected_y_size`.

src/fast_math.hpp
```diff
@@ -458,4 +459,32 @@ void getrs(fk::matrix<P, amem> const &A, fk::scalapack_matrix_info &ainfo,
...
+  int const cols_A = trans_A ? A.nrows() : A.ncols();
+
+  expect(rows_A == y.size());
+  expect(cols_A == x.size());
+
+  P alpha_          = alpha;
```
mkstoyanov: This is painful, see comment later on.

src/sparse.hpp
```diff
@@ -0,0 +1,398 @@
...
+#include <vector>
+
+namespace asgard
+{
+template<typename P>
+struct dense_item
+{
+  int row;
+  int col;
+  P data;
```
mkstoyanov: It's `value` or `val`, not `data`.

src/sparse.hpp
mkstoyanov: OK, figured it out, but there's no point in keeping both of them. We need `nnz_`, although it is also the size of the `col_indices_` so there's no real reason to keep it, and `nz_` is not needed. We can compute it as `rows()*cols() - nnz()` in the unlikely case that we may actually need it.

mkstoyanov (changes requested):

src/sparse.hpp
stefan-schnake: Yes it should depend on precision.   Also I believe `1e-10` is fine as the coefficients for the moments are 0 for any level > 0.

src/sparse.hpp
```diff
@@ -0,0 +1,363 @@
...
+    return false;
+  }
+  template<resource r_ = resrc, typename = enable_for_host<r_>>
+  bool operator!=(sparse<P> const &other) const
+  {
+    return !(*this == other);
+  }
+
+  template<resource r_ = resrc, typename = enable_for_host<r_>>
+  void print(std::string const label = "") const
```
mkstoyanov, line 304: The string `label` is not used.

src/time_advance.cpp
```diff
@@ -512,14 +513,16 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
+    first_time = false;
+    asgard::tools::timer.stop("update_system");
   }
 
   auto do_poisson_update = [&](fk::vector<P> const &f_in) {
     tools::timer.start("poisson_update");
     // Get 0th moment
-    fk::vector<P> mom0(dense_size);
-    fm::gemv(pde.moments[0].get_moment_matrix(), f_in, mom0);
+    fk::vector<P, mem_type::owner, resource::device> mom0(dense_size);
```
mkstoyanov: This is the kind of code that won't compile on the CPU. Before we made some of the recent changes, if CUDA was not enabled then `device` defaulted to the CPU, but that can be misleading. With the new changes, `device` cannot be used without CUDA.

Have a variable defined with pre-processor macros:
```
#ifdef ASGARD_USE_CUDA
constexpr resouce imex_resc = resouce::device;
#else 
constexpr resouce imex_resc = resouce::host;
#endif
```

The code may need a few more of these, but that's the way to go.

src/sparse.hpp
ckendrick: Doing this messes up how the row, col indices get printed. I left this as-is for the moment and we can come back to it. This is only ever used for debugging, so this is minor.

src/sparse.hpp
ckendrick: Currently, yes. However `exists()` is used in future PRs. I did update the print function so that it no longer brute force checks every matrix element.

src/time_advance_tests.cpp
```diff
@@ -1263,7 +1263,7 @@ TEMPLATE_TEST_CASE("implicit time advance - continuity 2", "[time_advance]",
   }
 }
 
-TEMPLATE_TEST_CASE("IMEX time advance - landau", "[imex]", test_precs)
+TEMPLATE_TEST_CASE("IMEX time advance - landau", "[imex]", double)
```
mkstoyanov: I'm guessing this is related to the tolerance issue, but we cannot do that. This code will prevent ASGARD from compiling using only single precision.

src/lib_dispatch.cpp
```diff
@@ -1196,6 +1221,93 @@ void scalapack_getrs(char *trans, int *n, int *nrhs, P const *A, int *descA,
...
+    {
+      // create cuSPARSE descriptors
+      cusparseSpMatDescr_t mat;
+      cusparseDnVecDescr_t vecX, vecY;
+      auto status = cusparseCreateCsr(
+          &mat, rows, cols, nnz, const_cast<int *>(row_offsets),
+          const_cast<int *>(col_indices), const_cast<P *>(vals),
+          CUSPARSE_INDEX_32I, CUSPARSE_INDEX_32I, CUSPARSE_INDEX_BASE_ZERO,
+          CUDA_R_64F);
+      expect(status == 0);
```
mkstoyanov: This should check against `CUSPARSE_STATUS_SUCCESS`

src/lib_dispatch.cpp
```diff
@@ -1196,6 +1221,93 @@ void scalapack_getrs(char *trans, int *n, int *nrhs, P const *A, int *descA,
...
+      // find tmp buffer size if needed
+      size_t buffer_size = 0;
+      status             = cusparseSpMV_bufferSize(
+          device.get_sp_handle(), cusparse_trans(trans), &alpha, mat, vecX,
+          &beta, vecY, CUDA_R_64F, CUSPARSE_SPMV_CSR_ALG2, &buffer_size);
+      expect(status == 0);
+
+      void *sp_buffer = NULL;
+      auto success    = cudaMalloc(&sp_buffer, buffer_size);
+      expect(success == 0);
```
mkstoyanov: :) `success == 0` when usually we take `0` to mean `false`

I believe the thing to use here is `success == CUDA_STATUS_SUCCESS`

src/lib_dispatch.cpp
```diff
@@ -1196,6 +1221,93 @@ void scalapack_getrs(char *trans, int *n, int *nrhs, P const *A, int *descA,
...
+    // device-specific specialization if needed
+#ifdef ASGARD_USE_CUDA
+    // no non-fp blas on device
+    expect(std::is_floating_point_v<P>);
+
+    // instantiated for these two fp types
+    if constexpr (std::is_same<P, double>::value)
+    {
+      // create cuSPARSE descriptors
+      cusparseSpMatDescr_t mat;
```
mkstoyanov: The transients `mat`, `vecX` and `vecY` are actually pointers and have to be deleted.

mkstoyanov (changes requested):

src/lib_dispatch.cpp
```diff
@@ -1196,6 +1221,93 @@ void scalapack_getrs(char *trans, int *n, int *nrhs, P const *A, int *descA,
...
+      expect(status == 0);
+
+      // find tmp buffer size if needed
+      size_t buffer_size = 0;
+      status             = cusparseSpMV_bufferSize(
+          device.get_sp_handle(), cusparse_trans(trans), &alpha, mat, vecX,
+          &beta, vecY, CUDA_R_64F, CUSPARSE_SPMV_CSR_ALG2, &buffer_size);
+      expect(status == 0);
+
+      void *sp_buffer = NULL;
```
mkstoyanov: If we are using the same matrix over and over again, we can put `sp_buffer` and the `mat` descriptor into the main sparse-matrix class to avoid extra allocations.

We can do that in another PR, no need to delay this one for that change.

ckendrick: I definitely agree. This is future work for the `sparse` class. It would be good to store this matrix descriptor in the class when the `resource = device`.

src/time_advance_tests.cpp
ckendrick: This was because `sparse_gemv` was hardcoded for `double` only: https://github.com/project-asgard/asgard/blob/945ded51e48a8a69b1a4e6768f221d8614f033bd/src/lib_dispatch.cpp#L1297-L1300.

I updated `sparse_gemv` to work with both and restored this test back to using `test_precs`.

mkstoyanov (approved):

PR #602, opened 2023-08-02
Allow PDE levels to be 1

ckendrick:

## Proposed changes

The CLI parser and PDE/dimension classes currently force the level to be > 2. Some problems, such as the relaxation problem, only need level = 1 in the x dimension. Additional problems will need level = 0 in some dimensions, so this allows those cases to pass CLI parsing as well.

This removes the checks for level > 2 and allows the user to set a level to be 1.

- [x] Bugfix
- [ ] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist

### DIFF
--- a/src/asgard_dimension.hpp
+++ b/src/asgard_dimension.hpp
@@ -161,7 +161,7 @@ std::vector<dimension_description<precision>> cli_apply_level_degree_correction(
     for (int &l : levels)
     {
       l = cli_input.get_starting_levels()(counter++);
-      expect(l >= 1);
+      expect(l >= 0);
     }
   }
   auto const cli_degree = cli_input.get_degree();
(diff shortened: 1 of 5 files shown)


stefan-schnake:
Why can't we just implement level 0 in this PR?

mkstoyanov (dismissed):

ckendrick:
> Why can't we just implement level 0 in this PR?

I think level 0 should work as far as I can tell. There don't appear to be any further changes. I think the `expect` statements need to be updated in the PDE class to allow level = 0 which I can update.

mkstoyanov:
I think level 0 may be a special case, since there is no wavelet transform. I suspect trying 0 may break something, e.g., some size was assumed to be more than that or some index uses `level-1` or something along those lines.

ckendrick:
We've been using level 0 in the first dimension with no issues for a few 1x2v and 1x3v PDEs. We should add some tests to cover the level 0 case though (some of the future PDE PRs should cover this).

I updated this PR to allow level 0 to be set, but it checks that at least one other dimension has a level > 0.

mkstoyanov (approved):

PR #605, opened 2023-08-02
Add relaxation 1x1v PDE

ckendrick:

## Proposed changes

This adds a simple relaxation 1x1v PDE.

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist

### DIFF
--- a/src/pde/pde_relaxation_1x1v.hpp
+++ b/src/pde/pde_relaxation_1x1v.hpp
@@ -352,7 +352,7 @@ private:
     // TODO: these are constants since we want dt always based on dim 2,
     //  but there is no way to force a different dim for this function!
     // (Lmax - Lmin) / 2 ^ LevX * CFL
-    return (6.0 - (-6.0)) / std::pow(2, 3);
+    return (6.0 - (-6.0)) / std::pow(static_cast<P>(2), static_cast<P>(3));
   }
 
   /* problem contains no sources */
--- a/src/pde/pde_relaxation_1x1v.hpp
+++ b/src/pde/pde_relaxation_1x1v.hpp
@@ -48,8 +48,7 @@ private:
   {
     ignore(t);
     fk::vector<P> fx(x.size());
-    std::transform(x.begin(), x.end(), fx.begin(),
-                   [](P const x_v) -> P { return 0.5; });
+    std::fill(fx.begin(), fx.end(), P{0.5});
     return fx;
   }
 
@@ -58,8 +57,7 @@ private:
   {
     ignore(t);
     fk::vector<P> fx(x.size());
-    std::transform(x.begin(), x.end(), fx.begin(),
-                   [](P const x_v) -> P { return 0.5; });
+    std::fill(fx.begin(), fx.end(), P{0.5});
     return fx;
   }
 
@@ -318,10 +316,7 @@ private:
   {
     ignore(t);
     fk::vector<P> fx(x.size());
-    std::transform(x.begin(), x.end(), fx.begin(), [](P const x_v) -> P {
-      ignore(x_v);
-      return 1.0;
-    });
+    std::fill(fx.begin(), fx.end(), P{1.0});
     return fx;
   }
 
--- a/src/pde/pde_relaxation_1x1v.hpp
+++ b/src/pde/pde_relaxation_1x1v.hpp
@@ -65,15 +65,15 @@ private:
   initial_condition_dim_v_0_0(fk::vector<P> const &x, P const t = 0)
   {
     ignore(t);
-    const P theta       = 0.5;
+    P constexpr theta   = 0.5;
+    P constexpr ux      = -1.0;
     P const coefficient = 1.0 / std::sqrt(2.0 * PI * theta);
 
     fk::vector<P> fx(x.size());
-    std::transform(
-        x.begin(), x.end(), fx.begin(), [coefficient, theta](P const x_v) -> P {
-          const P ux = -1.0;
-          return coefficient * std::exp(-(0.5 / theta) * std::pow(x_v - ux, 2));
-        });
+    for (int i = 0; i < x.size(); i++)
+    {
+      fx[i] = coefficient * std::exp(-(0.5 / theta) * std::pow(x[i] - ux, 2));
+    }
     return fx;
   }
 
@@ -81,15 +81,15 @@ private:
   initial_condition_dim_v_0_1(fk::vector<P> const &x, P const t = 0)
   {
     ignore(t);
-    const P theta       = 0.5;
+    P constexpr theta   = 0.5;
+    P constexpr ux      = 2.0;
     P const coefficient = 1.0 / std::sqrt(2.0 * PI * theta);
 
     fk::vector<P> fx(x.size());
-    std::transform(
-        x.begin(), x.end(), fx.begin(), [coefficient, theta](P const x_v) -> P {
-          const P ux = 2.0;
-          return coefficient * std::exp(-(0.5 / theta) * std::pow(x_v - ux, 2));
-        });
+    for (int i = 0; i < x.size(); i++)
+    {
+      fx[i] = coefficient * std::exp(-(0.5 / theta) * std::pow(x[i] - ux, 2));
+    }
     return fx;
   }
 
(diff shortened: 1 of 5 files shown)


src/time_advance_tests.cpp
```diff
@@ -1655,6 +1655,118 @@ TEMPLATE_TEST_CASE("IMEX time advance - twostream - ASG", "[imex][adapt]",
...
+    if (i == opts.num_time_steps - 1)
+    {
+      fk::vector<TestType> const analytic_solution = sum_separable_funcs(
+          pde->exact_vector_funcs, pde->get_dimensions(), adaptive_grid,
+          transformer, degree, time + pde->get_dt());
+
+      // calculate L2 error between simulation and analytical solution
+      fk::vector<TestType> const diff = f_val - analytic_solution;
+      auto const L2                   = [&diff]() -> TestType {
+        asgard::fk::vector<TestType> squared(diff);
```
mkstoyanov: You're allocating two unnecessary vectors and creating one extra copy. How about this code:
```
TestType L2 = TestType{0};
for(int i=0; i<f_val.size(); i++)
{
  TestType diff = f_val[i] - analytic_solution[i];
  L2 += diff * diff;
}
L2 = std::sqrt(L2);
```

src/time_advance_tests.cpp
```diff
@@ -1655,6 +1655,118 @@ TEMPLATE_TEST_CASE("IMEX time advance - twostream - ASG", "[imex][adapt]",
...
+        std::transform(squared.begin(), squared.end(), squared.begin(),
+                       [](TestType const &elem) { return elem * elem; });
+        auto const mean = std::accumulate(squared.begin(), squared.end(), 0.0);
+        return std::sqrt(mean);
+      }();
+      auto const relative_error =
+          L2 / asgard::inf_norm(analytic_solution) * 100;
```
mkstoyanov: It's always confusing to me, are we multiplying the `inf_norm` by 100 or the result by 100 (or dividing by 100).

Either use `( )` or pull the 100 in front.

src/pde/pde_relaxation_1x1v.hpp
```diff
@@ -0,0 +1,360 @@
...
+
+  // Analytic solution
+  static fk::vector<P> exact_dim_v_0(fk::vector<P> const &x, P const t = 0)
+  {
+    ignore(t);
+    P const theta       = 2.75;
+    P const coefficient = 1.0 / std::sqrt(2.0 * PI * theta);
+
+    fk::vector<P> fx(x.size());
+    std::transform(
```
mkstoyanov: Don't use transform, try to make your code look like a math formula, it's the most expressive way to do math.
```
const P scaled_coeff = coefficient * std::exp(-(0.5 / theta);
for(int i=0; i<x.size(); i++)
{
  fx[i] =  scaled_coeff * std::pow(x[i] - 0.5, 2));
}
```

mkstoyanov (changes requested):

ckendrick:
@mkstoyanov Thanks for the comments, I agree and will update the code. Most of that was copy/pasted directly from [main.cpp](https://github.com/project-asgard/asgard/blob/fc019dddef0f1420c8497d0a12544c5098d7197f/src/main.cpp#L276-L307). Do we want to update that code in this PR, or do that as a follow-up PR?

mkstoyanov:
The overuse of `std::transform` is a persistent problem throughout the code. I think we should stop using it (unless it really makes sense) and the code will slowly change.

In the case of allocating too much when computing L2 norms, it should be changed but the PR doesn't matter. Maybe better to make a separate one.

src/time_advance_tests.cpp
```diff
@@ -1655,6 +1655,111 @@ TEMPLATE_TEST_CASE("IMEX time advance - twostream - ASG", "[imex][adapt]",
...
+    if (i == opts.num_time_steps - 1)
+    {
+      fk::vector<TestType> const analytic_solution = sum_separable_funcs(
+          pde->exact_vector_funcs, pde->get_dimensions(), adaptive_grid,
+          transformer, degree, time + pde->get_dt());
+
+      // calculate L2 error between simulation and analytical solution
+      TestType const L2 = calculate_l2(f_val, analytic_solution);
+      TestType const relative_error =
+          TestType{100.0} * (L2 / asgard::inf_norm(analytic_solution));
```
stefan-schnake: If we're going to do this correctly we should take the L2 norm of the analytic solution (not the current L^\infty norm)

testing/tests_general.hpp
```diff
@@ -259,6 +259,25 @@ P calculate_integral(asgard::fk::vector<P> const &input,
   P integral = grid_spacing * sum;
   return integral;
 }
+
```
stefan-schnake, line 262: We should have a generic L2 norm function and just pass in the error between the two solutions.

mkstoyanov: BLAS has discrete L-2 norm, which we can use. Computing actual integrals is more involved and we have to write the method ourselves.

I also want to do this without computing the difference between two vectors first, I want to get two vectors and compute the difference on-the-fly.

stefan-schnake: Sorry my comment was in the wrong place.  I was referring to discrete norms.  Having a dedicated function that computes the error is fine, but we need a fast (discrete) L2 norm as well.  Currently the adaptivity algorithm uses L^\infty instead of L2 -- I want that to be changed in the future.

src/pde/pde_relaxation_1x1v.hpp
```diff
@@ -0,0 +1,362 @@
...
+  }
+
+  static fk::vector<P> moment2_f1(fk::vector<P> const &x, P const t = 0)
+  {
+    ignore(t);
+
+    fk::vector<P> f(x.size());
+    for (int i = 0; i < x.size(); i++)
+    {
+      f[i] = std::pow(x[i], 2);
```
mkstoyanov:
```suggestion
      f[i] = x[i] * x[i];
```

src/pde/pde_relaxation_1x1v.hpp
```diff
@@ -0,0 +1,362 @@
...
+  {
+    ignore(t);
+    P constexpr theta   = 2.75;
+    P constexpr u1      = 0.5;
+    P const coefficient = 1.0 / std::sqrt(2.0 * PI * theta);
+
+    fk::vector<P> fx(x.size());
+    for (int i = 0; i < x.size(); i++)
+    {
+      fx[i] = coefficient * std::exp(-(0.5 / theta) * std::pow(x[i] - u1, 2));
```
mkstoyanov:
```suggestion
      fx[i] = coefficient * std::exp(-(0.5 / theta) * (x[i] - u1) * (x[i] - u1));
```

src/pde/pde_relaxation_1x1v.hpp
```diff
@@ -0,0 +1,362 @@
...
+  {
+    ignore(t);
+    P constexpr theta   = 0.5;
+    P constexpr ux      = 2.0;
+    P const coefficient = 1.0 / std::sqrt(2.0 * PI * theta);
+
+    fk::vector<P> fx(x.size());
+    for (int i = 0; i < x.size(); i++)
+    {
+      fx[i] = coefficient * std::exp(-(0.5 / theta) * std::pow(x[i] - ux, 2));
```
mkstoyanov:
```suggestion
      fx[i] = coefficient * std::exp(-(0.5 / theta) * (x[i] - ux) * (x[i] - ux));
```

src/pde/pde_relaxation_1x1v.hpp
```diff
@@ -0,0 +1,362 @@
...
+  {
+    ignore(t);
+    P constexpr theta   = 0.5;
+    P constexpr ux      = -1.0;
+    P const coefficient = 1.0 / std::sqrt(2.0 * PI * theta);
+
+    fk::vector<P> fx(x.size());
+    for (int i = 0; i < x.size(); i++)
+    {
+      fx[i] = coefficient * std::exp(-(0.5 / theta) * std::pow(x[i] - ux, 2));
```
stefan-schnake: Modify the (x-ux)^2 as per @mkstoyanov's recommendations.

stefan-schnake (changes requested):

src/pde/pde_relaxation_1x1v.hpp
stefan-schnake: Also change u1 to ux for consistency

mkstoyanov (approved):

stefan-schnake (approved):

PR #606, opened 2023-08-03
Refactor IMEX moment calculation and add moments for 1x2v

ckendrick:

## Proposed changes

This refactors the moment calculation in IMEX to be a single lambda function `calculate_moments` rather than having two copies of the same code in IMEX.

This also adds initial support for calculating moments for 1x2v PDEs.

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist

### DIFF
--- a/src/time_advance.cpp
+++ b/src/time_advance.cpp
@@ -643,7 +643,6 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
   // Create rho_2s
   fk::vector<P, mem_type::owner, imex_resrc> mom0(dense_size);
   fm::sparse_gemv(pde.moments[0].get_moment_matrix_dev(), x_dev, mom0);
-
   fk::vector<P> &mom0_real = pde.moments[0].create_realspace_moment(
       pde_1d, mom0, adaptive_grid_1d.get_table(), transformer, tmp_workspace);
   param_manager.get_parameter("n")->value = [&](P const x_v,
@@ -663,17 +662,67 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
            param_manager.get_parameter("n")->value(x_v, t);
   };
 
-  fk::vector<P, mem_type::owner, imex_resrc> mom2(dense_size);
-  fm::sparse_gemv(pde.moments[2].get_moment_matrix_dev(), x_dev, mom2);
-  fk::vector<P> &mom2_real = pde.moments[2].create_realspace_moment(
-      pde_1d, mom2, adaptive_grid_1d.get_table(), transformer, tmp_workspace);
-  param_manager.get_parameter("theta")->value = [&](P const x_v,
-                                                    P const t = 0) -> P {
-    P const u = param_manager.get_parameter("u")->value(x_v, t);
-    return (interp1(nodes, mom2_real, {x_v})[0] /
-            param_manager.get_parameter("n")->value(x_v, t)) -
-           std::pow(u, 2);
-  };
+  if (pde.num_dims > 2 && pde.moments.size() > 3)
+  {
+    // Calculate additional moments for PDEs with more than one velocity
+    // dimension
+
+    // moment 2 = moment1v2 = \int f v_{2} dv
+    fk::vector<P, mem_type::owner, imex_resrc> mom2(dense_size);
+    fm::sparse_gemv(pde.moments[2].get_moment_matrix_dev(), x_dev, mom2);
+    fk::vector<P> &mom2_real = pde.moments[2].create_realspace_moment(
+        pde_1d, mom2, adaptive_grid_1d.get_table(), transformer, tmp_workspace);
+
+    // u2 = \int_v f v_1 dv / n(x)
+    param_manager.get_parameter("u2")->value = [&](P const x_v,
+                                                   P const t = 0) -> P {
+      return interp1(nodes, mom2_real, {x_v})[0] /
+             param_manager.get_parameter("n")->value(x_v, t);
+    };
+
+    // moment 3 = moment2v1
+    fk::vector<P, mem_type::owner, imex_resrc> mom3(dense_size);
+    fm::sparse_gemv(pde.moments[3].get_moment_matrix_dev(), x_dev, mom3);
+    fk::vector<P> &mom3_real = pde.moments[3].create_realspace_moment(
+        pde_1d, mom3, adaptive_grid_1d.get_table(), transformer, tmp_workspace);
+
+    // moment 4 = moment2v2
+    fk::vector<P, mem_type::owner, imex_resrc> mom4(dense_size);
+    fm::sparse_gemv(pde.moments[4].get_moment_matrix_dev(), x_dev, mom4);
+    fk::vector<P> &mom4_real = pde.moments[4].create_realspace_moment(
+        pde_1d, mom4, adaptive_grid_1d.get_table(), transformer, tmp_workspace);
+
+    // theta = \frac{ \int f(v_1^2 + v_2^2) dv }{ 2n(x)} - 0.5 * (u_1^2(x) +
+    // u_2^2(x))
+    param_manager.get_parameter("theta")->value = [&](P const x_v,
+                                                      P const t = 0) -> P {
+      P const mom3_x = interp1(nodes, mom3_real, {x_v})[0];
+      P const mom4_x = interp1(nodes, mom4_real, {x_v})[0];
+
+      P const u1 = param_manager.get_parameter("u")->value(x_v, t);
+      P const u2 = param_manager.get_parameter("u2")->value(x_v, t);
+
+      P const n = param_manager.get_parameter("n")->value(x_v, t);
+
+      return (mom3_x + mom4_x) / (2.0 * n) -
+             0.5 * (std::pow(u1, 2) + std::pow(u2, 2));
+    };
+  }
+  else
+  {
+    // theta moment for 1x1v case
+    fk::vector<P, mem_type::owner, imex_resrc> mom2(dense_size);
+    fm::sparse_gemv(pde.moments[2].get_moment_matrix_dev(), x_dev, mom2);
+    fk::vector<P> &mom2_real = pde.moments[2].create_realspace_moment(
+        pde_1d, mom2, adaptive_grid_1d.get_table(), transformer, tmp_workspace);
+    param_manager.get_parameter("theta")->value = [&](P const x_v,
+                                                      P const t = 0) -> P {
+      P const u = param_manager.get_parameter("u")->value(x_v, t);
+      return (interp1(nodes, mom2_real, {x_v})[0] /
+              param_manager.get_parameter("n")->value(x_v, t)) -
+            std::pow(u, 2);
+    };
+  }
 
   // f2 now
   P const tolerance  = program_opts.gmres_tolerance;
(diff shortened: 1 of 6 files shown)


src/quadrature.cpp
```diff
@@ -170,8 +170,9 @@ legendre_weights(int const degree, no_deduce<P> const lower_bound,
     return std::max(minimum_quadrature, degree_in + 1);
   };
 
-  int const num_points =
-      use_degree_points ? degree : default_quad_number(degree);
+  // int const num_points =
```
stefan-schnake: What's the reason for this change?

src/time_advance.cpp
```diff
@@ -598,16 +612,101 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
 #endif
   };
 
+  auto calculate_moments =
+      [&](fk::vector<P, mem_type::owner, imex_resrc> const &f_in) {
+        fk::vector<P, mem_type::owner, imex_resrc> mom0(dense_size);
```
stefan-schnake, line 611: Insert comment `//  \int f dv`

src/time_advance.cpp
```diff
@@ -598,16 +612,101 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
+        fm::sparse_gemv(pde.moments[0].get_moment_matrix_dev(), f_in, mom0);
+        fk::vector<P> &mom0_real = pde.moments[0].create_realspace_moment(
+            pde_1d, mom0, adaptive_grid_1d.get_table(), transformer,
+            tmp_workspace);
+        param_manager.get_parameter("n")->value = [&](P const x_v,
```
stefan-schnake, line 617: Insert `// n = \int f dv`

src/time_advance.cpp
```diff
@@ -598,16 +612,101 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
+                                                      P const t = 0) -> P {
+          ignore(t);
+          return interp1(nodes, mom0_real, {x_v})[0];
+        };
+
+        // TODO: refactor into more generic function
+        fk::vector<P, mem_type::owner, imex_resrc> mom1(dense_size);
```
stefan-schnake, line 624: `// \int f v_x dv`

src/time_advance.cpp
```diff
@@ -598,16 +612,101 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
+        fm::sparse_gemv(pde.moments[1].get_moment_matrix_dev(), f_in, mom1);
+        fk::vector<P> &mom1_real = pde.moments[1].create_realspace_moment(
+            pde_1d, mom1, adaptive_grid_1d.get_table(), transformer,
+            tmp_workspace);
+        param_manager.get_parameter("u")->value = [&](P const x_v,
```
stefan-schnake, line 631: `// u_x = \int f v_x  dv / n`

src/time_advance.cpp
```diff
@@ -598,16 +612,101 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
+          return interp1(nodes, mom1_real, {x_v})[0] /
+                 param_manager.get_parameter("n")->value(x_v, t);
+        };
+
+        if (pde.num_dims > 2 && pde.moments.size() > 3)
+        {
+          // Calculate additional moments for PDEs with more than one velocity
+          // dimension
+
+          // moment 2 = moment1v2 = \int f v_{2} dv
```
stefan-schnake: `// \int f v_y dv`

src/time_advance.cpp
```diff
@@ -598,16 +612,101 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
+          fk::vector<P, mem_type::owner, imex_resrc> mom2(dense_size);
+          fm::sparse_gemv(pde.moments[2].get_moment_matrix_dev(), f_in, mom2);
+          fk::vector<P> &mom2_real = pde.moments[2].create_realspace_moment(
+              pde_1d, mom2, adaptive_grid_1d.get_table(), transformer,
+              tmp_workspace);
+
+          // u2 = \int_v f v_1 dv / n(x)
```
stefan-schnake: `// u_y =  \int f v_y dv / n`

src/time_advance.cpp
```diff
@@ -598,16 +612,101 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
+          param_manager.get_parameter("u2")->value = [&](P const x_v,
+                                                         P const t = 0) -> P {
+            return interp1(nodes, mom2_real, {x_v})[0] /
+                   param_manager.get_parameter("n")->value(x_v, t);
+          };
+
+          // moment 3 = moment2v1
```
stefan-schnake: `// \int f v_x^2 dv`

src/time_advance.cpp
```diff
@@ -598,16 +612,101 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
+          fk::vector<P, mem_type::owner, imex_resrc> mom3(dense_size);
+          fm::sparse_gemv(pde.moments[3].get_moment_matrix_dev(), f_in, mom3);
+          fk::vector<P> &mom3_real = pde.moments[3].create_realspace_moment(
+              pde_1d, mom3, adaptive_grid_1d.get_table(), transformer,
+              tmp_workspace);
+
+          // moment 4 = moment2v2
```
stefan-schnake: `// \int f v_y^2 dv`

src/time_advance.cpp
```diff
@@ -598,16 +612,101 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
+          fk::vector<P, mem_type::owner, imex_resrc> mom4(dense_size);
+          fm::sparse_gemv(pde.moments[4].get_moment_matrix_dev(), f_in, mom4);
+          fk::vector<P> &mom4_real = pde.moments[4].create_realspace_moment(
+              pde_1d, mom4, adaptive_grid_1d.get_table(), transformer,
+              tmp_workspace);
+
+          // theta = \frac{ \int f(v_1^2 + v_2^2) dv }{ 2n(x)} - 0.5 * (u_1^2(x)
```
stefan-schnake: Replace `*_1` and `*_2` with `*_x` and `*_y` respectively

stefan-schnake: Also remove the `(x)`

src/time_advance.cpp
```diff
@@ -598,16 +612,101 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
+
+            P const n = param_manager.get_parameter("n")->value(x_v, t);
+
+            return (mom3_x + mom4_x) / (2.0 * n) -
+                   0.5 * (std::pow(u1, 2) + std::pow(u2, 2));
+          };
+        }
+        else
+        {
+          // theta moment for 1x1v case
```
stefan-schnake, line 688: `// \theta = \int f v_x^2 dv / n - u_x^2`

src/time_advance.cpp
```diff
@@ -616,70 +715,47 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
-  operator_matrices[matrix_entry::imex_explicit].apply(1.0, x.data(), 0.0,
-                                                       fx.data());
 
+  fk::vector<P, mem_type::owner, imex_resrc> fx(x.size());
+  if constexpr (imex_resrc == resource::device)
+  {
+    fx_host = x.clone_onto_host();
+    operator_matrices[matrix_entry::imex_explicit].apply(
+        1.0, x.clone_onto_host().data(), 0.0, fx_host.data());
+    // fx = fx_host.clone_onto_device();
```
stefan-schnake: Delete

src/time_advance.cpp
```diff
@@ -616,70 +715,47 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
-  fk::vector<P, mem_type::owner, imex_resrc> x_dev;
-  if constexpr (imex_resrc == resource::device)
+  if constexpr (imex_resrc == resource::host)
   {
-    x_dev = x.clone_onto_device();
+    reduce_results(fx, reduced_fx, plan, get_rank());
+
+    fk::vector<P, mem_type::owner, resource::host> f_2s(x_orig.size());
+    exchange_results(reduced_fx, f_2s, elem_size, plan, get_rank());
+    fm::axpy(f_2s, x, dt); // x here is f(1)
```
stefan-schnake: We really need to rename the variable names for the stages.  It's a nightmare right now and will not extend easily to higher order IMEX schemes.

stefan-schnake (changes requested):

src/pde/pde_relaxation_1x1v.hpp
```diff
@@ -0,0 +1,362 @@
...
+  static P get_dt_(dimension<P> const &dim)
+  {
+    ignore(dim);
+    /* return dx; this will be scaled by CFL from command line */
+    // return std::pow(0.25, dim.get_level());
+
+    // TODO: these are constants since we want dt always based on dim 2,
+    //  but there is no way to force a different dim for this function!
+    // (Lmax - Lmin) / 2 ^ LevX * CFL
+    return (6.0 - (-6.0)) / std::pow(2, 3);
```
mkstoyanov: I'm not sure what rule allows this to compile, but given that `pow` is defined only for floating point numbers and that `2` and `3` are ints, I have no clue what precision this is using.

Should be `std::pow(P{2}, P{3})`

src/time_advance.cpp
```diff
@@ -495,13 +495,27 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
   int const elem_size  = static_cast<int>(std::pow(degree, pde.num_dims));
   int const N_elements = std::pow(2, level);
 
-  fk::vector<P> x(x_orig);
+  fk::vector<P, mem_type::owner, imex_resrc> x;
+  fk::vector<P, mem_type::owner, imex_resrc> x_orig_dev;
```
mkstoyanov: `x_orig_dev` is not needed.

First you make a copy of `x_orig`, then you hold on the copy and not use it, then only on line 796 you copy `x_orig_dev` back into `x`. Instead, you can use `x_orig` which is kept persistent throughout the call.

src/time_advance.cpp
```diff
@@ -616,70 +715,47 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
   tools::timer.start("explicit_1");
   auto const apply_id = tools::timer.start("kronmult - explicit");
-  fk::vector<P, mem_type::owner, resource::host> fx(x.size());
-  operator_matrices[matrix_entry::imex_explicit].apply(1.0, x.data(), 0.0,
-                                                       fx.data());
 
+  fk::vector<P, mem_type::owner, imex_resrc> fx(x.size());
+  if constexpr (imex_resrc == resource::device)
+  {
+    fx_host = x.clone_onto_host();
```
mkstoyanov: `imex_resrc` is not a template parameter, so we have to choose between `if-constexpr` statements and per-processor macros `#ifdef ASGARD_USE_CUDA`

Here `fx_host` is defined (and allocated) on line 500, but it is not used until this point and it will only ever be used when CUDA is enabled. So, we should replace the `if-constexpr` with a `#ifdef` and define `fx_host` here.

src/pde/pde_relaxation_1x1v.hpp
ckendrick: I agree, these changes are from the other PR though. I'll update it there. That branch is merged into this one for now to make sure these changes pass the test added by that PR until it gets merged in.

src/time_advance.cpp
```diff
@@ -495,13 +495,27 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
   int const elem_size  = static_cast<int>(std::pow(degree, pde.num_dims));
   int const N_elements = std::pow(2, level);
 
-  fk::vector<P> x(x_orig);
+  fk::vector<P, mem_type::owner, imex_resrc> x;
+  fk::vector<P, mem_type::owner, imex_resrc> x_orig_dev;
+  fk::vector<P, mem_type::owner, resource::host> fx_host;
+
+  if constexpr (imex_resrc == resource::device)
```
mkstoyanov: `imex_resrc` is not a template parameter, so we have to choose between `if-constexpr` statements and per-processor macros `#ifdef ASGARD_USE_CUDA`

```
#ifdef ASGARD_USE_CUDA
  fk::vector<P, mem_type::owner, imex_resrc> x = x_orig.clone_onto_device();
#else 
  fk::vector<P, mem_type::owner, imex_resrc> x = x_orig;
#endif
```

PS: `x_orig_dev` is not needed, see my other comment.

src/quadrature.cpp
ckendrick: This was the temporary fix for getting problems with high nu values to work properly. This hard codes the number of points to be degree, instead of 10. I think this fix can be done a different way inside IMEX but that hasn't been done yet.

testing/tests_general.hpp
```diff
@@ -259,6 +259,25 @@ P calculate_integral(asgard::fk::vector<P> const &input,
   P integral = grid_spacing * sum;
   return integral;
 }
+
+/**
+ * \brief Calculates the l2 between two vectors
+ */
+template<typename P>
+P calculate_l2(asgard::fk::vector<P> const &sol,
```
mkstoyanov: * `l2` is a bit misleading, is it discrete or continuous?
* what's l2-between
* what happens if you switch `sol` and `analytic` (nothing), but why are they called with different names?
```
/**
  * \brief Calculate the discrete l-2 norm of the different between two vectors.
  */
P nrm2_dist(asgard::fk::vector<P> const &x, asgard::fk::vector<P> const &y)
{
  ...
}
```
The `nrm2` uses the BLAS convention and `dist` indicates that it's the distance between two vectors, as opposed to acting on a single vector.

mkstoyanov (changes requested):

src/quadrature.cpp
stefan-schnake: What does `use_degree_points` default to and how is it changed?  Seems odd to hardcode functionality that we already have the ability to switch.

src/pde/pde_relaxation_1x1v.hpp
ckendrick: Just for reference, it looks like if `std::pow` is given ints, then the arguments are statically casted to double. See the notes section here (second bullet point): https://en.cppreference.com/w/cpp/numeric/math/pow#Notes

testing/tests_general.hpp
ckendrick: These changes are added in the relaxation 1x1v PR.

src/pde/pde_relaxation_1x1v.hpp
mkstoyanov: Good point about `pow` but there is still an issue, if it is not obvious. Small details like that can potentially mess things up when switching between precisions.

src/time_advance.cpp
ckendrick: I agree `x_orig_dev` is not needed, although if I replace the `fm::copy(x_orig_dev, x)` call with `fm::copy(x_orig, x)`, there are template issues since `x_orig` from the function signature has no resource template defined.

I think future work is to change IMEX (and probably all time advance routines) to have a template parameter for the resource types, so this would allow this to be removed.

mkstoyanov (approved):

stefan-schnake (approved):

PR #610, opened 2023-08-08
Fix number of quadrature points for real space

ckendrick:

## Proposed changes

The functions for handling real space transformations were originally hardcoded to use only `degree` number of quadrature points. This causes issues when calculating real space moment vectors in IMEX and leads to errors (for example, instabilities for large nu >= 100 used in problems like collisional landau) since those moments are not evaluated directly at the quadrature nodes.

This changes these functions to allow more quadrature nodes to be used and changes IMEX to use the same number of quadrature points that are being used to generate coefficients.  This reproduces the energy conservation when we use peg the number of quadrature points to `degree`.

- [x] Bugfix
- [ ] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

## Checklist


src/io.hpp
```diff
@@ -97,13 +97,19 @@ void generate_initial_moments(
   adapt::distributed_grid adaptive_grid_1d(pde_1d, program_opts);
 
   // Create workspace for wavelet transform
-  auto const dense_size = dense_space_size(pde_1d);
-  fk::vector<P, mem_type::owner, resource::host> workspace(dense_size * 2);
+  auto const dense_size      = dense_space_size(pde_1d);
+  auto const quad_dense_size = std::accumulate(
```
```suggestion
  int const quad_dense_size = std::accumulate(
```

src/moment.cpp
```diff
@@ -235,9 +235,13 @@ fk::vector<P> &moment<P>::create_realspace_moment(
     basis::wavelet_transform<P, resource::host> const &transformer,
     std::array<fk::vector<P, mem_type::view, resource::host>, 2> &workspace)
 {
-  this->realspace.resize(wave.size());
+  // the realspace size uses the number of quadrature points (10) rather than
+  // degree
+  int const realspace_size =
+      10 * std::pow(2, pde_1d.get_dimensions()[0].get_level());
```
quantumsteve:
```suggestion
      10 * fm::two_raised_to(pde_1d.get_dimensions()[0].get_level());
```

src/moment.cpp
```diff
@@ -251,9 +255,13 @@ fk::vector<P> &moment<P>::create_realspace_moment(
     std::array<fk::vector<P, mem_type::view, resource::host>, 2> &workspace)
 {
   fk::vector<P> wave_host = wave.clone_onto_host();
-  this->realspace.resize(wave_host.size());
+  // the realspace size uses the number of quadrature points (10) rather than
+  // degree
+  int const realspace_size =
+      10 * std::pow(2, pde_1d.get_dimensions()[0].get_level());
```
quantumsteve:
```suggestion
      10 * fm::two_raised_to (pde_1d.get_dimensions()[0].get_level());
```

src/moment.cpp
```diff
@@ -251,9 +255,13 @@ fk::vector<P> &moment<P>::create_realspace_moment(
...
+  this->realspace = fk::vector<P>(realspace_size);
```
quantumsteve, line 264: should this be a call to `fk::vector<P>::resize(...)`?

src/time_advance.cpp
```diff
@@ -562,16 +567,16 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
                      return param_manager.get_parameter("S")->value(x_v, 0.0);
                    });
 
-    fk::vector<P> phi(dense_size);
-    fk::vector<P> poisson_E(dense_size);
+    fk::vector<P> phi(quad_dense_size);
+    fk::vector<P> poisson_E(quad_dense_size);
     solver::poisson_solver(poisson_source, pde.poisson_diag,
-                           pde.poisson_off_diag, phi, poisson_E, degree - 1,
+                           pde.poisson_off_diag, phi, poisson_E, 10 - 1,
```
quantumsteve: Where does `10` come from?

src/transformations.cpp
```diff
@@ -121,16 +124,18 @@ std::vector<fk::matrix<P>> gen_realspace_transform(
...
-gen_realspace_nodes(int const degree, int const level, P const min, P const max)
+gen_realspace_nodes(int const degree, int const level, P const min, P const max,
+                    bool const use_degree_quad)
 {
-  int const n        = pow(2, level);
-  int const mat_dims = degree * n;
-  P const h          = (max - min) / n;
-  auto const lgwt    = legendre_weights<P>(degree, -1.0, 1.0, true);
-  auto const roots   = lgwt[0];
+  int const n      = pow(2, level);
```
quantumsteve:
```suggestion
  int const n      = fm::two_raised_to(level);
```

src/transformations.hpp
```diff
@@ -284,14 +288,19 @@ inline int dense_space_size(PDE<P> const &pde)
   return dense_space_size(pde.get_dimensions());
 }
 
+inline int dense_dim_size(int const &degree, int const &level)
```
quantumsteve:
```suggestion
inline int dense_dim_size(int const degree, int const level)
```
`int` is smaller than a pointer

src/transformations_tests.cpp
```diff
@@ -272,7 +272,7 @@ void test_gen_realspace_transform(PDE<P> const &pde,
       make_options({"-l", std::to_string(level), "-d", std::to_string(degree)});
   basis::wavelet_transform<P, resource::host> const transformer(opts, pde);
   std::vector<fk::matrix<P>> const transforms =
-      gen_realspace_transform(pde, transformer);
+      gen_realspace_transform(pde, transformer, true);
```
quantumsteve: Should we also test `false`?

quantumsteve (changes requested):
See comments

src/transformations_tests.cpp
mkstoyanov: I don't like the `true/false` input. Can we make it into an expressive `enum`

src/moment.cpp
mkstoyanov: I think resize will also copy the old data (or whatever portion of the old data it can copy). This will not do any copy.

Cole can correct me, if I'm wrong.

ckendrick: This used to be a call to `fk::vector<P>::resize(...)`, but was changed in trying to debug the issue found in #625 (the out-of-bounds issue got amplified by this PR since it extended the number of elements accessed beyond the vector size).

I think this is fine to change back to `resize`, but I believe Miro is correct - `resize` seems to reallocate and then copy the old data.
In this particular case, the data in the `realspace` vector is always overwritten by the next call to `wavelet_to_realspace`, so we don't need to copy the previous data. We should keep whichever is faster, but I'm guessing the resize is probably slower.

src/io.hpp
```diff
@@ -97,13 +97,19 @@ void generate_initial_moments(
   adapt::distributed_grid adaptive_grid_1d(pde_1d, program_opts);
 
   // Create workspace for wavelet transform
-  auto const dense_size = dense_space_size(pde_1d);
-  fk::vector<P, mem_type::owner, resource::host> workspace(dense_size * 2);
+  int const dense_size      = dense_space_size(pde_1d);
+  int const quad_dense_size = std::accumulate(
```
mkstoyanov:
```
int quad_dense_size = 1;
auto const &dims = pde_1d.get_dimensions();
for(int i=0; i<dims.size(); i++)
{
  quad_dense_size *= asgard::dense_dim_size(10, dims[i].get_level());
}
```

src/transformations_tests.cpp
ckendrick: I updated the `true/false` to `quadrature_mode::use_degree`/`quadrature_mode::use_fixed`.

mkstoyanov (approved):

quantumsteve (approved):

PR #611, opened 2023-08-08
Add moment calculations for 1x3v PDEs

ckendrick:

## Proposed changes

This adds moment calculations for 1x3v PDEs to the IMEX time advance function.

Refactoring `calculate_moments()` inside IMEX is planned and will be done in follow-up work which will combine the different cases for each dimension into a single function.

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist

### DIFF
--- a/src/solver.cpp
+++ b/src/solver.cpp
@@ -35,7 +35,9 @@ simple_gmres_euler(const P dt, kronmult_matrix<P> const &mat,
       [&](fk::vector<P, mem_type::owner, resrc> const &x_in,
           fk::vector<P, mem_type::owner, resrc> &y, P const alpha,
           P const beta) -> void {
+        auto const apply_id = tools::timer.start("kronmult - implicit");
         mat.template apply<resrc>(-dt * alpha, x_in.data(), beta, y.data());
+        tools::timer.stop(apply_id, mat.flops());
         int one = 1, n = y.size();
         lib_dispatch::axpy<resrc>(n, alpha, x_in.data(), one, y.data(), one);
       },
--- a/src/solver.cpp
+++ b/src/solver.cpp
@@ -115,6 +115,9 @@ simple_gmres(matrix_replacement mat, fk::vector<P, mem_type::owner, resrc> &x,
     return (norm == 0.0) ? static_cast<P>(1.0) : norm;
   }();
 
+  // controls how often the inner residual print occurs
+  int const print_freq = restart / 3;
+
   fk::vector<P, mem_type::owner, resrc> residual(b);
   auto const compute_residual = [&]() {
     P const alpha = -1.0;
@@ -248,7 +251,7 @@ simple_gmres(matrix_replacement mat, fk::vector<P, mem_type::owner, resrc> &x,
         break; // depart the inner iteration loop
       }
 
-      if (i % 5 == 0)
+      if (i % print_freq == 0)
       {
         std::cout << "   -- GMRES inner iteration " << i << " / " << restart
                   << " w/ residual " << error << std::endl;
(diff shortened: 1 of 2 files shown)


src/time_advance.cpp
```diff
@@ -683,6 +682,59 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
                    0.5 * (std::pow(u1, 2) + std::pow(u2, 2));
           };
         }
+        else if (pde.num_dims == 4 && pde.moments.size() > 3)
```
stefan-schnake:
```suggestion
        else if (pde.num_dims == 4 && pde.moments.size() > 6)
```

src/time_advance.cpp
```diff
@@ -683,6 +682,59 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
+        {
+          // Moments for 1X3V case
+          // TODO: this will be refactored to replace dimension cases in the
+          // future
+          std::vector<fk::vector<P, mem_type::owner, imex_resrc>> moments;
+          std::vector<fk::vector<P> *> moments_real;
+          // Create moment matrices and realspace moments for all moments in PDE
+          for (size_t mom = 2; mom < pde.moments.size(); mom++)
+          {
+            // \int f v_{mom} dv
```
stefan-schnake:
```suggestion
            // \int f v_x dv
```

src/time_advance.cpp
```diff
@@ -683,6 +682,59 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
...
+
+            P const u1 = param_manager.get_parameter("u")->value(x_v, t);
+            P const u2 = param_manager.get_parameter("u2")->value(x_v, t);
+            P const u3 = param_manager.get_parameter("u3")->value(x_v, t);
+
+            P const n = param_manager.get_parameter("n")->value(x_v, t);
+
+            return (mom4_x + mom5_x + mom6_x) / (3.0 * n) -
+                   (1.0 / 3.0) *
+                       (std::pow(u1, 2) + std::pow(u2, 2) + std::pow(u3, 2));
```
stefan-schnake:
```suggestion
                       (u1*u1 + u2*u2 + u3*u3);
```
I don't trust `std::pow`.

src/solver.cpp
```diff
@@ -245,6 +247,12 @@ simple_gmres(matrix_replacement mat, fk::vector<P, mem_type::owner, resrc> &x,
           fm::gemv(m, s_view, x, false, P{1.0}, P{1.0});
         break; // depart the inner iteration loop
       }
+
+      if (i % 5 == 0)
+      {
+        std::cout << "   -- GMRES inner iteration " << i << " / " << restart
+                  << " w/ residual " << error << std::endl;
+      }
```
stefan-schnake, line 258: I like this as an output, but it can get annoying.  Is there a `quiet' option like the matlab version had?

stefan-schnake (changes requested):

src/solver.cpp
ckendrick: I agree, I don't believe there is a quiet option implemented yet unfortunately. For now, I could decrease the frequency this prints (instead of every 5 inner iterations).

We should probably create an issue to make a "quiet" option as there is a lot of other print statements which might also be nice to disable for particular runs.

stefan-schnake: We could also set it to run every restart/3 iterations?  That way it's a fixed number per outer iteration.

mkstoyanov (approved):

stefan-schnake (approved):

PR #618, opened 2023-08-18
Low memory dense

mkstoyanov:
Another rework to kronmult, now there are in intermediate arrays with indexes, instead we are working directly with the multi-index sets.

The multi-index set is being copied onto the device and held there. The indexes are computed `on-the-fly` which dramatically reduces the memory footprint of the method. For example, level 8 full-stream dense-grid uses <1MB of RAM.

The negative side is that the performance drops significantly.


src/asgard_kronmult_matrix.cpp
```diff
@@ -86,198 +85,77 @@ make_kronmult_dense(PDE<precision> const &pde,
...
-    }
-    // if reached the end of the list and there is more to go
-    if (ia == ilist->end())
+    for (int d = 0; d < num_dimensions; d++)
     {
-      ilist++;
-      if (ilist != list_iA.end())
-        ia = ilist->begin();
+      elem[i * num_dimensions + d] =
+          (idx[d] == 0) ? 0 : ((1 << (idx[d] - 1)) + idx[d + num_dimensions]);
```
```suggestion
          (idx[d] == 0) ? 0 : (fm::two_raised_to(idx[d] - 1) + idx[d + num_dimensions]);
```

src/asgard_kronmult_matrix.hpp
```diff
@@ -167,15 +215,16 @@ class kronmult_matrix
   template<resource input_mode>
   kronmult_matrix(int num_dimensions, int kron_size, int num_rows, int num_cols,
                   int num_terms,
-                  fk::vector<int, mem_type::owner, input_mode> const &&row_indx,
-                  fk::vector<int, mem_type::owner, input_mode> const &&col_indx,
+                  fk::vector<int, mem_type::owner, input_mode> &&row_indx,
+                  fk::vector<int, mem_type::owner, input_mode> &&col_indx,
```
quantumsteve, line 219: When would someone use a const r-value reference? :thinking:

mkstoyanov: As far as I can tell, the `const` was only preventing the `move` semantics from working properly, i.e., we were copying instead of moving.

src/time_advance_tests.cpp
```diff
@@ -1269,6 +1269,7 @@ TEMPLATE_TEST_CASE("IMEX time advance - landau", "[imex]", test_precs)
   {
     return;
   }
+  return;
```
quantumsteve: @ckendrick Is this just a precision issue?

mkstoyanov: No, it was leftover from debugging. Good that you caught it.

src/device/asgard_kronmult_cyclex.hpp
```diff
@@ -18,10 +18,42 @@ __device__ constexpr int int_pow(int p)
...
+build_indexes(int64_t const num_batch, int const num_cols, int const elem[],
+              int const row_offset, int const col_offset,
+              int const num_1d_blocks, int iA[])
+{
+  int i = threadIdx.y + blockIdx.x * blockDim.y;
+
+  while (i < num_batch)
+  {
+    int const rowy = i / num_cols;
+    int const colx = i % num_cols;
```
ckendrick, line 37: Minor comment, but since `num_batch` is `int64_t` we should be careful with using regular `int`s here in case of overflow.

mkstoyanov: This kernel is currently inactive. We cannot do dense case in 6D with cubic basis (we can do quadratic or lower). I'm getting errors about registers when I try to compile.

Will fix the issue later.

ckendrick (approved):
This is definitely an improvement for now. Large problems can now be run on a small machine without breaking.

Additionally, running the Landau 1x3v FG level 4 with these new dense changes is slightly faster than the sparse version and is significantly faster than the previous dense version.

quantumsteve:
:shipit:

mkstoyanov:
> these new dense changes is slightly faster than the sparse version and is significantly faster than the previous dense version.

That's what I like to hear.

PR #619, opened 2023-08-22
Add preconditioner to GMRES and use block jacobi in IMEX

ckendrick:

## Proposed changes

Adds an initial version of a generic preconditioner class that can be used in GMRES. For IMEX, a block Jacobi preconditioner is implemented.

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist


src/lib_dispatch.hpp
```diff
@@ -59,17 +59,25 @@ void batched_gemm(P **const &a, int lda, char transa, P **const &b, int ldb,
                   char transb, P **const &c, int ldc, int m, int n, int k,
                   P alpha, P beta, int num_batch);
 
-template<typename P>
+template<resource resrc = resource::host, typename P>
 int gesv(int n, int nrhs, P *A, int lda, int *ipiv, P *b, int ldb);
 
+template<resource resrc = resource::device, typename P>
```
quantumsteve, line 65: I think it is better to maintain consistency and default to `resource::host`

src/lib_dispatch.hpp
```diff
@@ -59,17 +59,25 @@ void batched_gemm(P **const &a, int lda, char transa, P **const &b, int ldb,
...
 template<resource resrc = resource::host, typename P>
 void tpsv(const char uplo, const char trans, const char diag, const int n,
           const P *ap, P *x, const int incx);
 
-template<typename P>
+template<resource resrc = resource::host, typename P>
 int getrs(char trans, int n, int nrhs, P const *A, int lda, int const *ipiv,
           P *b, int ldb);
 
+template<resource resrc = resource::device, typename P>
```
quantumsteve, line 80: I think it is better to maintain consistency and default to `resource::host`

src/lib_dispatch.cpp
```diff
@@ -966,15 +978,203 @@ int gesv(int n, int nrhs, P *A, int lda, int *ipiv, P *b, int ldb)
...
+  static_assert(std::is_same_v<P, double> or std::is_same_v<P, float>);
 
   int info{1};
-  if constexpr (std::is_same<P, double>::value)
+
+  if constexpr (resrc == resource::device)
   {
-    dgesv_(&n, &nrhs, A, &lda, ipiv, b, &ldb, &info);
+    // device-specific specialization if needed
+#ifdef ASGARD_USE_CUDA
```
quantumsteve, line 1018: cusolver has a gesv function
https://docs.nvidia.com/cuda/cusolver/index.html?highlight=gesv#cusolverdn-t1-t2-gesv

src/lib_dispatch.cpp
```diff
@@ -966,15 +978,203 @@ int gesv(int n, int nrhs, P *A, int lda, int *ipiv, P *b, int ldb)
...
+
+    success = cudaFree(dinfo);
+    expect(success == cudaSuccess);
+#endif
   }
-  else if constexpr (std::is_same<P, float>::value)
+  else if constexpr (resrc == resource::host)
   {
-    sgesv_(&n, &nrhs, A, &lda, ipiv, b, &ldb, &info);
+    if constexpr (std::is_same<P, double>::value)
```
quantumsteve:
```suggestion
    if constexpr (std::is_same_v<P, double>)
```

src/lib_dispatch.cpp
```diff
@@ -966,15 +978,203 @@ int gesv(int n, int nrhs, P *A, int lda, int *ipiv, P *b, int ldb)
...
+    {
+      dgesv_(&n, &nrhs, A, &lda, ipiv, b, &ldb, &info);
+    }
+    else if constexpr (std::is_same<P, float>::value)
```
quantumsteve:
```suggestion
    else if constexpr (std::is_same_v<P, float>)
```

src/lib_dispatch.cpp
```diff
@@ -966,15 +978,203 @@ int gesv(int n, int nrhs, P *A, int lda, int *ipiv, P *b, int ldb)
...
+    expect(stat == cudaSuccess);
+
+    stat = cudaFree(A_d);
+    expect(stat == cudaSuccess);
+    stat = cudaFree(b_d);
+    expect(stat == cudaSuccess);
+    stat = cudaFree(info_d);
+    expect(stat == cudaSuccess);
+#endif
   }
```
quantumsteve, line 1197: CPU implmentation?

src/lib_dispatch.cpp
ckendrick: I think you already fixed most of these, I need to rebase to get those changes in this branch.

src/lib_dispatch.cpp
```diff
@@ -1113,16 +1313,129 @@ int getrs(char trans, int n, int nrhs, P const *A, int lda, int const *ipiv,
...
+    cusolverStatus_t status;
+
+    cudaDataType_t constexpr fp_prec =
+        std::is_same_v<P, double> ? CUDA_R_64F : CUDA_R_32F;
+
+    cusolverDnParams_t params;
+    status = cusolverDnCreateParams(&params);
+    expect(status == CUSOLVER_STATUS_SUCCESS);
+
+    status = cusolverDnSetAdvOptions(params, CUSOLVERDN_GETRF, CUSOLVER_ALG_0);
```
quantumsteve: Where are we not using the default solver?

src/lib_dispatch.cpp
ckendrick: I was originally using that function ([see this commit](https://github.com/project-asgard/asgard/pull/619/commits/9cb9dbd6d21d25f203994c1b7290192f4da8f09e) for that implementation) but ran into issues with it. It is marked on their API as legacy, so I'm not sure what the long-term support will be (it may be removed on future CUDA releases?)

The main issue was that it leaves the A matrix unchanged if it converged to a solution. I believe our code assumes that `A` is updated with the factorization `A = P * L * U` on exit, but this was not the case.

src/lib_dispatch.cpp
```diff
@@ -1113,16 +1313,129 @@ int getrs(char trans, int n, int nrhs, P const *A, int lda, int const *ipiv,
...
+    // no non-fp blas on device
+    expect(std::is_floating_point_v<P>);
+
+    cusolver_int_t *dinfo;
+    auto success = cudaMalloc(&dinfo, sizeof(cusolver_int_t));
+    assert(success == cudaSuccess);
+
+    status = cusolverDnXgetrs(device.get_solver_handle(), params,
+                              cublas_trans(trans), n, nrhs, fp_prec, A, lda,
+                              (int64_t *)(const_cast<int *>(ipiv)), fp_prec, b,
```
quantumsteve: Casting `int* ` to `int64_t` won't work.

src/lib_dispatch.cpp
```diff
@@ -1113,16 +1313,129 @@ int getrs(char trans, int n, int nrhs, P const *A, int lda, int const *ipiv,
...
+
+    status = cusolverDnSetAdvOptions(params, CUSOLVERDN_GETRF, CUSOLVER_ALG_0);
+    expect(status == CUSOLVER_STATUS_SUCCESS);
+
+    // no non-fp blas on device
+    expect(std::is_floating_point_v<P>);
+
+    cusolver_int_t *dinfo;
+    auto success = cudaMalloc(&dinfo, sizeof(cusolver_int_t));
+    assert(success == cudaSuccess);
```
quantumsteve: expect

src/lib_dispatch.cpp
```diff
@@ -1113,16 +1313,129 @@ int getrs(char trans, int n, int nrhs, P const *A, int lda, int const *ipiv,
...
+
+    status = cusolverDnXgetrs(device.get_solver_handle(), params,
+                              cublas_trans(trans), n, nrhs, fp_prec, A, lda,
+                              (int64_t *)(const_cast<int *>(ipiv)), fp_prec, b,
+                              ldb, dinfo);
+
+    if (status != CUSOLVER_STATUS_SUCCESS)
+    {
+      throw std::runtime_error("Error in CUSOLVER getrs:");
```
quantumsteve: Tempted to keep this a debug assertion and only check `info` in Release. What do you think?

src/lib_dispatch.cpp
```diff
@@ -1113,16 +1313,129 @@ int getrs(char trans, int n, int nrhs, P const *A, int lda, int const *ipiv,
...
+  expect(info == cudaSuccess);
+
+  stat = cudaFree(A_d);
+  expect(stat == cudaSuccess);
+  stat = cudaFree(b_d);
+  expect(stat == cudaSuccess);
+#endif
+
   return info;
 }
```
quantumsteve, line 1471: host implementation?

src/lib_dispatch.cpp
mkstoyanov: We probably want that, a single if-statement doesn't matter for a large/complex method and it can save us when we bump up the problem size (eventually).

The debug-mode only assertions should be fast ones, allocate/free/copy, etc.

src/preconditioner.hpp
```diff
@@ -0,0 +1,387 @@
...
+    int const block_size = std::pow(this->degree, this->num_dims);
+    int const offset     = block_index * block_size;
+
+    if constexpr (resrc == resource::host)
+    {
+      // extract the given block from the preconditioner matrix
+      auto B_block =
+          fk::vector<P, mem_type::view>(B, offset, offset + block_size - 1);
+
+      if (!this->is_factored)
```
mkstoyanov: Don't like this. We should factorize in the constructor and the apply method should be `const`.

src/preconditioner.hpp
```diff
@@ -0,0 +1,387 @@
...
+    if (!this->is_factored)
+    {
+      this->is_factored = true;
+    }
+    asgard::tools::timer.stop(id);
+  }
+#endif
+
+  virtual fk::matrix<P, mem_type::owner, resource::host>
+  get_matrix() const override
```
mkstoyanov, line 302: This looks like it's being used only in testing, why have it virtual? The way I see it, the only virtual method should be `apply()`.

src/fast_math.hpp
```diff
@@ -221,8 +221,50 @@ void gesv(fk::matrix<P, amem> &A, fk::vector<P, bmem> &B,
   int lda = A.stride();
   int ldb = B.size();
 
-  int info = lib_dispatch::gesv(rows_A, cols_B, A.data(), lda, ipiv.data(),
-                                B.data(), ldb);
+  int info = lib_dispatch::gesv<resource::host>(rows_A, cols_B, A.data(), lda,
+                                                ipiv.data(), B.data(), ldb);
+  if (info < 0)
```
mkstoyanov: In the good scenario, this will yield two checks, `info > 0` and `info < 0`

A better version is:
```
if (info == 0)
{
  if (info > 0)
  {
  }
  else
  {
  }
}
```

src/fast_math.hpp
```diff
@@ -221,8 +221,50 @@ void gesv(fk::matrix<P, amem> &A, fk::vector<P, bmem> &B,
...
+
+  int rows_ipiv = ipiv.size();
+  expect(cols_A == rows_B);
+  expect(rows_ipiv >= rows_A);
+
+  int lda = A.stride();
+  int ldb = B.size();
+
+  int info = lib_dispatch::gesv<resrc>(rows_A, cols_B, A.data(), lda,
+                                       (int *)ipiv.data(), B.data(), ldb);
```
mkstoyanov: This will not work, `int64_t*` and `int*` are not compatible.

mkstoyanov:
@stefan-schnake Are we still using any of this?

We should close this unless you are still using the code. If we want block-Jacobi preconditioner we will have to redo it in the current data-structures anyway.

stefan-schnake:
Close

mkstoyanov:
closed without prejudice

PR #621, opened 2023-08-23
Update restart for max_adapt_levels

ckendrick:

## Proposed changes

This saves the `--max_adapt_levels` option to file and loads it on restart. Updates the file restart method to handle adaptivity better and check for FG/SG mismatch.

This also changes the initial element table allocation size to be more reasonable when using sparse grids instead of always trying to reserve the FG DOF.

- [x] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Ubuntu 20.04

## Checklist


src/io.hpp
```diff
@@ -354,6 +358,16 @@ void read_restart_metadata(parser &user_vals, std::string const &restart_file)
   }
   int const max_level = H5Easy::load<int>(file, std::string("max_level"));
   int const dof       = H5Easy::load<int>(file, std::string("dof"));
+  bool const use_fg   = H5Easy::load<bool>(file, std::string("using_fullgrid"));
+
+  if (use_fg != user_vals.using_full_grid())
+  {
+    throw std::runtime_error("Mismatch of grid type between CLI and restart "
```
mkstoyanov: Are we sure about this? I mean, whatever grid we read from the file that's what we should use. The only question should be whether we turn adaptivity on/off and even in that case we should use whatever is in the CLI.

The grid that we specify in the CLI should only apply to the initial grid of the problem. What do you guys think?

stefan-schnake: > The grid that we specify in the CLI should only apply to the initial grid of the problem

I agree.

src/elements.cpp
stefan-schnake, line 1: When restarting, the dofs are known ahead of time.  Can we use this information here?

stefan-schnake (changes requested):

src/io.hpp
ckendrick: Yes, you are right - we only use the grid that is loaded from the file. However, the code does create the grid based on the CLI first, and [then re-creates it if restarting from file](https://github.com/project-asgard/asgard/blob/5715625b08cee81247ce94bcbdfccafad42ece07/src/main.cpp#L213-L224).

This is was meant more as a sanity check to let the user know if the grid options they provide are ignored. For example, if they added a `-f` when loading an adaptive grid, that the `-f` option on the CLI is not used. 

The current behavior is to just silently ignore (minus creating the full grid initially before it gets overwritten by the restart) - we can remove this check if that is the preferred behavior.

mkstoyanov: Remove the check and show a warning, but only if the user explicitly specifies a grid and it is not the one in the file. For example, the default is sparse-grid so if the user just says reload and the file contains a full-grid, then there should be no warning. If the user explicitly says `-f` and the grid is not full, then we can give a warning about ignored CLI parameters.

stefan-schnake (dismissed):

src/io.hpp
```diff
@@ -363,9 +377,37 @@ void read_restart_metadata(parser &user_vals, std::string const &restart_file)
...
+    for (const int &lev : max_adapt_levels)
+    {
+      max_adapt_str += std::to_string(lev) + " ";
+    }
+  }
+
   std::cout << "  - PDE: " << pde_string << ", ndims = " << ndims
             << ", degree = " << degree << "\n";
   std::cout << "  - time = " << time << ", dt = " << dt << "\n";
+  std::cout << "  - adaptivity = " << (do_adapt ? "true" : "false") << "\n";
```
```suggestion
  std::cout << "  - adaptivity = " << (do_adapt ? "true" : "false") << '\n';
```

src/io.hpp
```diff
@@ -363,9 +377,37 @@ void read_restart_metadata(parser &user_vals, std::string const &restart_file)
...
+        H5Easy::load<std::vector<int>>(file, std::string("max_adapt_levels"));
+    assert(max_adapt_levels.size() == static_cast<size_t>(ndims));
+
+    parser_mod::set(user_vals, parser_mod::max_adapt_level,
+                    fk::vector<int>(max_adapt_levels));
+
+    for (const int &lev : max_adapt_levels)
+    {
+      max_adapt_str += std::to_string(lev) + " ";
+    }
```
quantumsteve: this leaves an unnecessary space at the end.

src/io.hpp
```diff
@@ -363,9 +377,37 @@ void read_restart_metadata(parser &user_vals, std::string const &restart_file)
...
+  }
+
   std::cout << "  - PDE: " << pde_string << ", ndims = " << ndims
             << ", degree = " << degree << "\n";
   std::cout << "  - time = " << time << ", dt = " << dt << "\n";
+  std::cout << "  - adaptivity = " << (do_adapt ? "true" : "false") << "\n";
+  if (do_adapt)
+  {
+    std::cout << "    - max_level = " << max_level
+              << ", max_adapt_levels = " << max_adapt_str << "\n";
```
```suggestion
              << ", max_adapt_levels = " << max_adapt_str << '\n';
```

stefan-schnake (approved):

PR #622, opened 2023-08-23
Move matrix and vector to separate files

quantumsteve:

## Proposed changes

This splits the file tensors.hpp into 4 files: matrix.hpp, matrix.tpp, vector.hpp, vector.tpp.

- [ ] Bugfix
- [ ] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

## Checklist


src/vector.tpp
```diff
@@ -0,0 +1,703 @@
+//
+// This would otherwise be the start of the tensors.cpp, if we were still doing
+// the explicit instantiations
+//
+
+#include <algorithm>
```
mkstoyanov: LOL, new extension .tpp

This is a mistake, right?

src/asgard_dimension.hpp
```diff
@@ -13,9 +13,10 @@
 
 #include "fast_math.hpp"
 #include "matlab_utilities.hpp"
+#include "matrix.hpp"
 #include "program_options.hpp"
-#include "tensors.hpp"
 #include "tools.hpp"
+#include "vector.hpp"
```
mkstoyanov: Can we namespace the vector and matrix files.

I had in mind a small change to Cmake to allow modules to be called vector and matrix together with the test files, but the actual source files to be prefixed with asgard. Namely, I don't want to type asgard when I want to call kronmult_matrix-test, but still have the asgard prefix for the names of the source files.

src/vector.tpp
mkstoyanov: There is also matrix.tpp

mkstoyanov (changes requested):
Also, check with @ckendrick to make sure this won't mess up with his changes when rebasing.

src/vector.tpp
quantumsteve: What extension would you prefer? In the past I've used `.tcc` but `.ipp` (boost's convention) and `.tpp` fit within our current clang-format rules.

https://stackoverflow.com/questions/44774036/why-use-a-tpp-file-when-implementing-templated-functions-and-classes-defined-i

mkstoyanov: Sorry, I was making a joke, I didn't realize that you actually did that on purpose and there is a convention.

Although, I am still baffled by the point in it. Why split the files when you can have just one .hpp file and have all templates defined in line. Then I don't have to jump back and forth between definition and implementation, especially when many methods are just onelinesrs.

src/vector.tpp
```diff
@@ -0,0 +1,703 @@
...
+
+  vector const &X = (*this);
+  vector<P> Y(A.ncols());
+
+  int m     = A.nrows();
+  int n     = A.ncols();
+  int lda   = A.stride();
+  int one_i = 1;
+
+  P zero = 0.0;
```
mkstoyanov: Can we inline all those variables in the call to lib dispatch? We don't need a separate variables for one and zero anymore

src/vector.tpp
```diff
@@ -0,0 +1,703 @@
...
+  expect(i < size_);
+  return data_[i];
+}
+
+// vector comparison operators - set default tolerance above
+// see https://stackoverflow.com/a/253874/6595797
+// FIXME do we need to be more careful with these fp comparisons?
+template<typename P, mem_type mem, resource resrc>
+template<mem_type omem, resource, typename>
+bool fk::vector<P, mem, resrc>::operator==(vector<P, omem> const &other) const
```
mkstoyanov: Are we using this, we have a compare method in testing that allows for custom tolerance, I don't think we should have this as a member

mkstoyanov: Ugh, it is in a million places in testing. Leave it be.

src/vector.tpp
quantumsteve: It can give more control over instantiation (though not in the way I'm currently including vector.tpp and matrix.tpp). 

https://softwareengineering.stackexchange.com/questions/373916/c-preferred-method-of-dealing-with-implementation-for-large-templates/375323#375323

Since the definition and implementation was already split in tensors.hpp, I wanted the filenames to show that and personally I find it easier to find a different file than scroll hundreds of lines down to the implementation.

mkstoyanov: I never liked the old split in `tensors.hpp` it just meant going up and down to see what's going on. Having two files is better than one split file, but I still prefer the single inline file.

We should be considering this not from a general philosophical, but with respect to the specific use-case. The matrix and vector classes have many methods, but most of those are a few short lines, e.g., some expect clauses and one-two commands. Splitting the files does not make things "cleaner", it just means we have to switch to another file to see the short implementation, which may just as well be appended to the declaration. We don't really get any reduction in compile time, i.e., the methods are small and should be visible globally so they can be properly inlined. I don't see any benefit in splitting the files, compared to inlining.

Our options are the following:
* old way of splitting the declaration and implementation in the same file, which I think is the worst case scenario
* current way of splitting into two files, looking into a separate file is better than scrolling the same file
* inlining the implementation into the file, same way that STL does it, which means we need only to scroll for a few lines (everything is compact)

My vote goes for the last one, it is more work for you to make the PR, but I think it is the right way to do this. If there was a huge benefit from splitting vector and matrix type of classes, then STL would have done it for `std::vector` and `std::span`.

On a general philosophical note, if I have a class that contains methods with complex logic (many many lines), I prefer to express the logic in a functional manner in another file, then the class method becomes just a simple call to an external function passing in the relevant member-data. This is easier to reuse and generalize across algorithms, maybe it's just a feature of math-libraries, but it works pretty well. If I really have to write long/hard logic and I want to limit compile times, then I split the class into `.cpp` and `.hpp` (we don't want that for matrix and vector).

I'm not saying `.tpp` files have no use-case, I cannot make such a clam. I can only claim that it is fringe compared to what we do and it doesn't really help for the vector and matrix classes.

quantumsteve: I removed `matrix.tpp` and `vector.tpp`

mkstoyanov (approved):
The declaration and implementation are still broken up and hard to read. If it is too much to change for this PR, we can go ahead and move with the merge.

ckendrick:
Can we pause merging this temporarily? I don't think this will cause a lot of rebasing issues for other PRs, but I would like to check real quick before this is merged in.

ckendrick (approved):
This should be fine to merge now.

Issue #521, opened 2023-03-31
Add ability to skip performance timers


ckendrick:
There should be compile time option added to disable `tools::timer.start()` and `tools::timer.stop()` with a noop.
This can probably be done similarly to how `expect` is implemented:
https://github.com/project-asgard/asgard/blob/01016aa01129e560a14093b557e2b7f2c4c17ab0/src/tools.hpp#L15-L19

mkstoyanov:
I agree but this should be done with another option and not the `NDEBUG`, i.e., `-DASGARD_USE_TIMER=ON/OFF` where the default is `ON`.

The time does not take much and sometimes we want it in `Release` mode too, but not always.

This should be easy with just replacing the existing methods with no-op and returning empty strings.

PR #644, opened 2023-09-29
No cost timer

mkstoyanov:
This relates to #521 

Switching many of the calls to timer start/stop with raii style of structs (start is called by the constructor and stop is called by the destructor). The CMake option (comes later) will replace the struct with a no-opt one.

### DIFF
--- a/src/time_advance.cpp
+++ b/src/time_advance.cpp
@@ -241,10 +241,10 @@ explicit_advance(PDE<P> const &pde, matrix_list<P> &operator_matrices,
 
   // -- RK step 2
   {
-    tools::time_event performance("kronmult");
+    tools::time_event performance(
+        "kronmult", operator_matrices[matrix_entry::regular].flops());
     operator_matrices[matrix_entry::regular].apply(1.0, x.data(), 0.0,
                                                    fx.data());
-    performance.flops = operator_matrices[matrix_entry::regular].flops();
   }
   reduce_results(fx, reduced_fx, plan, get_rank());
 
--- a/src/time_advance.cpp
+++ b/src/time_advance.cpp
@@ -566,7 +566,7 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
   auto do_poisson_update = [&](fk::vector<P, mem_type::owner, imex_resrc> const
                                    &f_in) {
     {
-      tools::time_event performance("poisson_update");
+      tools::time_event pupdate_("poisson_update");
       // Get 0th moment
       fk::vector<P, mem_type::owner, imex_resrc> mom0(dense_size);
       fm::sparse_gemv(pde.moments[0].get_moment_matrix_dev(), f_in, mom0);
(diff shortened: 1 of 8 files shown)


src/tools.hpp
```diff
@@ -106,4 +106,28 @@ class simple_timer
 
 extern simple_timer timer;
 
+/*!
+ * Allows for RAII style of timing for blocks of code.
+ * The constructor will initiate the timer for the given event,
+ * the destructor will stop the timer.
+ */
+struct time_event
```
quantumsteve, line 114: Would you add a unit test in `tools_tests.cpp`?

src/tools.hpp
```diff
@@ -106,4 +106,28 @@ class simple_timer
...
+{
+  //! \brief Constructor, start timing.
+  time_event(std::string event_name)
```
quantumsteve:
```suggestion
  time_event(std::string const &event_name)
```

src/tools.hpp
```diff
@@ -106,4 +106,28 @@ class simple_timer
...
+      : event_name_(timer.start(event_name)), flops(-1)
+  {}
+  //! \brief Constructor, start timing for flop count.
+  time_event(std::string event_name, int op_flops)
```
quantumsteve:
```suggestion
  time_event(std::string const &event_name, int op_flops)
```

src/tools.hpp
mkstoyanov: The `event_name` is usually short string and I want to use it inline and move out of it.

Isn't passing by non-const value better for that purpose?

src/time_advance.cpp
```diff
@@ -212,11 +212,14 @@ explicit_advance(PDE<P> const &pde, matrix_list<P> &operator_matrices,
...
-  auto const apply_id = tools::timer.start("kronmult");
   fk::vector<P> fx(row_size);
-  operator_matrices[matrix_entry::regular].apply(1.0, x.data(), 0.0, fx.data());
 
-  tools::timer.stop(apply_id, operator_matrices[matrix_entry::regular].flops());
+  {
+    tools::time_event performance("kronmult");
+    operator_matrices[matrix_entry::regular].apply(1.0, x.data(), 0.0,
+                                                   fx.data());
+    performance.flops = operator_matrices[matrix_entry::regular].flops();
```
quantumsteve, line 221: Copying flops like this looks error-prone. (don't need to change in this PR).

src/tools.hpp
quantumsteve: Assuming SSO and time_event's implementation in the header, I think passing by value or reference will give very similar runtime. In the worst case scenario (size() greater than fits inside `std::string`) passing by const reference would be preferred.

Another alternative would be `std::string_view`, but that would require an overload to `simple_timer::start(...)` and changing the comparator in `id_to_start_` to allow heterogeneous comparison lookup.

src/time_advance.cpp
mkstoyanov: You're right, I missed that.

src/tools.hpp
mkstoyanov: Just did.

quantumsteve (dismissed):

src/tools.hpp
mkstoyanov: You're right, the implementation of the `simple_timer` makes the point mute.

I'm not changing `simple_timer` in this PR.

src/tools_tests.cpp
```diff
@@ -25,20 +25,37 @@ double shuffle_random(int const num_items)
...
   int const items_to_gen       = 100000;
   int const iterations         = 10;
   std::string const identifier = "waste_time";
   for (int i = 0; i < iterations; ++i)
   {
-    timer.start(identifier);
-    double const val = shuffle_random(items_to_gen);
-    timer.stop(identifier);
-    expect(val > 0.0); // to avoid comp. warnings
+    if (std::is_same_v<TestType, simple_timer_tag>)
```
quantumsteve:
```suggestion
    if constexpr (std::is_same_v<TestType, simple_timer_tag>)
```

src/tools.hpp
```diff
@@ -106,4 +106,28 @@ class simple_timer
...
+ */
+struct time_event
+{
+  //! \brief Constructor, start timing.
+  time_event(std::string const &event_name)
+      : event_name_(timer.start(event_name)), flops(-1)
+  {}
+  //! \brief Constructor, start timing for flop count.
+  time_event(std::string const &event_name, int op_flops)
+      : event_name_(timer.start(event_name)), flops(op_flops)
```
quantumsteve, line 122: Is the cast from `int` to `double` intentional?

quantumsteve (changes requested):
Just a comment about making if statement (explicitly) constexpr and asking about a cast between types.

src/tools.hpp
mkstoyanov: No, it wasn't. Good catch.

quantumsteve (approved):

PR #649, opened 2023-11-28
fix issue with assertions in axpy

mkstoyanov:
Small problem I cam across.


quantumsteve (approved):
When `alpha == 0.` is this just a copy?

mkstoyanov:
> When `alpha == 0.` is this just a copy?

No, it should be a no-op, the operation is `y += alpha * x`

PR #651, opened 2023-12-11
Global kronecker

mkstoyanov:
Lots of changes:
* global kronmult with cpu implementation
    * still working on better way to handle the edge connection
* support for cuda (mixed mode, global kron is on the cpu)
* moving to better memory management with gpu::vector
* updated the line length of the clang format
    * disabled formatting for some files to avoid redundant changes
* global kron is enabled with CMake command KRON_MODE_GLOBAL
* global kron also skips the identity matrices
* new indexset data structure


src/asgard_kronmult_matrix.cpp
```diff
@@ -936,7 +1003,575 @@ compute_mem_usage(PDE<P> const &pde,
...
+template<typename precision>
+bool check_identity_term(PDE<precision> const &pde, int term_id, int dim)
+{
+  auto const &tterm = pde.get_terms()[term_id][dim];
+  if (tterm.get_partial_terms().size() == 1)
+    return (tterm.get_partial_terms()[0].g_func == nullptr and
+            tterm.get_partial_terms()[0].coeff_type == coefficient_type::mass);
+  else
+    return false;
+}
```
mkstoyanov, line 1012: @stefan-schnake Can you take a look at this, I think this is the logic to identify the identity terms.

stefan-schnake: I think you also need the volume jacobian to the `nullptr` for the pterm to be identity.  @quantumsteve should verify as he implemented this change.  

Also, if all of the pterms are identity, then the resulting term for that dimension is identity.

mkstoyanov: Makes sense, I think I got it now.

.clang-format
```diff
@@ -44,7 +44,7 @@ BreakConstructorInitializersBeforeComma: false
 BreakConstructorInitializers: BeforeColon
 BreakAfterJavaFieldAnnotations: false
 BreakStringLiterals: true
-ColumnLimit:     80
+ColumnLimit:     0
```
quantumsteve, line 47: How do you feel about a column limit of 100 or 120? I fear no limit will break split view on many monitors.

https://llvm.org/docs/CodingStandards.html#source-code-width
https://google.github.io/styleguide/cppguide.html#Line_Length

CMakeLists.txt
```diff
@@ -111,6 +111,19 @@ set_property (CACHE ASGARD_PRECISIONS PROPERTY STRINGS "float" "double" "float\;
 option (ASGARD_RECOMMENDED_DEFAULTS "Enable OpenMP, set some flags, download OpenBLAS if system BLAS is missing." OFF)
 option (ASGARD_BUILD_DOCS "Build the documentation." OFF)
 
+if (ASGARD_USE_MPI)
+  option (KRON_MODE_GLOBAL "Global or local Kronecker products" OFF)
+else()
+  option (KRON_MODE_GLOBAL "Global or local Kronecker products" ON)
+endif()
```
quantumsteve, line 118: Can this be done with a CMake generator expression?

src/asgard_grid_1d.hpp
```diff
@@ -19,25 +19,36 @@ namespace asgard
...
+  static tag_level_edge_only level_edge_only;
   /*!
    *  \brief Constructor, makes the connectivity up to and including the given
    *         max-level.
    */
-  connect_1d(int const max_level)
-      : levels(max_level), cells(1 << levels), pntr(cells + 1, 0),
-        indx(2 * cells)
+  connect_1d(int const max_level, same_level neighbor = level_edge_include)
+      : levels(max_level), rows(1 << levels), pntr(rows + 1, 0),
```
quantumsteve:
```suggestion
      : levels(max_level), rows(fm::two_raised_to(levels)), pntr(rows + 1, 0),
```

src/asgard_grid_1d.hpp
```diff
@@ -147,13 +169,121 @@ class connect_1d
...
-      return col;
-    else if (row == 1)
-      return cells + col;
+    // if we use the hierarchy with all elements connected by volume
+    // the first two elements have lots of connection which slows the search
+    // but the search result is trivial
+    //if (row == 0)
+    //  return col;
+    //else if (row == 1)
+    //  return rows + col;
```
quantumsteve, line 287: should we delete these lines?

src/asgard_grid_1d.hpp
```diff
@@ -176,23 +306,48 @@ class connect_1d
...
   int max_loaded_level() const { return levels; }
 
+  void dump() const // for debugging
+  {
+    std::cerr << "dumping connectivity to std::cerr\n";
+    for(int r=0; r < num_rows(); r++)
+    {
+      for(int j=row_begin(r); j<row_end(r); j++)
+        std::cerr << indx[j] << "  ";
+      std::cerr << "\n";
```
quantumsteve:
```suggestion
      std::cerr << '\n';
```

quantumsteve: couple other nearby places where a single character can replace a string literal.

src/asgard_indexset.hpp
```diff
@@ -0,0 +1,345 @@
...
+
+  //! \brief Get the j-th global offset
+  int map(int dimension, int j) const { return map_[dimension][j]; }
+  //! \brief Get the 1d index of the j-th entry
+  int operator()(indexset const &iset, int dimension, int j) const { return iset[map_[dimension][j]][dimension]; }
+  //! \brief Get the 1d index of the j-th entry
+  int operator()(vector2d<int> const &list, int dimension, int j) const { return list[map_[dimension][j]][dimension]; }
+
+private:
+  std::vector<std::vector<int>> map_;
```
quantumsteve: What's a reasonable synonym that doesn't remind us of `std::map`?

src/asgard_kronmult_matrix.cpp
```diff
@@ -161,8 +160,67 @@ make_kronmult_dense(PDE<precision> const &pde,
...
+                                 int const *const col)
+{
+  // different levels, check if the points are connected by volume
+  auto check_diff_volumes = [](int l1, int p1, int l2, int p2)
+      -> bool {
+    if (l1 < l2)
+    {
+      while (l1 < l2)
+      {
+        l2--;
```
quantumsteve, line 174: I was taught to prefer the pre-increment operator (realizing for integers an optimizing compiler will generate the same output)

src/asgard_kronmult_matrix.cpp
```diff
@@ -161,8 +160,67 @@ make_kronmult_dense(PDE<precision> const &pde,
...
+  int edge_conn = 0;
+  for (int j = 0; j < num_dimensions; j++)
+  {
+    if (row[j] == col[j]) // same level, consider only edge connections
+    {
+      if (row[num_dimensions + j] != col[num_dimensions + j])
+      {
+        if ((row[num_dimensions + j] == 0 and col[num_dimensions + j] == ((1 << (col[j] - 1)) - 1)) or
+            (col[num_dimensions + j] == 0 and row[num_dimensions + j] == ((1 << (row[j] - 1)) - 1)))
+          edge_conn += 1; // periodic boundary
```
```suggestion
          ++edge_conn; // periodic boundary
```

src/asgard_kronmult_matrix.cpp
```diff
@@ -161,8 +160,67 @@ make_kronmult_dense(PDE<precision> const &pde,
...
+        else
+        {
+          if (std::abs(row[num_dimensions + j] - col[num_dimensions + j]) == 1)
+            edge_conn += 1; // adjacent elements
```
```suggestion
            ++edge_conn; // adjacent elements
```

src/asgard_kronmult_matrix.cpp
```diff
@@ -936,7 +1001,573 @@ compute_mem_usage(PDE<P> const &pde,
...
+    // and then to record the non-zeros in the indx structures
+    for (int stage = 0; stage < 2; stage++)
+    {
+      if (stage == 1)
+      {
+        int64_t num_nz = 0;
+        for (auto n : nz_count)
+          num_nz += n;
+        indx  = std::vector<int>(num_nz);
+        ivals = std::vector<int>(2 * num_nz);
```
```suggestion
        indx.resize(num_nz);
        ivals.resize(2 * num_nz);
```

CMakeLists.txt
mkstoyanov: No clue. This is a stop-gap thing right now, we want to have both global kron and MPI in the future.

That's why the variable is names without prefix.

src/asgard_kronmult_matrix.cpp
mkstoyanov: Why? What's the benefit?

src/asgard_grid_1d.hpp
mkstoyanov: There is a potential here, I updated the comment to remind me to look into it later.

Maybe not relevant, but not sure.

src/asgard_kronmult_matrix.cpp
```diff
@@ -936,7 +1001,573 @@ compute_mem_usage(PDE<P> const &pde,
...
+                                                  ? program_options.max_level
+                                                  : pde.max_level);
+
+  int const block_size = kron_size * kron_size;
+
+  int const num_terms = static_cast<int>(used_terms.size());
+  if (num_terms == 0)
+    return;
+
+  for (auto t : used_terms)
```
```suggestion
  for (int t : used_terms)
```

src/asgard_vector.hpp
```diff
@@ -1112,4 +1112,121 @@ fk::vector<P, mem, resrc>::vector(fk::matrix<P, omem, resrc> const &source, int,
...
+namespace gpu
+{
+/*!
+ * \brief Simple container for GPU data, interoperable with std::vector
+ *
+ * This simple container allows for RAII memory management,
+ * resizing (without relocating the data) and easy copy from/to std::vector
+ */
+template<typename T>
+class vector
```
quantumsteve, line 1127: Should this be in a separate file?

src/asgard_resources_tests.cpp
```diff
@@ -55,3 +55,73 @@ TEMPLATE_TEST_CASE("device functions", "[resources]", test_precs, int)
...
+  std::vector<typename vec_type_2::value_type> y = v2;
+  if (x.size() != y.size()) // something happened during copy
+    return false;
+  for (size_t i = 0; i < x.size(); i++)
+    if (x[i] != y[i]) // this checks data copies, so it's OK for floating point numbers
+      return false;
+  return true;
+}
+
+TEMPLATE_TEST_CASE("gpu::vector", "[gpu::vector]", test_precs, int)
```
quantumsteve: should the gpu::vector tests also be in a separate file?

.clang-format
mkstoyanov: I don't have a problem with the idea of keeping lines short, but the hard limit causes the code to jumble up even if one character steps out of line. It's the mindless enforcement that is more frustrating then anything else.

The split-view argument works both ways, short lines mean more lines and an algorithm expressed in more lines is both harder to follow up-and-down and thus harder to read. 

The first version of the kron kernels that I wrote used lines that would break any split-view, but I could read the code and follow the common patterns. After I pushed it through clang-format, I can no longer read. I won't use ultra long lines as a matter of practice, but I want the option if it will make the code easier to read.

I have two monitors and no line length will break that split-view. Currently the success of the project depends heavily on my coding efficiency. How much should I hinter myself and my current progress in favor of a hypothetical future contributor that will have limited hardware. I myself travel often enough and have to code on a much smaller laptop screen, yet between intelligent warp around and alt-tab, wide lines don't bother me at all.

Here is the opinion of someone else:
>People with restrictive hardware shouldn't make it more inconvenient for people who have better resources.

If you want split-view, get a second monitor.

https://lkml.iu.edu/hypermail/linux/kernel/2005.3/08168.html
https://www.theregister.com/2020/06/01/linux_5_7/

src/asgard_indexset.hpp
mkstoyanov: Replaced it with `iorder_` as in index-order

src/asgard_vector.hpp
mkstoyanov: As we are phasing out `fk::vector` this will be the `asgard_vector`

This is a good place.

src/asgard_resources_tests.cpp
mkstoyanov: It's the main gpu "resource" for data management. It fits just fine with "resource".

src/asgard_resources_tests.cpp
```diff
@@ -55,3 +55,73 @@ TEMPLATE_TEST_CASE("device functions", "[resources]", test_precs, int)
...
+{
+  if (static_cast<int64_t>(v1.size()) != static_cast<int64_t>(v2.size()))
+    return false;
+  static_assert(std::is_same_v<typename vec_type_1::value_type, typename vec_type_2::value_type>);
+  std::vector<typename vec_type_1::value_type> x = v1;
+  std::vector<typename vec_type_2::value_type> y = v2;
+  if (x.size() != y.size()) // something happened during copy
+    return false;
+  for (size_t i = 0; i < x.size(); i++)
+    if (x[i] != y[i]) // this checks data copies, so it's OK for floating point numbers
```
quantumsteve: `std::memcmp`?

mkstoyanov: Nah!

The code will have to be:
```
return (std::memcmp(x.data(), y.data(), x.size() * sizeof(typename vec_type_1::value_type)) == 0);
```
That's way too messy to read.

src/device/asgard_kronmult.hpp
```diff
@@ -128,6 +130,152 @@ void gpu_sparse(int const dimensions, int const n, int const output_size,
...
+      for (int d = 0; d < num_dimensions; d++)
+      {
+        fill[perm][d] = (direction[perm][d] < 0) ? matrix_fill::upper : ((direction[perm][d] > 0) ? matrix_fill::lower : matrix_fill::both);
+
+        direction[perm][d] = std::abs(direction[perm][d]);
+      }
+    }
+  }
+  //! \brief Convert the fill to a string (for debugging).
+  const char *fill_name(int perm, int stage) const
```
quantumsteve:
```suggestion
  std::string_view fill_name(int perm, int stage) const
```

src/asgard_kronmult_matrix.cpp
quantumsteve: > Preincrement is more efficient than postincrement, because for postincrement the object must increment itself and then return a temporary containing its old value. Note that this is true even for builtins like int!
http://www.gotw.ca/gotw/002.htm

Looks like it may make a (tiny?) difference with iterators.
https://bastian.rieck.me/blog/posts/2015/postincrement_vs_preincrement/

mkstoyanov: It can make a difference for a more complex object, maybe even an iterator, but I'm using mostly ints. There's no compiler that would miss to optimize `++i` vs `i++` when dealing with `int, size_t, int64_t`.

Good point though, I didn't know the general difference. I'll try to use pre-fix for iterators.

src/asgard_resources_tests.cpp
mkstoyanov: I'll move this to the tensor tests.

mkstoyanov: Done!

mkstoyanov:
@quantumsteve do you have more suggestions here or are we ready to merge

quantumsteve (approved):
Saw @stefan-schnake put this code through its paces before Wednesday's meeting.

Let's :shipit:

PR #657, opened 2023-12-15
Cusparse mats

mkstoyanov:
Another code dump:
* global sparse operations now use the cusparse library
* handles the logic of doing all work on the GPU without moving back to the CPU side for the global kronecker
* did not use the existing sparse method, wanted to avoid allocations and keep data-structures persistent
* removed the whole pinned memory thing

### DIFF
--- a/src/asgard_kronmult_matrix.cpp
+++ b/src/asgard_kronmult_matrix.cpp
@@ -1181,19 +1181,17 @@ make_global_kron_matrix(PDE<precision> const &pde,
     std::vector<std::vector<int>> tindx  = std::move(global_indx);
     std::vector<std::vector<int>> tivals = std::move(global_ivals);
 
-    std::vector<std::vector<int>> global_pntr(3 * num_dimensions - 2);
-    std::vector<std::vector<int>> global_indx(3 * num_dimensions - 2);
-    std::vector<std::vector<int>> global_ivals(3 * num_dimensions - 2);
+    std::vector<std::vector<int>> global_pntr(3 * num_dimensions);
+    std::vector<std::vector<int>> global_indx(3 * num_dimensions);
+    std::vector<std::vector<int>> global_ivals(3 * num_dimensions);
 
-    global_pntr[0]  = std::move(tpntr[0]); // do not split dimension 0
-    global_indx[0]  = std::move(tindx[0]);
-    global_ivals[0] = std::move(tivals[0]);
-
-    for (int d = 0; d < num_dimensions - 1; d++)
+    for (int d = 0; d < num_dimensions; d++)
     {
-      global_pntr[3 * (d - 1)]  = std::move(tpntr[d + 1]); // do not split dimension 0
-      global_indx[3 * (d - 1)]  = std::move(tindx[d + 1]);
-      global_ivals[3 * (d - 1)] = std::move(tivals[d + 1]);
+      global_pntr[3 * d]  = std::move(tpntr[d]); // copy the full pattern
+      global_indx[3 * d]  = std::move(tindx[d]);
+      global_ivals[3 * d] = std::move(tivals[d]);
+      if (d == 0) // never split pattern 0, the matrix is always used full
+        continue;
       split_pattern(global_pntr[3 * d], global_indx[3 * d], global_diag[d], global_ivals[3 * d],
                     global_pntr[3 * d + 1], global_indx[3 * d + 1], global_ivals[3 * d + 1],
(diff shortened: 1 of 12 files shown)


mkstoyanov:
@quantumsteve Let's try to get this done before Wednesday, this is the same algorithm as before with the addition of cuSparse for the global portion of the work.

The next PR is much shorter and has lots of cleanup. Also, it removes the need for the old kronmult code.

src/asgard_kronmult_matrix.cpp
```diff
@@ -1360,9 +1449,15 @@ void set_specific_mode(PDE<precision> const &pde,
...
   int64_t gflops = 0; // flops for global component
+#ifdef ASGARD_USE_CUDA
+  for (auto t : used_terms)
+    for (int d = 0; d < num_dimensions; d++)
+      gflops += mat.gvals_[3 * t * num_dimensions + d].size();
+#else
   for (auto t : used_terms)
     for (int d = 0; d < num_dimensions; d++)
       gflops += mat.gvals_[t * num_dimensions + d].size();
+#endif
```
quantumsteve: I think you only need the `#ifdef` for lines 1455 and 1459.

mkstoyanov: Updated!

src/asgard_kronmult_matrix.hpp
```diff
@@ -919,16 +954,48 @@ class global_kron_matrix
...
+    if (static_cast<int64_t>(w.size()) < min_size)
+    {
+      w.resize(min_size);
+      // x must always be padded by zeros
+      if constexpr (wid == workspace::pad_x)
+      {
+#ifdef ASGARD_USE_CUDA
+        // zero out the a-vector (maybe run a kernel here)
+        std::vector<precision> tmp(min_size, precision{0});
+        fk::copy_to_device(w.data(), tmp.data(), min_size);
```
quantumsteve, line 984: Let's create an issue about this missing functionality.

src/device/asgard_cusparse.hpp
```diff
@@ -0,0 +1,247 @@
...
+  sparse_matrix(sparse_matrix<T> &&other)
+    : desc_(std::exchange(other.desc_, nullptr)),
+      x_(std::exchange(other.x_, nullptr)),
+      y_(std::exchange(other.y_, nullptr)),
+      scale_factors_(std::move(other.scale_factors_))
+  {}
+  //! \brief Container is movable, via constructor or assignment
+  sparse_matrix<T> &operator =(sparse_matrix<T> &&other)
+  {
+    sparse_matrix<T> tmp(std::move(other));
```
quantumsteve, line 177: Why do we need to move into a temporary, then swap member variables?

src/device/asgard_glkronmult_gpu.cpp
```diff
@@ -0,0 +1,101 @@
...
+        if (gindx_[pid].empty())
+          gindx_[pid] = gindx[pid];
+        if (gvals_[vid].empty())
+          gvals_[vid] = gvals[vid];
+
+        mats_.emplace_back(
+            gpu::sparse_matrix<precision>(num_rows, num_rows, gindx_[pid].size(),
+                                          gpntr_[pid].data(), gindx_[pid].data(),
+                                          gvals_[vid].data())
+        );
```
quantumsteve:
```suggestion
        mats_.emplace_back(num_rows, num_rows, gindx_[pid].size(),
                                          gpntr_[pid].data(), gindx_[pid].data(),
                                          gvals_[vid].data());
```

src/device/asgard_cusparse.hpp
mkstoyanov: It's something that I saw at a cppcon lecture. The main question is about what would happen if "this" and "other" are actually the same and you're moving into yourself. You have to free the resources in `this` but not free anything in `other`, also you don't want to repeat the code for `free` if you already have it in the destructor.

Normally, you don't want to move onto yourself, but through aliases and references it could happen. This is the safest way to do it.
* if `other` and `this` are the same, we can always safely move the data into the temporary, then swap it back.
* if the two are different, then all of the resources in `this` will be freed by the destructor of the temporary and we don't have to duplicate the code.

src/asgard_kronmult_matrix.hpp
mkstoyanov: It's a 4 line kernel, I added it to the branch for the next PR.

quantumsteve (approved):
LGTM

PR #662, opened 2024-01-25
Add parallel gmres implementation

quantumsteve:

## Proposed changes

Haven't tried with GPUs. I think this will be easiest with CUDA-aware MPI.

Kept existing serial implementation to avoid additional copies. We may want to revisit this soon.

- [ ] Bugfix
- [x] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

Only tested on continuity_1 CPU and 4 processes. Need to try more PDEs with different numbers of processes.

## Checklist

### DIFF
--- a/src/distribution.cpp
+++ b/src/distribution.cpp
@@ -315,7 +315,7 @@ void reduce_results(fk::vector<P, src_mem, resrc> const &source,
       MPI_Comm_split(global_communicator, my_row, my_col, &row_communicator);
   expect(success == 0);
 
-  MPI_Datatype const mpi_type =
+  MPI_Datatype constexpr mpi_type =
       std::is_same_v<P, double> ? MPI_DOUBLE : MPI_FLOAT;
   success = MPI_Allreduce((void *)source.data(), (void *)dest.data(),
                           source.size(), mpi_type, MPI_SUM, row_communicator);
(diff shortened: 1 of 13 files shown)


src/solver.cpp
```diff
@@ -282,7 +315,155 @@ simple_gmres(matrix_abstraction mat, fk::vector<P, mem_type::view, resrc> x,
...
+    tolerance = std::is_same_v<float, P> ? 1e-6 : 1e-12;
+  expect(tolerance >= std::numeric_limits<P>::epsilon());
+
+  int const n = b.size();
+  expect(n == x.size());
+
+  if (restart == parser::NO_USER_VALUE)
+    restart = default_gmres_restarts<P>(n);
+  expect(restart > 0); // checked in program_options
+  if (restart > n)
```
quantumsteve: @mkstoyanov  With MPI `n` now varies by the number of processes. How can I easily determine the full matrix size without communication between processes?

mkstoyanov: The method should be guided by the global `n` and not the local one.

When you have the `adapt::distributed_grid`, I think every MPI rank "knows" the basic dimensions of the other ranks, so instead of reading the local one, you just find the global. I may have to dig for the specifics, but it should be in there.

src/distribution.cpp
```diff
@@ -330,6 +330,49 @@ void reduce_results(fk::vector<P, src_mem, resrc> const &source,
...
+  int const num_cols = get_num_subgrid_cols(plan.size());
+
+  int const my_row = my_rank / num_cols;
+  int const my_col = my_rank % num_cols;
+
+  MPI_Comm row_communicator;
+  MPI_Comm const global_communicator = distro_handle.get_global_comm();
+
+  auto success =
+      MPI_Comm_split(global_communicator, my_row, my_col, &row_communicator);
```
mkstoyanov, line 358: The `MPI_Comm_split()` call causes global communication, see the algorithm description here:

https://www.mpich.org/static/docs/v3.1.3/www3/MPI_Comm_split.html

It's OK to use (and for prototype purposes), but eventually this step should be done only once. You can have a reducer class, where you initialize an object once and you keep the communicator, then you just call `MPI_Allreduce()` for every iteration.

mkstoyanov:
Stale and out of date with respect to the new API

PR #668, opened 2024-03-12
Get building again with ASGARD_USE_MATLAB=ON

quantumsteve:

## Proposed changes

Changes required to successfully build with ASGARD_USE_MATLAB=ON

- [x] Bugfix
- [ ] New feature
- [ ] Build related changes

- [ ] Yes
- [x] No

my laptop

## Checklist

### DIFF
--- a/src/matlab_plot.cpp
+++ b/src/matlab_plot.cpp
@@ -218,10 +218,6 @@ fk::vector<T> matlab_plot::generate_nodes(int const degree, int const level,
 
   for (int i = 0; i < n; i++)
   {
-    auto p_val = legendre(roots, degree, legendre_normalization::lin);
-
-    p_val[0] = p_val[0] * sqrt(1.0 / h);
-
     std::vector<T> xi(dof);
     for (std::size_t j = 0; j < dof; j++)
     {
--- a/src/matlab_plot.cpp
+++ b/src/matlab_plot.cpp
@@ -209,7 +209,7 @@ fk::vector<T> matlab_plot::generate_nodes(int const degree, int const level,
 
   // TODO: fully implement the output_grid options from matlab (this is just
   // the 'else' case)
-  T const roots =
+  auto const roots =
       legendre_weights<T>(degree, -1.0, 1.0, quadrature_mode::use_degree)[0];
 
   unsigned int const dof = roots.size();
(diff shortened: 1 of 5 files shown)


src/matlab_plot.cpp
```diff
@@ -209,7 +209,7 @@ fk::vector<T> matlab_plot::generate_nodes(int const degree, int const level,
 
   // TODO: fully implement the output_grid options from matlab (this is just
   // the 'else' case)
-  T const roots =
+  auto const roots =
```
mkstoyanov: I don't like the overuse of `auto`, if a type is simple, e.g., `int` `double` `T`, then we should just use the type.

`auto` is useful when the output depends on template parameters or is something potentially messy (e.g., some crazy iterator).

src/time_advance.cpp
```diff
@@ -571,6 +571,9 @@ imex_advance(PDE<P> &pde, matrix_list<P> &operator_matrices,
 
   auto do_poisson_update = [&](fk::vector<P, mem_type::owner, imex_resrc> const
                                    &f_in) {
+    fk::vector<P> poisson_source(quad_dense_size);
+    fk::vector<P> phi(quad_dense_size);
+    fk::vector<P> poisson_E(quad_dense_size);
```
mkstoyanov, line 576: I guess `tools::time_event` wasn't such a good idea after all ...

mkstoyanov (changes requested):
Just change the `auto`, the rest of the changes contain lots of **ouch**, good you caught those.

src/matlab_plot.cpp
mkstoyanov: If the type is `fk::vector<T>` then how did it work before with just `T`?

quantumsteve: I tried searching the history yesterday but didn't figure that out. Line 215 calls `roots.size()` which wouldn't work for T = float or T = double.

src/time_advance.cpp
quantumsteve: It shows the potential benefits of a CI runner with MATLAB or removing the MATLAB code entirely.

src/matlab_plot.cpp
mkstoyanov: Very strange ... `legendre_weights` does return two `fk::vectors` so the `fk::vector` should be correct.

src/time_advance.cpp
mkstoyanov: We are going  towards no-MATLAB but we have to make sure we don't have a major regression in plotting if we were to remove the code. Should check with Stefan today.

Either way the JOSS submission can include MATLAB for now.

mkstoyanov (approved):

PR #678, opened 2024-04-18
fix the init of workspaces

### LINKED ISSUE #670: Divergence in Kronmults
**Describe the bug**

There is a discrepancy between the global and local kronmults when adaptivity is enabled.  This is causing instabilities in the collision operator solve.

**To Reproduce**
1. Current git hash
2. ```OMP_NUM_THREADS=8 ./asgard -p two_stream -d 3 -l "7 7" -x -t 2.3438e-3 -n 5 -m 7 --wave_freq 1 --adapt --thresh 1e-6 --kron-mode sparse ```

**Expected behavior**
Results should be the same within machine epsilon when `USE_GLOBAL_KRON=ON/OFF`.  Diverges at the 4th timestep.

**System:**
 - andremarie

**Additional context**
Issue only arrises when adaptivity is enabled and could be due to padding.  If `thresh=1e-3`, then the results match for 100 timesteps.  In this case, only four elements are added for padding.

mkstoyanov:
Avoiding excessive allocations, the workspace buffer does not shrink if the number of degrees of freedom is reduced. On the other hand, the workspace buffer for `x` needs to be padded with zeros for the padding cells that complete the hierarchy. If the buffer is not reset to zero every time the problem size-changes, then at some point we will be padding with the old entries for `x` from the previous time-step or previous refinement iteration, i.e., padding with non-zeros. Most of the time, using non-zeros for the padding will change the time-step significantly, which will trigger a refinement cycle that rejects the time-step and tries to repeat it on a different grid. If the grid is denser, the workspace vector will be resized and zeroed out, leaving no clearly observable problem. However, a small but non-zero entry can introduce a small enough change to fly below the refinement threshold, yet large enough to break the physics over several iterations.

TL;DR: Sometimes we were padding with non-zeros and that breaks the physics, but only over multiple iterations/timesteps.

closes #670

### DIFF
--- a/src/asgard_kronmult_matrix.hpp
+++ b/src/asgard_kronmult_matrix.hpp
@@ -954,12 +954,10 @@ protected:
                   "no buffer associated with the last entry (num_spaces)");
     default_vector<precision> &w = (*work_)[static_cast<int>(wid)];
     if (static_cast<int64_t>(w.size()) < min_size)
-    {
       w.resize(min_size);
-      // x must always be padded by zeros
-      if constexpr (wid == workspace::pad_x)
-        kronmult::set_buffer_to_zero(w);
-    }
+    // x must always be padded by zeros
+    if constexpr (wid == workspace::pad_x)
+      kronmult::set_buffer_to_zero(w);
   }
 
 private:


quantumsteve:
I successfully ran our `vlasov` example with `KRON_MODE_GLOBAL=ON`

mkstoyanov:
> I successfully ran our `vlasov` example with `KRON_MODE_GLOBAL=ON`

Sounds great! Did you do adaptivity? The issue manifests only during adaptive refinement.

quantumsteve:
> > I successfully ran our `vlasov` example with `KRON_MODE_GLOBAL=ON`
> 
> Sounds great! Did you do adaptivity? The issue manifests only during adaptive refinement.

./asgard -p vlasov -d 3 -l "6 5" -x -t 8e-4 -n 125 -m 6 --kron-mode dense --inner_it 50 --tol 1e-6 --adapt --max_adapt_levels "6 5" --thresh 1e-4

stefan-schnake (approved):
For reference before this PR the vlasov PDE would stagnate in the GMRES run on the global kronmult.  Now it runs.  

First I ran
```
./asgard -p vlasov -t 1.9531e-4 -n 512 -x -d 3 -m 8 --adapt --thresh 1e-4 --wave_freq 32 --tol 1e-6 --inner_it 50
```
on both kronmults.  The error started at `2.11e-07` @ 32 timestep and finished at `6.09e-06` with a max error in time at `1.39e-05`.  There errors are well below the tolerance.  The number of active elements were the same for outputted timesteps between each run, but the order was not.

Next I bumped down the threshold is `1e-8`, the GMRES tolerance is `1e-10`.  and capped the adaptivity at "8 6".  Error is `9.11e-10` @ timestep 32 and `9.77e-09` at the end.  Max error in time is `1.02e-08`.  The number of active elements started to differ after 100 or so timesteps.  I think this has more to do the GMRES than the algorithms. 

Based on the errors I believe the global kroneckor is as accurate as the  local one.

mkstoyanov:
Thanks to both of you for checking this out.

PR #683, opened 2024-06-10
handle case when non-identity and no flux term

mkstoyanov:
@stefan-schnake

The flux term should always be at position `0` in the `active_terms` as this is the one that doesn't get split into upper/lower part. Active term `0` does not require completeness of the hierarchy, the rest of directions can be given in any order.

This fixes the case when we have a term that is a mass term (no flux) but is not identity. We should also double-check the logic in `get_flux_direction()`.

### DIFF
--- a/src/asgard_kronmult_matrix.cpp
+++ b/src/asgard_kronmult_matrix.cpp
@@ -1327,7 +1327,7 @@ make_global_kron_matrix(PDE<precision> const &pde,
     if (num_active > 1)
     {
       int const flux_dir = get_flux_direction(pde, t);
-      if (flux_dir != active_dirs[0]) // make the flux direction first
+      if (flux_dir > -1 and flux_dir != active_dirs[0]) // make the flux direction first
         std::swap(active_dirs[0], active_dirs[flux_dir]);
     }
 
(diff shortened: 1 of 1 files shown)


stefan-schnake (commented):
If were reviewing this now:
`pt.coeff_type == coefficient_type::penalty` needs to be added to L1199.

stefan-schnake (commented):
This looks fine.  The bug would show itself if we had a term of all masses, but we don't have a PDE that does that (yet).  Maybe we should have a test for $u_t + u = \Delta u$.

mkstoyanov:
@stefan-schnake How does it look now?

mkstoyanov:
> This looks fine. The bug would show itself if we had a term of all masses, but we don't have a PDE that does that (yet). Maybe we should have a test for ut+u=Δu.

True enough, we are missing tests for lots of corner cases. Part of the problem is the clumsy way of adding new PDEs and the need to integrate them into the library as opposed to just having them externally defined.

stefan-schnake:
BGK includes a mass term, so we'll just wait for that.

stefan-schnake (approved):

PR #686, opened 2024-06-17
Fix pde initialization

mkstoyanov:
Allows for instantiation of an empty PDE object with no dimensions and terms. The PDE can later be re-initialized. This allows for delayed initialization.

### DIFF
--- a/src/pde/pde_base.hpp
+++ b/src/pde/pde_base.hpp
@@ -567,22 +567,22 @@ public:
             moments_in, do_collision_operator_in)
   {}
   PDE(parser const &cli_input, int const num_dims_in, int const num_sources_in,
-      int const max_num_terms, std::vector<dimension<P>> const dimensions,
-      term_set<P> const terms, std::vector<source<P>> const sources_in,
-      std::vector<md_func_type<P>> const exact_vector_funcs_in,
-      scalar_func<P> const exact_time_in, dt_func<P> const get_dt,
-      bool const do_poisson_solve_in          = false,
-      bool const has_analytic_soln_in         = false,
-      std::vector<moment<P>> const moments_in = {},
-      bool const do_collision_operator_in     = true)
+      int const max_num_terms, std::vector<dimension<P>> dimensions,
+      term_set<P> terms, std::vector<source<P>> sources_in,
+      std::vector<md_func_type<P>> exact_vector_funcs_in,
+      scalar_func<P> exact_time_in, dt_func<P> get_dt,
+      bool const do_poisson_solve_in      = false,
+      bool const has_analytic_soln_in     = false,
+      std::vector<moment<P>> moments_in   = {},
+      bool const do_collision_operator_in = true)
   {
     initialize(cli_input, num_dims_in, num_sources_in,
-      max_num_terms, dimensions, terms, sources_in,
-      exact_vector_funcs_in,
-      exact_time_in, get_dt,
+      max_num_terms, std::move(dimensions), std::move(terms), std::move(sources_in),
+      std::move(exact_vector_funcs_in),
+      std::move(exact_time_in), std::move(get_dt),
       do_poisson_solve_in,
       has_analytic_soln_in,
-      moments_in,
+      std::move(moments_in),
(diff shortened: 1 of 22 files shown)


src/pde/pde_base.hpp
```diff
@@ -525,16 +575,53 @@ class PDE
...
-        terms_(terms)
+      // : num_dims(num_dims_in), num_sources(num_sources_in),
+      //   num_terms(get_num_terms(cli_input, max_num_terms)),
+      //   max_level(get_max_level(cli_input, dimensions)), sources(sources_in),
+      //   exact_vector_funcs(exact_vector_funcs_in), moments(moments_in),
+      //   exact_time(check_exact_time(exact_time_in)),
+      //   do_poisson_solve(do_poisson_solve_in),
+      //   do_collision_operator(do_collision_operator_in),
+      //   has_analytic_soln(has_analytic_soln_in), dimensions_(dimensions),
+      //   terms_(terms)
```
quantumsteve: remove commented out code

src/pde/pde_base.hpp
```diff
@@ -525,16 +575,53 @@ class PDE
...
+      bool const has_analytic_soln_in         = false,
+      std::vector<moment<P>> const moments_in = {},
+      bool const do_collision_operator_in     = true)
   {
+    num_dims    = num_dims_in;
+    num_sources = num_sources_in;
+    num_terms   = get_num_terms(cli_input, max_num_terms);
+    max_level   = get_max_level(cli_input, dimensions);
+
+    sources            = std::move(sources_in);
```
quantumsteve: Will one only call initialize once? Pass by value works but can be slower.

https://youtu.be/xnqTKD8uD64?si=BrLtY84hnEDVT20W

mkstoyanov: This one is tricky. I want to avoid copying the funcitonals inside.

Should I pass in by & or && and move from the outside?

I may have to think for a bit on this one.

src/pde_tests.cpp
```diff
@@ -60,20 +60,20 @@ void test_source_vectors(PDE<P> const &pde, std::filesystem::path base_dir,
...
   {
     auto const source_string = filename + "source" + std::to_string(i) + "_";
-    for (auto j = 0; j < pde.num_dims; ++j)
+    for (auto j = 0; j < pde.num_dims(); ++j)
     {
       auto const full_path = base_dir.replace_filename(
           source_string + "dim" + std::to_string(j) + ".dat");
       auto const gold = read_vector_from_txt_file<P>(full_path);
-      auto const fx   = pde.sources[i].source_funcs[j](x, time);
+      auto const fx   = pde.sources()[i].source_funcs()[j](x, time);
```
quantumsteve: Let's change `pde.sources()[i].source_funcs()` to a reference outside of the loop over j.

src/pde/pde_base.hpp
mkstoyanov: Actually, looking at the original initialization code, they were incurring unnecessary copies. First, they always copied going into the constructor as opposed to moving. Then they did not move from the constructor input into the internal variable but copied yet again. It's doubtful that the compiler can get rid of both of those, especially since many variables (e.g., `dimensions`) were passed in as `const` which will explicitly prohibit the move and replace it with a copy.

I agree with the opinion that C++ should have made variables `const` by default but this is how the language is structured and smearing `const` all over the place has unintended consequences such as forcing an extra copy when a move would do the trick or prohibiting relocation and re-initializing of variables. 

I'll try to do it correctly without excessive copy and/or move.

src/pde/pde_base.hpp
mkstoyanov: done

src/pde/pde_base.hpp
mkstoyanov: I think I got it now, you can take a look when you get a chance.

src/pde_tests.cpp
mkstoyanov: Fixed

quantumsteve (approved):

PR #687, opened 2024-06-17
Block global kron cpu

mkstoyanov:
Block-global kronmult will be used for interpolation.


mkstoyanov:
The block-global kronmult is faster than the original one so we will replace the old version but only after we get the GPU kernels. Right now, we need the extra CMake option `-DKRON_MODE_GLOBAL_BLOCK` that only work on the CPU side but will be removed in the future.

src/asgard_kronmult_matrix.cpp
```diff
@@ -1600,6 +1597,158 @@ void global_kron_matrix<precision>::apply(
...
+  int const porder    = pde.get_dimensions()[0].get_degree() - 1;
+  int const pterms    = porder + 1; // poly degrees of freedom
+  int const max_level = (program_options.do_adapt_levels) ? program_options.max_level : pde.max_level();
+
+  int const num_dimensions = pde.num_dims();
+  int const num_terms      = pde.num_terms();
+
+  int64_t block_size = pterms;
+  for (int d = 1; d < num_dimensions; d++)
+    block_size *= pterms;
```
quantumsteve: Would a common integer implementation of `std::pow` make sense?

src/asgard_kronmult_matrix.cpp
```diff
@@ -1600,6 +1597,158 @@ void global_kron_matrix<precision>::apply(
...
+
+  mat.term_groups_[imex_indx] = get_used_terms(pde, program_options, imex);
+
+  std::vector<int> const &used_terms = mat.term_groups_[imex_indx];
+
+  int const n = mat.blockn_;
+
+  int const num_dimensions = pde.num_dims();
+
+  for (int t : used_terms)
```
```suggestion
  for (int const t : used_terms)
```

src/asgard_kronmult_matrix.hpp
```diff
@@ -1276,4 +1278,222 @@ struct matrix_list
...
+  //! \brief Return the number of flops for the current matrix type
+  int64_t flops(matrix_entry etype) const
+  {
+    int i = flag2int(etype);
+    if (flops_[i] == -1)
+    {
+      flops_[i] = kronmult::block_global_count_flops(num_dimensions_, blockn_, block_size_, ilist_, dsort_,
+                                                     perms_, flux_dir_, conn_volumes_, conn_full_,
+                                                     term_groups_[i], *workspace_);
+      switch (etype)
```
quantumsteve, line 1342: Does one only want output if called once? Should the if close before the switch-case statement?

src/asgard_kronmult_matrix.hpp
```diff
@@ -1276,4 +1278,222 @@ struct matrix_list
...
+  //! \brief Convert the imex flag to an index of the arrays.
+  static int flag2int(imex_flag imex)
+  {
+    return (imex == imex_flag::imex_implicit) ? 2 : ((imex == imex_flag::imex_explicit) ? 1 : 0);
+  }
+  //! \brief Convert the matrix entry to an index of the arrays.
+  static int flag2int(matrix_entry imex)
+  {
+    return (imex == matrix_entry::imex_implicit) ? 2 : ((imex == matrix_entry::imex_explicit) ? 1 : 0);
+  }
```
quantumsteve: What about `static_cast<int>` like currently done on line 1487?

src/asgard_kronmult_matrix.hpp
```diff
@@ -1276,4 +1278,222 @@ struct matrix_list
...
+make_block_global_kron_matrix(PDE<precision> const &pde,
+                              adapt::distributed_grid<precision> const &dis_grid,
+                              options const &program_options,
+                              kronmult::block_global_workspace<precision> *workspace);
+
+template<typename precision>
+struct matrix_list
+{
+  //! \brief Makes a list of uninitialized matrices
+  matrix_list() {}
```
quantumsteve:
```suggestion
  matrix_list() = default;
```

src/asgard_kronmult_matrix.hpp
```diff
@@ -1276,4 +1278,222 @@ struct matrix_list
...
+
+  //! \brief Frees the matrix list and any cache vectors
+  ~matrix_list() {}
```
quantumsteve:
```suggestion
  ~matrix_list() = default;
```

src/asgard_kronmult_matrix.hpp
```diff
@@ -1276,4 +1278,222 @@ struct matrix_list
...
+                          options const &opts)
+  {
+    if (not kglobal)
+      make(entry, pde, grid, opts);
+    else
+      set_specific_mode(pde, grid, opts, imex(entry), kglobal);
+  }
+
+  //! \brief Clear the specified matrix
+  //void clear(matrix_entry entry)
```
quantumsteve: delete line?

src/device/asgard_glkronmult_bcpu.cpp
```diff
@@ -0,0 +1,671 @@
...
+            for (int j = 0; j < n; j++)
+              for (int s = 0; s < n; s++)
+                for (int l = 0; l < n; l++)
+                  for (int k = 0; k < n; k++)
+                    y[ipow<n, 5>() * w + ipow<n, 4>() * v + ipow<n, 3>() * p + n * n * s + n * l + k] +=
+                        x[ipow<n, 5>() * w + ipow<n, 4>() * v + ipow<n, 3>() * p + n * n * j + n * l + k] * A[j * n + s];
+    }
+    else if constexpr (dim == 2)
+    {
+      for (int w = 0; w < n; w++)
```
quantumsteve, line 228:
```suggestion
      ASGARD_PRAGMA_OMP_SIMD(collapse(7))
      for (int w = 0; w < n; w++)
```

src/device/asgard_glkronmult_bcpu.cpp
```diff
@@ -0,0 +1,671 @@
...
+{
+  //std::cout << " ------------------------------------------------------ \n";
+  constexpr int n2 = n * n;
+
+  int const num_vecs = dsort.num_vecs(dim);
+
+#ifdef _OPENMP
+  int const max_threads = omp_get_max_threads();
+#else
+  int max_threads = 1;
```
quantumsteve:
```suggestion
  int const max_threads = 1;
```

src/device/asgard_glkronmult_bcpu.cpp
```diff
@@ -0,0 +1,671 @@
...
+#endif
+
+  if (static_cast<int>(row_wspace.size()) < max_threads)
+    row_wspace.resize(max_threads);
+
+  int threadid = 0;
+#pragma omp parallel
+  {
+    int tid;
+#pragma omp critical
```
quantumsteve, line 295:
```suggestion
#pragma omp atomic
```

src/device/asgard_glkronmult_bcpu.cpp
```diff
@@ -0,0 +1,671 @@
...
+      for (int j = vec_begin; j < vec_end; j++)
+        xidx[dsort(ilist, dim, j)] = dsort.map(dim, j) * block_size;
+
+      // matrix-vector product using xidx as a row
+      for (int rj = vec_begin; rj < vec_end; rj++)
+      {
+        // row in the 1d pattern
+        int const row = dsort(ilist, dim, rj);
+
+        precision *const local_y = &y[xidx[row]];
```
quantumsteve:
```suggestion
        precision *const local_y = y + xidx[row];
```
or use [std::next](https://en.cppreference.com/w/cpp/iterator/next)

src/device/asgard_glkronmult_bcpu.cpp
```diff
@@ -0,0 +1,671 @@
...
+    vector2d<int> const &ilist, dimension_sort const &dsort,
+    std::vector<permutes> const &perms,
+    std::vector<int> const &flux_dir,
+    connect_1d const &conn_volumes, connect_1d const &conn_full,
+    std::vector<int> const &terms,
+    block_global_workspace<precision> &workspace)
+{
+  number_of_blocks_ = 0;
+
+  for (int t : terms)
```
quantumsteve:
```suggestion
  for (int const t : terms)
```

src/device/asgard_glkronmult_bcpu.cpp
```diff
@@ -0,0 +1,671 @@
...
+  {
+    // terms can have different effective dimension, since some of them are identity
+    permutes const &perm  = perms[t];
+    int const active_dims = perm.num_dimensions();
+    if (active_dims == 0)
+      continue;
+
+    for (size_t i = 0; i < perm.fill.size(); i++)
+    {
+      int dir = perm.direction[i][0];
```
quantumsteve:
```suggestion
```

src/device/asgard_glkronmult_bcpu.cpp
```diff
@@ -0,0 +1,671 @@
...
+
+      for (int d = 0; d < active_dims; d++)
+      {
+        dir = perm.direction[i][d];
```
quantumsteve:
```suggestion
        int const dir = perm.direction[i][d];
```

src/device/asgard_glkronmult_bcpu.cpp
```diff
@@ -0,0 +1,671 @@
...
+        global_cpu<precision>(num_dimensions, -1, block_size, ilist, dsort, dir, perm.fill[i][d],
+                              (perm.fill[i][d] == permutes::matrix_fill::both and flux_dir[t] != -1) ? conn_full : conn_volumes,
+                              std::vector<precision>{}, nullptr, nullptr, workspace.row_map);
+      }
+    }
+  }
+
+  for (int d = 0; d <= num_dimensions; d++)
+    number_of_blocks_ *= n;
```
quantumsteve: fm::pow or similar function for integers?

CMakeLists.txt
```diff
@@ -113,13 +113,24 @@ option (ASGARD_BUILD_DOCS "Build the documentation." OFF)
...
   option (KRON_MODE_GLOBAL "Global or local Kronecker products" OFF)
+  option (KRON_MODE_GLOBAL_BLOCK "Use block variant for global kron" OFF)
 else()
-  option (KRON_MODE_GLOBAL "Global or local Kronecker products" ON)
+  if (ASGARD_USE_CUDA)
+    option (KRON_MODE_GLOBAL "Global or local Kronecker products" ON)
+    option (KRON_MODE_GLOBAL_BLOCK "Use block variant for global kron" OFF)
+  else()
+    option (KRON_MODE_GLOBAL "Global or local Kronecker products" ON)
+    option (KRON_MODE_GLOBAL_BLOCK "Use block variant for global kron" ON)
```
quantumsteve, line 123: Do you want to expand CI coverage of the various kronmult implementaitons? In particular I don't think we have anything running `KRON_MODE_GLOBAL=ON` and `KRON_MODE_GLOBAL_BLOCK=OFF` on CPU.

src/asgard_kronmult_matrix.cpp
mkstoyanov: Where would we put this? Utils?

CMakeLists.txt
mkstoyanov: Eventually kronmult-global-block will replace kronmult-global and the block implementation is so much better (faster) that it makes no sense except for testing and reference study, e.g., if we find that something isn't working right.

I consider `KRON_MODE_GLOBAL=ON` and `KRON_MODE_GLOBAL_BLOCK=OFF` to be deprecated, except for the possibility that the new kronmult isn't working right just yet.

src/asgard_kronmult_matrix.cpp
quantumsteve: I'd put it next to [fm::two_raised_to](https://github.com/project-asgard/asgard/blob/ad62a76765ce3075692ba525ee4fcfc34cdb9711/src/fast_math.hpp#L17-L28)

src/asgard_kronmult_matrix.hpp
mkstoyanov: In older versions, when I pre-computed the inedexing, counting the flops was easy, e.g., just multiply the number of indexes by the flops per index. Here I have to do some work, so I don't want to count the flops unless the user explicitly asks for the flops.

Currently, we ask the flops every time we do the timing, so the first time we time the number of flops will be computed, saved and outputed to the console just as before. If I move the switch statement, then we will have output on every GMRES iteration.

src/asgard_kronmult_matrix.hpp
mkstoyanov: Yes, this is better, I just need to double-check that the order of the enums is the same everywhere.

src/device/asgard_glkronmult_bcpu.cpp
mkstoyanov: I know atomic-increment-add works but this is increment-assign, the increment-add uses 2 memory locations (the variable and the amount to increment) but the increment-assign uses 3 (tid, threadid and the +1).

I will test it but it is not clear to me.

src/device/asgard_glkronmult_bcpu.cpp
mkstoyanov: actually, just removed the entire line

src/asgard_kronmult_matrix.cpp
mkstoyanov: done!

src/device/asgard_glkronmult_bcpu.cpp
mkstoyanov: done!

src/asgard_kronmult_matrix.hpp
mkstoyanov: Done! The local kronmult implementation needs some more work, I hate the syntax in time advance that we have nested `#ifdef` and `if` based on solver and kronmult ... will be another PR.

src/device/asgard_glkronmult_bcpu.cpp
mkstoyanov: The `atomic` version doesn't even compile, the problem is that we are trying to do atomic change to two variables at the same time.

quantumsteve (dismissed):

mkstoyanov:
@quantumsteve Sorry but I put some pretty lines here and there, should be good for your re-review now. Nothing major has changed.

quantumsteve (approved):

PR #697, opened 2024-06-27
allows for single level to be defined in the cli

### LINKED ISSUE #367: default to uniform level across dimensions if only one level provided at command line
currently, users need to enter a starting level for each dimension in their selected PDE. e.g., `-l "4 4"` for a 2d PDE. Allow users to enter just one value, setting uniform starting level across dimensions.

mkstoyanov:
closes  #367

I've been meaning to do this in a while, now can pass single number to the `-l` option, e.g.,
```
./asgard -p continuity_6 -l 6
```

### DIFF
--- a/src/pde/pde_base.hpp
+++ b/src/pde/pde_base.hpp
@@ -652,7 +652,7 @@ public:
     // modify for appropriate level/degree
     // if default lev/degree not used
     auto const user_levels = cli_input.get_starting_levels().size();
-    if (user_levels != 0 && user_levels != num_dims_)
+    if (user_levels >= 2 && user_levels != num_dims_)
     {
       std::cerr << "failed to parse dimension-many starting levels - parsed "
                 << user_levels << " levels\n";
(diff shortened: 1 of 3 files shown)


src/pde/pde_base.hpp
```diff
@@ -668,6 +668,12 @@ class PDE
         d.set_level(num_levels);
       }
     }
+    else if (user_levels == 1)
+    {
+      auto const num_levels = cli_input.get_starting_levels()[0];
```
quantumsteve, line 673: `get_starting_levels` returns an `fk::vector<int>` causing an unnecessary call to `new` and `delete[]` on this line. Line 666 is worse, since it calls  get_starting_levels() for each dimensions.

Can we change the return type to `fk::vector<int> const &` without making more changes to the code?
https://github.com/project-asgard/asgard/blob/9500b5389749de28575f7031e08c54ae3c0879c8/src/program_options.cpp#L473

In hindsight, supporting `fk::vector<int>` was a mistake and makes it harder to call BLAS inside `fk::vector`

mkstoyanov: The code actually read into `std::vector<int>` but later copies into `fk::vector` and keeps rolling with the copies.

Fixing all that is a bigger refactor, at least we can do the `const &` return

quantumsteve (approved):

PR #699, opened 2024-07-04
new interpolation framework

mkstoyanov:
First draft the interpolation framework, the PR is already large, it's a good idea to merge it before we start piling in features.

### DIFF
--- a/src/asgard_interpolation.hpp
+++ b/src/asgard_interpolation.hpp
@@ -1,0 +1,134 @@
+#pragma once
+
+#include "asgard_interpolation1d.hpp"
+#include "asgard_kronmult_matrix.hpp"
+
+namespace asgard
+{
+#ifdef KRON_MODE_GLOBAL_BLOCK
+
+/*!
+ * \brief Handles the interpolation operations in multidimensional context
+ *
+ * The interpolation takes the number of dimensions, a volume connectivity
+ * object and kronmult workspace. Aliases are kept for the connect_1d
+ * and workspace objects, those will be held in the kron_operators.
+ */
+template<typename precision>
+class interpolation
+{
+public:
+  //! constructs and empty interpolation class
+  interpolation() : num_dimenisons_(0), block_size(0), workspace_(nullptr)
+  {}
+  //! convert to true if the class has been initialized
+  operator bool() const { return (num_dimenisons_ > 0); }
+
+  //! new interpolation method for the given dimensions and connect_1d
+  interpolation(int num_dimenisons, connect_1d const *conn_in,
+                kronmult::block_global_workspace<precision> *workspace)
+      : num_dimenisons_(num_dimenisons), block_size(fm::ipow(pterms, num_dimenisons_)),
+        perms(num_dimenisons_), wav1d(conn_in), workspace_(workspace)
+  {}
+
+  //! get the interpolation nodes for the cells
+  vector2d<precision> get_nodes(vector2d<int> const &cells) const
+  {
+    int const num_incell = fm::ipow(pterms, num_dimenisons_);
+
+    vector2d<precision> nodes(num_dimenisons_, num_incell * cells.num_strips());
+
+    std::array<int, max_num_dimensions> offsets;
+
+    for (int c = 0; c < cells.num_strips(); c++)
+    {
+      for (int d = 0; d < num_dimenisons_; d++)
+        offsets[d] = pterms * cells[c][d];
+
+      for (int i = 0; i < num_incell; i++)
+      {
+        int const inode = c * num_incell + i;
+
+        int t = i;
+        for (int d = num_dimenisons_ - 1; d >= 0; d--)
+        {
+          nodes[inode][d] = wav1d.node(offsets[d] + t % pterms);
+          t /= pterms;
+        }
+      }
+    }
+
+    return nodes;
+  }
+
+  //! convert projected coefficients to nodal values
+  void get_nodal_values(vector2d<int> const &cells,
+                        dimension_sort const &dsort,
+                        precision const proj[], precision nodal[]) const
+  {
+    kronmult::global_cpu(num_dimenisons_, pterms, block_size, cells, dsort, perms,
+                         wav1d.get_conn(), wav1d.get_eval_matrix(), proj, nodal,
+                         *workspace_);
+  }
+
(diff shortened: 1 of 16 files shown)


src/asgard_interpolation.hpp
```diff
@@ -0,0 +1,134 @@
+#pragma once
+
+#include "asgard_interpolation1d.hpp"
+#include "asgard_kronmult_matrix.hpp"
+
+namespace asgard
+{
+#ifdef KRON_MODE_GLOBAL_BLOCK
```
quantumsteve, line 8: Did you consider conditionally including this file and the associated asgard_interpolation.cpp and asgard_interpolation_tests.cpp based on the CMake flag?

src/asgard_interpolation1d.hpp
```diff
@@ -0,0 +1,101 @@
+#pragma once
+
+#include "asgard_indexset.hpp"
+
+namespace asgard
+{
+#ifdef KRON_MODE_GLOBAL_BLOCK
```
quantumsteve, line 7: same question as with `asgard_interpolation`, is this better done within CMake?

src/device/asgard_glkronmult_bcpu.cpp
```diff
@@ -335,7 +335,82 @@ void global_cpu(int64_t block_size,
...
+        int col_end   = conn.row_diag(row);
+
+        for (int c = col_begin; c < col_end; c++)
+        {
+          int64_t const xj = xidx[conn[c]];
+          if (xj != -1)
+          {
+            if constexpr (n == -1)
+#pragma omp atomic
+              number_of_blocks_ += 1;
```
quantumsteve: Can one use `number_of_blocks_ ++` with `omp atomic`?

src/device/asgard_glkronmult_bcpu.cpp
```diff
@@ -620,14 +842,65 @@ int64_t block_global_count_flops(
...
+    int dir = perm.direction[i][0];
+
+    global_cpu(num_dimensions, n, ilist, dsort, dir, perm.fill[i][0],
+               vconn, gvals, x, w1, workspace.row_map);
+
+    for (int d = 1; d < num_dimensions; d++)
+    {
+      dir = perm.direction[i][d];
+      global_cpu(num_dimensions, n, ilist, dsort, dir, perm.fill[i][d],
+                 vconn, gvals, w1, w2, workspace.row_map);
```
quantumsteve:
```suggestion
    auto &dir = perm.direction[i];

    global_cpu(num_dimensions, n, ilist, dsort, dir[0], perm.fill[i][0],
               vconn, gvals, x, w1, workspace.row_map);

    for (int d = 1; d < num_dimensions; d++)
    {
      global_cpu(num_dimensions, n, ilist, dsort, dir[d], perm.fill[i][d],
                 vconn, gvals, w1, w2, workspace.row_map);
```

src/asgard_interpolation1d.hpp
```diff
@@ -0,0 +1,101 @@
...
+  //! number of polynomial terms, e.g., 2 for linear basis
+  static constexpr int pterms = order + 1;
+  //! size of the matrix block
+  static constexpr int matsize = pterms * pterms;
+
+  //! null constructor
+  wavelet_interp1d() : conn(nullptr)
+  {}
+  //! cache interpolation points and transformation matrices
+  wavelet_interp1d(connect_1d const *conn_in) : conn(conn_in)
```
quantumsteve, line 34: Should `wavelet_interp1d` own the `conn` object?

src/asgard_interpolation.hpp
```diff
@@ -0,0 +1,134 @@
...
+public:
+  //! constructs and empty interpolation class
+  interpolation() : num_dimenisons_(0), block_size(0), workspace_(nullptr)
+  {}
+  //! convert to true if the class has been initialized
+  operator bool() const { return (num_dimenisons_ > 0); }
+
+  //! new interpolation method for the given dimensions and connect_1d
+  interpolation(int num_dimenisons, connect_1d const *conn_in,
+                kronmult::block_global_workspace<precision> *workspace)
```
quantumsteve, line 29: ownership of `workspace`?

src/asgard_interpolation1d.cpp
```diff
@@ -0,0 +1,612 @@
...
+        mat[i + 2] = -linear<precision>::iwav1R(x[i]);
+      }
+  }
+}
+
+template<int order, typename precision>
+void wavelet_interp1d<order, precision>::prepare_wmatrix()
+{
+  eval_matrix.resize(conn->num_connections() * matsize);
+  auto em = eval_matrix.begin();
```
quantumsteve, line 275:
```suggestion
  auto em = eval_matrix.cbegin();
```

src/asgard_interpolation.hpp
mkstoyanov: Doing several answers in one:

This PR has the code that can go through the steps of the interpolation procedure but lacks the link to the PDE and the rest of the code. I want to keep the PRs more manageable instead of pushing that is many thousands of lines (this is >2K already).

In the final design, those files will not be conditional but will be included with every combination of options. To get this right, I will need to reconcile interpolation with local kronmult and that's a mid size rework in it's own. The if-defs will go away, just like any CMake code that I use as an alternative.

The second point is that the connectivity and workspace will be shared with the kronmult matrix, so the owner will be the `kron_operatos` class that will then hold the interpolation and matrix (for the linear separable case). Right now the `kron_operatos` is holding the workspace and the matrix takes a non-owning alias. The matrix owns the connectivity but I will pull that one out too. 

Connectivity, workspace and the interpolation matrices can be reused even when the refinement and grid change Connectivity and matrices have to change if we add an extra level, otherwise they can stay the same and save on rebuilding/reallocating those. That would be the job of `kron_operatos` class, which is just more of exactly what it does now.

src/asgard_interpolation1d.cpp
mkstoyanov: `em` is used to write the entries of the matrix, it cannot be `const`

quantumsteve: :facepalm:

quantumsteve (approved):

PR #700, opened 2024-07-11
fix some math

mkstoyanov:
RMCE does not require 2 allocations and 3 loops of solution-size data.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -281,14 +281,14 @@ int main(int argc, char **argv)
           transformer, degree, time + pde->get_dt());
 
       // calculate root mean squared error
-      auto const diff = f_val - analytic_solution;
-      auto const RMSE = [&diff]() {
-        asgard::fk::vector<prec> squared(diff);
-        std::transform(squared.begin(), squared.end(), squared.begin(),
-                       [](prec const &elem) { return elem * elem; });
-        auto const mean = std::accumulate(squared.begin(), squared.end(), 0.0) /
-                          squared.size();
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -281,14 +281,14 @@ int main(int argc, char **argv)
...
-                          squared.size();
-        return std::sqrt(mean);
+      auto const RMSE = [&]() {
+        prec s{0}, d{0};
+        for (int i = 0; i < f_val.size(); i++)
+        {
+          d = f_val[i] - analytic_solution[i];
+          s += d * d;
+        }
+        return std::sqrt(s);
```
quantumsteve: The old code divides by `squared.size()` before taking the square root.

```suggestion
        prec s{0};
        for (int i = 0; i < f_val.size(); i++)
        {
          prec const d = f_val[i] - analytic_solution[i];
          s += d * d;
        }
        return std::sqrt(s / f_val.size());
```

quantumsteve (commented):
Do we want RMSE of the vector or per element?

src/main.cpp
mkstoyanov: Duh!

mkstoyanov:
> Do we want RMSE of the vector or per element?

"mean" error means average across all elements, per-element will have to average over time or some external dimension, or it is simply an error vector of some sort. The old code did compute it down to a scalar, even if it is not obvious due to the overuse of `auto`

quantumsteve (approved):

PR #703, opened 2024-07-19
allow the use of asgard as a library

mkstoyanov:
Major change in the build system and the potential usage:

* ASGarD can not be build and installed as a stand-alone library
   * does not work with custom build BLAS and HDF5, must use external installations
   * requires the `-DCMAKE_INSTALL_PREFIX` else it defaults to `/usr/local` and that's not desirable
* the installation can be tested with `make test_install`

* an ASGarD installation can be included in separate project
* the users have the ability to create a new PDE that exists outside of the main library
* for both cases, see the examples and specifically those in `<CMAKE_INSTALL_PREFIX>/share/asgard/exmaples`

PS: The install prefix defaults to `<build-folder>/asgard_install/` to avoid issues with the system files in `/usr/local`


mkstoyanov:
@quantumsteve We should add `-DASGARD_BUILD_HDF5=ON` to the failing build. I don't want ASGarD to build large third party projects unless the user explicitly asks for that.

I also want to figure out how to "install" the build HDF5 (and OpenBLAS) but that will take a while as I have to test build those a few times. I want to keep the project rolling, so if it takes too long, I will add the HDF5 install in another PR.

PS: The code can now install both HDF5 and OpenBLAS.

mkstoyanov:
@stefan-schnake Take a look at the "example" of continuity2d and see if my language is fitting what we want to do.

Also, this may help a lot with your work.

quantumsteve:
> @quantumsteve We should add `-DASGARD_BUILD_HDF5=ON` to the failing build. I don't want ASGarD to build large third party projects unless the user explicitly asks for that.

buildscript's updated.

src/pde.hpp
```diff
@@ -54,8 +54,29 @@ namespace asgard
...
 
-template<typename P>
+template<typename type_to_test, typename control_type = void>
+struct has_precision_mode : std::false_type
+{};
+
+template<typename type_to_test>
+struct has_precision_mode<type_to_test, std::void_t<typename type_to_test::precision_mode>> : std::true_type
+{};
+
```
quantumsteve: May we add a [variable template](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3854.htm) `has_precision_mode_v`?

mkstoyanov: Yes but there's also a better way to do this, no need for the helper template.

PR #705, opened 2024-07-23
python plotting

mkstoyanov:
Added `asgard.py` python module. We can now read HDF5 files with wavelet data and reconstruct the solution on an arbitrary set of points for plotting purposes.

In progress: will add examples once we have the installation logic from #703

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,7 @@ endif()
 
 # set up possible commandline input variable defaults (override with -D)
 include(CMakeDependentOption)
+option (BUILD_SHARED_LIBS "Build the ASGarD library with dynamic or static linking" ON)
 option (ASGARD_BUILD_TESTS "Build tests for asgard" ON)
 option (ASGARD_BUILD_OPENBLAS "Download and build our own OpenBLAS" OFF)
 option (ASGARD_IO_HIGHFIVE "Use the HighFive HDF5 header library for I/O" OFF)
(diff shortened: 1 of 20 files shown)


mkstoyanov:
@quantumsteve  Do you know why the failing build is using libblis.so and libflame.so in place of BLAS/LAPACK or OpenBLAS. The error seems related to libflame.so being build against a different version of libgcc

quantumsteve:
Historically, we encountered CI failures with the OpenBLAS version distributed by Ubuntu. See, for example, https://github.com/OpenMathLib/OpenBLAS/issues/3494.

mkstoyanov:
> Historically, we encountered CI failures with the OpenBLAS version distributed by Ubuntu. See, for example, [OpenMathLib/OpenBLAS#3494](https://github.com/OpenMathLib/OpenBLAS/issues/3494).

That must have been 20.04 or even 18.04. I've been using 22.04 and 24.04 with OpenBLAS and had no issues.

I don't mind sticking to libflame but I will have to check which container it is so I can rebuild and hopefully reproduce locally where I can diagnose.

mkstoyanov:
Managed to reproduce with the container, there is a warning with relinking of libblis against libm and then some of the tests randomly  crash with a segfault while others work. I say randomly, because I went back several commits including before the recent push of changes. I think Ubuntu 22.04 updated libm and forgot to relink libblis against the updated library. The all is working with both OpenBLAS on 22.04 and libblis on 24.04.

We should switch the container to OpenBLAS. Would it work if I make the PR or the containers have to be manually pushed to AWS?

quantumsteve:
> We should switch the container to OpenBLAS. Would it work if I make the PR or the containers have to be manually pushed to AWS?

Yes, the new container must be pushed to AWS
https://github.com/project-asgard/asgard/wiki/AWS-CodeBuild-CI

mkstoyanov:
> > We should switch the container to OpenBLAS. Would it work if I make the PR or the containers have to be manually pushed to AWS?
> 
> Yes, the new container must be pushed to AWS https://github.com/project-asgard/asgard/wiki/AWS-CodeBuild-CI

Can you please do that?

quantumsteve:
> Can you please do that?

done

PR #706, opened 2024-07-24
Gone by the wind

mkstoyanov:
Now we only need to move by one degree more ...

@hauckc Do you like this better? You will have to change your PDE after we merge this and you sync your branch.

### DIFF
--- a/src/pde/pde_advection1.hpp
+++ b/src/pde/pde_advection1.hpp
@@ -134,8 +134,8 @@ private:
       coefficient_type::div, // type
       g_func_0,              // g func
       nullptr,               // lhs
-      flux_type::downwind,   // flux = "-1" (downwind), "0" (central), "+1"
-                             // (upwind)
+      flux_type::upwind,     // flux = "-1" (upwind), "0" (central), "+1"
+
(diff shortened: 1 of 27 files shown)


src/pde/pde_advection1.hpp
```diff
@@ -134,8 +134,8 @@ class PDE_advection_1d : public PDE<P>
       coefficient_type::div, // type
       g_func_0,              // g func
       nullptr,               // lhs
-      flux_type::downwind,   // flux = "-1" (downwind), "0" (central), "+1"
-                             // (upwind)
+      flux_type::upwind,     // flux = "-1" (downwind), "0" (central), "+1"
```
stefan-schnake:
```suggestion
      flux_type::upwind,     // flux = "-1" (upwind), "0" (central), "+1"
```

stefan-schnake (commented):
Looks good other than the comment change.

mkstoyanov:
The main question here is whether we want to switch upwind-downwind or just move the operator term definitions to the left-hand-side. Or are those orthogonal ideas?

We can discuss this on Wednesday and figure out the correct solution.

PR #732, opened 2024-08-19
Added more infiles

stefan-schnake:
Added infiles for more complicated problems.

### DIFF
--- a/infiles/riemann1x3vmg.txt
+++ b/infiles/riemann1x3vmg.txt
@@ -1,0 +1,31 @@
+-title    : Riemann 1x3v
+-subtitle : mixed grid, quadratic basis
+
+# df / dt + v d_x f = \nu C(f)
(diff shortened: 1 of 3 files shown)


infiles/riemann1x3vmg.txt
```diff
@@ -0,0 +1,31 @@
...
+-pde : riemann_1x3v
+
+# disctretization in space
+# mixed grid with level 7 full grid in x 
+# and level 6 sparse grid in v.
+# quadratic basis
+-grid         : mixed
+-split        : 1
+-start-levels : 6 6 6 6
+-max_level    : 6
```
mkstoyanov, line 15: Don't need to change this but if we skip the max_level then the largest of the start level values will be used.

mkstoyanov:
@stefan-schnake  This is great! Thanks a lot!

mkstoyanov (approved):

PR #741, opened 2024-09-19
fix osx issue with reading floats

mkstoyanov:
This should fix the issue with float precision in Mac.

Basically, the `>>` fails to read the double data as floats.

### DIFF
--- a/src/asgard_matlab_utilities.cpp
+++ b/src/asgard_matlab_utilities.cpp
@@ -318,9 +318,11 @@ fk::vector<P> read_vector_from_txt_file(std::filesystem::path const &path)
   // create output vector
   fk::vector<P> values(num_elems);
 
+  double x = 0;
   for (auto i = 0; i < num_elems; ++i)
   {
-    infile >> values(i);
+    infile >> x;
+    values(i) = static_cast<P>(x);
   }
 
   return values;


quantumsteve:
What do you think about creating a minimal reproducible example for libc++?

quantumsteve (approved):
Let's get this into develop and rebase #738.

mkstoyanov:
@quantumsteve 

I'm overriding the test requirement, we have too many tests now, we should disable the duplicates from AWS.

mkstoyanov:
> What do you think about creating a minimal reproducible example for libc++?

We can do that but I also want to check what the standard says too.

quantumsteve:
> @quantumsteve
> 
> I'm overriding the test requirement, we have too many tests now, we should disable the duplicates from A

I've disabled the duplicate tests from running on pull requests but haven't figured out why GitHub is still waiting for results from them.

mkstoyanov:
> > @quantumsteve
> > I'm overriding the test requirement, we have too many tests now, we should disable the duplicates from A
> 
> I've disabled the duplicate tests from running on pull requests but haven't figured out why GitHub is still waiting for results from them.

So we're waiting on tests that will never run.

Must be something in the hooks, there must be something in the github setup that is calling AWS. Maybe something in the AWS has to be force-updated ... ugh!

PR #770, opened 2025-03-15
removed the old relaxation pde

mkstoyanov:
* remove the relaxation pde from the old api


mkstoyanov:
* remove the relaxation pde from the old api

PR #787, opened 2025-05-06
Cleanup quadrature

mkstoyanov:
* convert the quadrature routines to std-vector


mkstoyanov:
* failing on older compilers, have to switch system to test and it is better to not stress the CI

PR #828, opened 2025-11-20
Lbmd

mkstoyanov:
* Lenard-Bernstein using interpolation


mkstoyanov:
I guess putting 6D PDEs in the free CI is not a good idea.

PR #848, opened 2026-06-23
Multi-dimensional Poisson Solver

TTBurnett:
The capabilities of the Poisson solver were extended to multiple spatial dimensions. These changes allow electrostatic plasmas to be simulated in 2 and 3 spatial dimensions. The electric field resulting from the Poisson solve is treated as a special type of moment, as such the poisson solver was moved to the moment manager. Additionally a new test src/pde/two_stream_md.cpp was added for the multi-dimensional poisson solver. Other key changes include: electric moments were added to src/asgard_momentset.hpp, domain_id struct was moved to src/asgard_momenset.hpp for use with electric moments, and methods to register and manage electric moments were added to src/asgard_pde.

Two stream rollup in 2D2V (~10 min to run with -a 1.e-3 -t 20):
Screenshot 2026-06-23 at 11 52 09 AM


src/asgard_discretization.cpp
```diff
@@ -390,6 +401,34 @@ discretization_manager<precision>::project_function(
 
 template<typename precision>
 std::vector<precision> discretization_manager<precision>::get_moment(moment_id id) const {
+  if (terms.moms.needs_poisson(id)) {
+    std::visit([&](auto &p) {
+      if constexpr (not std::is_same_v<std::decay_t<decltype(p)>, poisson::poisson_none>)
```
mkstoyanov: what is "posson_none", isn't that "monostate"?

src/asgard_discretization.cpp
```diff
@@ -250,9 +262,8 @@ template<typename precision>
 void discretization_manager<precision>::compute_moments_local(
     group_id gid, std::vector<precision> const &f) const
 {
-  terms.moms.cache_moments(gid, terms.grid, f);
+  terms.moms.cache_moments(gid, terms.grid, f, terms.conn, terms.hier, terms.kwork);
   terms.moms.load_interp(gid, terms.interp, terms.kwork);
```
mkstoyanov: Merge `load_interp` logic into `cache_moments`

src/pde/two_stream_md.cpp
```diff
@@ -0,0 +1,539 @@
...
+
+  pde += asgard::term_md<P>(std::vector<asgard::term_1d<P>>{
+    asgard::term_identity{},
+    asgard::term_div<P>(1, asgard::flux_type::downwind, asgard::boundary_type::periodic),
+    asgard::term_identity{},
+    asgard::term_volume<P>(negative)
+    });
+
+#ifdef ASGARD_USE_GPU
+  auto gpu_md_positive_x = [=](int64_t const num, P t, P const x[], asgard::momentset_gpu<P> const &moments, P const f[], P fx[]) {
```
mkstoyanov: `num` doesn't need to be const

src/pde/two_stream_md.cpp
```diff
@@ -0,0 +1,539 @@
...
+    // integral of moment 0 by moment 1, by delta_ij orthogonality of the basis
+    // just sum up the product of the coefficients
+    P mv0 = 0;
+    for (size_t j = 0; j < mom0.size(); j++)
+      mv0 += mom0[j] * momp0[j];
+    P mv1 = 0;
+    for (size_t j = 0; j < mom0.size(); j++)
+      mv1 += mom0[j] * momp1[j];
+
+    std::cout << "X momentum error: " << mv0 << "\n";
```
mkstoyanov: Comment out for the PR, do not delete

src/asgard_momentset.hpp
```diff
@@ -41,11 +68,24 @@ struct moment
...
   }
+  //! check whether a moment is an electric field moment
+  bool is_electric() const {
+    for (int const pow : pows)
+      if (pow == electric_flag) return true;
+    return false;
+  }
+  //! get the direction of the electric field corresponding to the moment
+  int get_electric_direction() const {
+    rassert(is_electric(), "The moment must be an electric field moment to get its direction");
```
mkstoyanov: The runtime assert `rassert()` is always executed, effectively this method is looping over the moments twice.

Put the rassert() down where the unreachable statement is.

src/asgard_momentset.hpp
```diff
@@ -63,6 +103,10 @@ struct moment
 
   //! holds the powers
   std::array<int, max_mom_dims> pows;
+
+  private:
+    static int const unset_flag = -1;
+    static int const electric_flag = -0xef; // ef for electric field
```
mkstoyanov: `ef` is clever, but comment what is the actual numeric value? 2^7 - 1?

src/asgard_momentset.hpp
```diff
@@ -130,6 +174,13 @@ class moments_list {
     moms_.push_back(mom);
     return moment_id{static_cast<int>(moms_.size() - 1)};
   }
+  //! returns the ID of the moment if it exists, otherwise it returns an unset moment ID
+  moment_id get_check_id(moment const &mom) const {
+    for (int i = 0; i < static_cast<int>(moms_.size()); i++)
```
mkstoyanov: Perfect place to use `iindexof(moms_)`

src/asgard_poisson.cpp
```diff
@@ -0,0 +1,313 @@
...
+  : num_dims(num_pos), pdof(hier.degree() + 1), iter_solve(iter_solve_func), mom0(m0)
+{
+  rassert((num_dims > 1) and (num_dims <= max_pos_dims), "poisson_md should only be used for 2 or 3 spatial dimensions");
+
+  // set up moms_electric
+  for (int const d : iindexof(num_dims))
+    moms_electric[d] = mlist.get_check_id(moment::electric(dimension_id(d)));
+
+  // set up the terms for the laplician: laplacian(f) = div(grad(f))
+  term_1d<P> div = term_div<P>(-1, flux_type::upwind, boundary_type::periodic);
```
mkstoyanov: This is probably `boundary_type::none`

src/asgard_poisson.cpp
```diff
@@ -0,0 +1,313 @@
...
+  term_1d<P> grad = term_grad<P>(1, flux_type::upwind, boundary_type::periodic);
```
mkstoyanov: I would have to think about this, if we remove the average, then this may have to be homogeneous BC.

src/asgard_poisson.cpp
```diff
@@ -0,0 +1,313 @@
...
+  term_1d<P> laplacian({div, grad});
+  laplacian.set_penalty(10);
```
mkstoyanov: Should be `1 / dx` ... well, we need 3 matrices.

src/asgard_momentset.hpp
```diff
@@ -17,18 +35,27 @@ struct moment
...
+  moment(int pv1) : pows{pv1, unset_flag, unset_flag} {}
   //! create a 2D moment with the given powers
-  moment(int pv1, int pv2) : pows{pv1, pv2, -1} {}
+  moment(int pv1, int pv2) : pows{pv1, pv2, unset_flag} {}
   //! create a 3D moment with the given powers
   moment(int pv1, int pv2, int pv3) : pows{pv1, pv2, pv3} {}
   //! number of valid powers
   int num_dims() const {
+    if (is_electric())
+      return max_mom_dims;
```
mkstoyanov: why? 2x2v electic moment should return 2 for `num_dims()`

src/asgard_poisson.hpp
```diff
@@ -0,0 +1,282 @@
...
+  //! poisson solve using periodic boundary conditions
+  void solve_periodic(std::vector<P> const &density, momentset<P> &moms,
+                      sparse_grid const &position_grid, connection_patterns const &conn,
+                      kronmult::workspace<P> &work)
+  {
+    solve(density, moms, position_grid, conn, work, poisson_bc::periodic);
+  }
+
+  //! indicates whether the solver has been initialized
+  operator bool() const { return (num_dims >= 0); }
```
mkstoyanov: Should this be `num_dims > 0`, I'm not sure 0d Poisson solve makes sense.

src/asgard_poisson.cpp
```diff
@@ -0,0 +1,313 @@
...
+  else {
+    std::vector<P> fp2d(pdof * pdof);
+    std::copy_n(p2d[0], pdof * pdof, fp2d.begin());
+    fill_pattern(fp2d.data(), diag);
+  }
+  derivative_mat = hier.diag2hierarchical(diag, max_level, conn);
+
+  // set up scaling for each direction
+  for (int const d : iindexof(num_dims))
+    dim_scalings[d] = static_cast<P>(num_cells) / (xright[d] - xleft[d]);
```
mkstoyanov: Think about this for a minute, the variable has 2 words: `dim` and `scaling`, which one is more important?

How about `derivative_scale[d]`, the long word is "derivative" which is the logical "group" where the variable belongs to. Then we have the shorter `scale` and it's clear that we are getting a value per dimension.

The size of words is a way to think about what you're emphasizing. The more dramatic example would be `number_d` as opposed to `num_dimensions` or even `num_dims`.

src/asgard_poisson.cpp
```diff
@@ -0,0 +1,313 @@
...
+  // get the electric field from the potential
+  for (int const d : iindexof(num_dims))
+  {
+    moment_id const &mid = moms_electric[d];
+    if (mid == moment_id::unset()) continue;
+    moms[mid].resize(n);
+    kronmult::permutes perm = std::vector<int>{d, };
+    block_cpu(pdof, position_grid, conn, perm, derivative_mat,
+              P{1.0}, potential.data(), P{0.0}, moms[mid].data(), work);
+    smmat::scal(n, dim_scalings[d], moms[mid].data());
```
mkstoyanov: This adds another sweep of the data, which is not needed when you're calling the method above with `P{1.0}`. You should use `dim_scaling[d]` in place of `P{1.0}` then there is no reason to scale.

src/asgard_poisson.cpp
```diff
@@ -0,0 +1,313 @@
...
+      block_cpu(pdof, position_grid, conn, tentry.perm, tentry.coeffs, alpha, x, b, y, work);
+      b = 1; // next iteration appends on y
+    }
+  };
+
+  // 7. Execute the Solve
+  int num_iters = iter_solve(apply_lhs, rhs, potential);
+
+  // 8. Anchor the resulting potential's mean to 0
+  if (bc == poisson_bc::periodic) {
```
mkstoyanov: Not needed.

src/asgard_poisson.hpp
```diff
@@ -0,0 +1,282 @@
...
+                        connection_patterns const &conn, kronmult::workspace<P> &work, poisson_bc const bc);
+  #ifdef ASGARD_USE_GPU
+  // Solves for just the electric potential, used as a substep inside the solver
+  void solve_potential_(gpu::vector<P> const &density, sparse_grid const &grid,
+                        connection_patterns const &conn, kronmult::workspace<P> &work, poisson_bc const bc);
+  #endif
+
+  int num_dims = -1;
+  int pdof = -1; 
+  moment_id mom0 = moment_id{-1};
```
mkstoyanov: This should be `moment_id mom0 = moment::unset();`

src/asgard_poisson.hpp
```diff
@@ -0,0 +1,282 @@
...
+namespace asgard::poisson
+{
+
+/*!
+ * \internal
+ * \brief Boundary conditions for the Poisson solver
+ *
+ * \endinternal
+ */
+enum class poisson_bc
```
mkstoyanov, line 45: This leads to a repetition, `poisson::poisson_bc`

There are 2 ways to namespace variables, you can use `namespace` and `name_` prefix, you can mix-and-match but try to avoid repetition.

Why not call this `boundary`. Then we can have `poisson::boundary::dirichlet` which is pretty clear, and inside the namespace we have just `boundary::dirichlet`.

src/asgard_solver_tests.cpp
```diff
@@ -6,11 +6,11 @@ using namespace asgard;
 // returns the result from comparison against the du_ref, which should be u_x
 template<typename P>
 P test_poisson(std::function<P(P)> du_ref, std::function<P(P)> rhs, P xleft, P xright,
-               P dleft, P dright, solvers::poisson_bc const bc, int degree, int level)
+               P dleft, P dright, poisson::poisson_bc const bc, int degree, int level)
```
mkstoyanov: Move the 1D Poisson tests to `asgard_poisson_tests.cpp`, the build system will pickup the filename automatically.

src/asgard_term_build.cpp
```diff
@@ -8,12 +8,13 @@ namespace asgard
...
 
 template<typename P>
-term_entry<P>::term_entry(term_md<P> tin)
+term_entry<P>::term_entry(term_md<P> tin, moments_list const &mlist)
   : tmd(std::move(tin)), has_poisson(false)
 {
   assert(not tmd.is_chain());
   if (tmd.is_interpolatory()) {
-    return; // interpolation poisson dependence goes here
+    has_poisson = tmd.is_electric(mlist);
```
mkstoyanov: I don't like having this logic here. The constructor of the term-manager (inside term_build.cpp around line 415) has logic to process the Poisson dependencies.

Let's set all term-entries to `has_poisson(false)` and if this is needed, the value is updated for all entries in one place at the constructor of the term-manager. Basically, the `term_entry` should have only the logic for the kronmult permutations.

src/asgard_term_build.cpp
```diff
@@ -773,16 +774,19 @@ void term_manager<P>::build_raw_mat(
     case operation_type::volume:
       switch (t1d.depends()) {
         case term_dependence::electric_field_only:
+        {
+          moment_id mid_electric = moms.find_id(moment::electric(dimension_id(d)));
           if (t1d.rhs()) {
             // using w1 as workspaces, it probably has enough space already
             size_t const n = kwork.w1.size();
-            t1d.rhs(moms.poisson_level(), kwork.w1);
+            t1d.rhs(moms.get_cached_raws()[mid_electric], kwork.w1);
```
mkstoyanov: This should not be the raw moment but the one completed to the max-level.

Now I'm curious why the test are passing and how come we're not calling this method.

src/asgard_term_manager.hpp
```diff
@@ -396,7 +396,7 @@ struct term_manager
   void build_const_terms(int const tid,
                          precon_method precon = precon_method::none, P alpha = 0);
   //! rebuild term[tmd][t1d], assumes non-identity
-  void rebuild_term1d(term_entry<P> &tentry, int const dim, int level,
+  void rebuild_term1d(term_entry<P> &tentry, int const dim, int const level,
```
mkstoyanov: No need to make this `const`

src/asgard_term_manager.hpp
```diff
@@ -428,6 +428,8 @@ struct term_manager
   //! assign compute resources to the terms
   void assign_compute_resources();
 
+  friend class discretization_manager<P>;
```
mkstoyanov: What is it that the discretization manager needs to access as a "friend".

The term manager has lots of public members, perhaps we can just expose one of those, especially if it's about the build term1d then we should just make the function public.

mkstoyanov (changes requested):
Make sure you enable `-Wall -Wextra -Wshadow -pedantic` for your compiler flags.

src/asgard_momentset.hpp
TTBurnett: Should I change get_add_id and get_id as well? They all use the static cast syntax I copied

src/asgard_momentset.hpp
TTBurnett: One solution is for num_dims() to return the number of position dimensions for an electric moment (since the values correspond to spatial directions) and it will return the number of velocity dimensions for a regular moment since pows correspond to velocity dimensions

src/asgard_poisson.cpp
TTBurnett: I never really thought about naming this way but it makes sense, thanks!

src/asgard_poisson.hpp
TTBurnett: Should this same logic be applied to the Poisson classes poisson_1d and poisson_md?

src/asgard_term_build.cpp
TTBurnett: I ended up removing has_poisson from term_entry entirely since the logic for finding poisson dependencies is inside the term manager anyways. You can fill the std::vector<bool> has_poisson_ in the term manager just using term_entry.tmd.is_electric() so that is what I did instead. Let me know if you think this should be done differently

src/asgard_poisson.hpp
TTBurnett: I'm thinking about going with poisson::solver_md and poisson::solver_1d

src/asgard_term_manager.hpp
TTBurnett: I primarily did this to be consistent with the fact that dim is const, I will undo that change

src/asgard_term_manager.hpp
TTBurnett: Yes it is for build term1d, I will make that public instead

src/asgard_pde.hpp
```diff
@@ -1249,6 +1249,26 @@ class term_md
...
+  bool is_electric(moments_list const &moments) const {
+    auto check_poisson = [](term_1d<P> const &single)
+    -> bool {
+      return (single.depends() == term_dependence::electric_field or
+              single.depends() == term_dependence::electric_field_only);
+    };
+    switch(mode_) {
+      case mode::chain:
+        for (term_md<P> const &ch : chain_)
+          return ch.is_electric(moments);
```
mkstoyanov: These for-loops are bogus, each is missing an if-statement, missing breaks and a return false if non of the tests pass.

src/asgard_poisson.hpp
TTBurnett: This is done now

src/asgard_term_build.cpp
TTBurnett: The level is used instead of the raws here now

src/asgard_moment_manager.cpp
```diff
@@ -73,6 +74,53 @@ moment_manager<P>::moment_manager(pde_domain<P> const &domain, int max_level,
...
+                                    build_term_func<P> build_func,
+                                    iter_solve_func<P> iter_func,
+                                    iter_solve_func_gpu<P> iter_func_gpu)
+{
+  if (mlist.has_electric()) {
+    moment_id const m0 = find_id(moment::zero(num_vel_));
+    if (num_pos_ == 1) {
+      moment_id const melectric = find_id(moment::electric(dimension_id(0), num_pos_));
+      poisson_solver = poisson_1d<P>(hier.degree(), xleft[0], xright[0],
+                                              grid.current_level(0), m0, melectric);
```
mkstoyanov: Code alignment, nitpick but let's keep things consistent:
```
poisson_solver = poisson_1d<P>(hier.degree(), xleft[0], xright[0],
                                                           grid.current_level(0), m0, melectric);
```

src/asgard_test_pdes.hpp
```diff
@@ -395,14 +395,16 @@ double get_qoi_indicator(asgard::discretization_manager<P> const &disc) {
     int const level0   = disc.get_grid().current_level(0);
     int const num_cell = fm::ipow2(level0);
     P const dx         = disc.domain().length(0) / num_cell;
+    int const npos     = disc.domain().num_pos();
```
mkstoyanov, line 398: unused variable

src/asgard_poisson.hpp
```diff
@@ -0,0 +1,276 @@
...
+
+  //! indicates whether the solver has been initialized
+  operator bool() const { return (num_dims > 0); }
+
+  //! returns the id for the zero moment
+  moment_id const &moment0() const { return mom0; }
+
+  #ifdef ASGARD_USE_GPU
+  //! initialize Poisson solver over the multi-dimensional domain
+  poisson_md(int const num_pos, int const max_level, std::array<P, max_num_dimensions> const &xleft,
```
mkstoyanov: For the `poisson_1d`  case, we implement the solver to work on the CPU only. When using the 1d solver in GPU mode, we move the moment data to the CPU, compute the E-field and we store the result on the CPU, since it will be needed on the CPU (by far the most likely case).

For the `poisson_md`, it is useful to be able to initiate the solution either from the CPU or the GPU, but since the solver is so much more expensive, in GPU mode when starting from the CPU, we move the data to the GPU, solve using the fast GPU and then move the result to the CPU.

Thus, `poisson_md` should need only one `iter_solve` method, the GPU mode does not need the CPU method.

Another important note about the `iter_solve` functions is that this operation prevents relocations. If you capture the `this` pointer into a lambda and pass that down, then there are two options. If the method/class accepting the lambda uses it but does not store a copy of the lambda (and respectively the pointer), then everything is good. However, if the class stores a copy, as it is the case of `iter_solve`, then we are in danger zone. If the user moves the discretization manager, then we will get a segfault on the next call to `solve()`.

There are two possible solutions here:
1. Do not store `iter_solve` and make as input to the `poisson_md::solve()` method. That way, it will be used and discarded and a fresh value of `this` will be used on each call.

2. Split the logic between the solver and the workspace into two different objects. So that the solver will hold high level data, e.g., iteration tolerance, while the workspace can be reused across different solvers.

We should use the second method, but that is for another PR. Leave this one as-is, we are not currently moving the discretization manager in any of our examples.

mkstoyanov (changes requested):
Almost there!

src/asgard_moment_manager.cpp
```diff
@@ -523,29 +547,62 @@ void moment_manager<P>::mcompute(sparse_grid const &grid, moment_id id,
...
+        has_poisson = true;
+      else
+        mcompute(grid, *mid, state, raw_vals.get(*mid));
       full_level.get(*mid).resize(0);
     }
+    if (has_poisson) solve_poisson(conn, hier, interp, work);
+  }
+
+  #ifdef ASGARD_USE_GPU
+  bool const skip_poisson_moments = true; // these moments are already interpolated on GPU
```
mkstoyanov: First, this should be `bool constexpr` but a better way is to define a lambda
```
auto skip_poisson = [&](moment_id mid) -> bool {
  #ifdef USE_GPU
    return needs_poisson(mid);
  #else  
    return false;
  #endif
}
```

src/asgard_moment_manager.cpp
```diff
@@ -846,6 +946,27 @@ void moment_manager<P>::compute_moments(
...
+      if (mom.is_electric()) continue; // skip electric field moments
+
+      // using work[g] as workspace without resizing
+      P* target_data = work1[g].data();
+
+      // using poisson_md.d_density for long term storage, needed for subsequent poisson solve
+      if (has_poisson() and mom == moment::zero(num_vel_)) {
+        std::visit([&](auto &p) {
+          if constexpr (std::is_same_v<std::decay_t<decltype(p)>, poisson_md<P>>) {
+            p.d_density.resize(num_entries);
```
mkstoyanov: 1. Lots of CUDA examples have the notation `d_` to define a device variable, but in ASGarD we are using `gpu_`

2. Why save the density, why not compute the moment into `w1` as usual and then if this is the zeroth moment, call `solve_poisson_gpu()` and pass in the current `w1`.

src/asgard_discretization.cpp
```diff
@@ -390,14 +399,41 @@ discretization_manager<precision>::project_function(
...
 std::vector<precision> discretization_manager<precision>::get_moment(moment_id id) const {
+  if (terms.moms.needs_poisson(id)) {
+    rassert(domain_.num_pos() > 1, "Only the moment level is computed for poisson_1d, use get_moment_level instead")
+    std::visit([&](auto &p) {
+      if constexpr (not std::is_same_v<std::decay_t<decltype(p)>, std::monostate>)
+        terms.moms.cache_moment(p.moment0(), terms.grid, state);
+      else
+        throw std::runtime_error("an electric moment was requested but no poisson solver is set");
+    }, get_poisson());
+    terms.moms.solve_poisson(terms.grid, terms.conn, terms.hier, terms.interp, terms.kwork);
```
mkstoyanov, line 410: `compute_moments_()` should load the electric moments, the get method should return what's loaded, should not compute.

src/asgard_moment_manager.hpp
```diff
@@ -106,14 +107,22 @@ class moment_manager {
   momentset<P> const &get_cached_interps() const { return interps; }
   //! return the current position grid where the raw moments are defined
   sparse_grid const &get_position_grid() const { return pos_grid; }
+  //! return the poisson solver
+  auto &get_poisson() const { return poisson_solver; }
```
mkstoyanov: Make the `poisson_solver` public and remove this method.

mkstoyanov: Well, if this is used only for the "update_level" method, then keep the solver private and add the update level method.

src/asgard_momentset.hpp
```diff
@@ -17,18 +35,25 @@ struct moment
...
+  //! produces a moment for the electric field in the dim dimension
+  //! i.e. dim = dimension_id(0) corresponds to E_x
+  static moment electric(dimension_id dim, int num_pos_dims) {
+    moment m = moment::zero(num_pos_dims);
+    m.pows[dim()] = electric_flag;
+    return m;
+  }
   //! creating a placeholder (invalid) moment
-  moment() : pows{-1, -1, -1} {}
+  moment() : pows{unset_flag, unset_flag, unset_flag} {}
```
mkstoyanov: Add a static-assert here that `max_mom_dims` is 3. It is an assumption in all of these methods, but having it in one place is enough.

src/asgard_momentset.hpp
```diff
@@ -41,11 +66,32 @@ struct moment
...
+  bool is_electric() const {
+    for (int const pow : pows)
+      if (pow == electric_flag) return true;
+    return false;
+  }
+  //! get the direction of the electric field corresponding to the moment
+  int get_electric_direction() const {
+    for (int const i : iindexof(pows))
+      if (pows[i] == electric_flag) return i;
+    rassert(false, "The moment must be an electric field moment to get its direction");
```
mkstoyanov: Just `throw std::runtime_error()` here, no need to assert anything.

src/asgard_poisson.cpp
```diff
@@ -0,0 +1,306 @@
...
+    {
+      block_gpu(gpu::device{0}, pdof, position_grid, conn, tentry.perm, tentry.gpu_coeffs,
+                alpha, x, b, y, work, tentry.coeffs);
+      b = 1; // next iteration appends on y
+    }
+  };
+
+  // Execute the Solve
+  compute->set_device(gpu::device{0});
+  std::ignore = iter_solve(apply_lhs, density, gpu_potential);
```
mkstoyanov: Store the number of iterations to a local variable and ignore the variable. This reads as if we are ignoring the solver.

mkstoyanov (approved):

PR #849, opened 2026-07-09
Preconditioned Conjugate Gradient Solver

TTBurnett:
I added a preconditioned conjugate gradient solver for solving the Poisson equation. Unit tests were also added for the multi-dimensional Poisson solver.


src/pde/two_stream.cpp
```diff
@@ -144,7 +144,7 @@ asgard::pde_scheme<P> make_two_stream(asgard::prog_opts options) {
 
   auto ic_v = [](std::vector<P> const &v, P /* time */, std::vector<P> &fv) ->
     void {
-      P const c = P{1} / std::sqrt(PI);
+      P const c = P{2} / std::sqrt(PI);
```
mkstoyanov, line 147: Why change the initial condition here?

src/asgard_discretization.hpp
```diff
@@ -843,7 +847,8 @@ class discretization_manager
   std::vector<aux_field_entry<precision>> aux_fields;
 
   //! Just testing for now
-  solvers::bicgstab<precision> poisson_iter;
+  solvers::cg<precision> poisson_iter;
```
mkstoyanov: I think these should be part of the poisson_solver class rather than the discretization manager.

src/asgard_discretization.hpp
```diff
@@ -770,13 +772,15 @@ class discretization_manager
...
-          p.update_level(terms.grid.current_level(0));
-        }
-      }, get_poisson());
+    std::visit([&](auto &p)
+    {
+      if constexpr (std::is_same_v<std::decay_t<decltype(p)>, poisson_1d<precision>>) {
+        p.update_level(terms.grid.current_level(0));
+      } else if constexpr (std::is_same_v<std::decay_t<decltype(p)>, poisson_md<precision>>) {
+        terms.moms.update_position_grid(terms.grid);
+        p.update_preconditioner(terms.moms.get_position_grid(), terms.conn, poisson_preconditioner);
```
mkstoyanov: Make the preconditioner part of the poisson_solver class.

src/asgard_gpu_algorithms_cuda.cpp
```diff
@@ -235,6 +235,73 @@ int num_non_finite(int64_t num, P const x[]) {
   return cpu_res;
 }
 
+template<typename P>
+__global__ void cg_calc_alpha_kernel(P const *rho, P const *p_dot_q, P *alpha) {
```
mkstoyanov: That's a lot of kernels and simple ones.

Can we not merge some of these, i.e., launch the `cg_update_x_r_kernel` and compute the simple division across all threads, use the result (probably in a register) and finally thread 0 on block 0 will write back to the output.

src/asgard_moment_manager.hpp
```diff
@@ -115,7 +115,7 @@ class moment_manager {
 
   //! solves the poisson equation and caches them as electric moments
   void solve_poisson(sparse_grid const &grid, connection_patterns const &conn, hierarchy_manipulator<P> const &hier,
-                     interpolation_manager<P> const &interp, kronmult::workspace<P> &work) const;
+                     interpolation_manager<P> const &interp, kronmult::workspace<P> &work, bool raw_on_cpu = false) const;
```
mkstoyanov: I try to avoid default parameters in internal API, especially something that will be called across function (i.e., not a local lambda).

If you use a boolean like that, then all call patterns must become:
```
bool const raw_on_cpu = false; // or true
solve_poisson(..., raw_on_cpu);
```

src/asgard_poisson.cpp
```diff
@@ -194,7 +195,7 @@ void poisson_md<P>::solve_potential_(gpu::vector<P> &density, sparse_grid const
   // We must force the mean of the RHS to 0 so the iterative solver doesn't blow up.
   if (bc == poisson_bc::periodic) {
     gpu::memcopy_dev2dev(1, density.data(), gpu_density0.data());
-    gpu::fill_zeros(1, density.data());
+    gpu::set_zero(density.data());
```
mkstoyanov: I think I already have a method that sets arbitrary number of bytes to zero, something like `compute->set_zero(density.data(), 1);`

It's a bit cleaner that you're setting the first entry to zero.

src/asgard_poisson.cpp
```diff
@@ -222,6 +223,77 @@ void poisson_md<P>::solve_potential_(gpu::vector<P> &density, sparse_grid const
...
+      jacobi.resize(num_entries);
+    else {
+      jacobi.resize(num_entries);
+      std::fill(jacobi.begin(), jacobi.end(), P{0});
+    }
+
+    for (term_entry<P> const &tentry : laplacian_terms)
+      kron_diag(tentry, position_grid, conn, block_size, jacobi);
+
+    jacobi[0] = P{0};
```
mkstoyanov: I think this is true only for periodic boundary conditions.

You can double-check with the Jacobi preconditioner applied to the elliptic example, that one uses the same operator but with Dirichlet and Neumann boundary conditions.

mkstoyanov (changes requested):

src/pde/two_stream.cpp
TTBurnett: I changed the initial condition here because the integral over velocity was 0.5 instead of 1. The canonical two-stream instability uses an average charge density of 1 not 0.5 and it seemed like this was an accident as 1 / sqrt(pi) is the correct constant for a Maxwellian distribution, but not for the x^2 exp(-x^2) distribution used here.

src/asgard_discretization.hpp
TTBurnett: The only reason this was not done is because the dependency chain of the headers needs to be reorganized to allow this. Currently solvers.hpp -> term_manager.hpp -> moment_manager.hpp -> poisson_solver.hpp so including solvers.hpp would create a cyclic dependency. I would think the solution is to move solvers to the end of this chain but the problem is the direct solver uses the term manager for factoring its matrices. Let me know if you have any thoughts on what this dependency chain should look like and I can fix it

src/asgard_gpu_algorithms_cuda.cpp
TTBurnett: Yes, we should merge them, this will get us better performance from fewer kernel launches as well. I will do this

PR #850, opened 2026-07-15
Command Line Parsing for Poisson Solver Options

TTBurnett:
Added command line parsing for Poisson solver options:
-poisson-precon <none/jacobi> (or -ppc)
-poisson-tol <double> (or -pt)
-poisson-iter <int> (or -pi)

Also included a warm start for the Poisson solver.

### DIFF
--- a/src/asgard_poisson.cpp
+++ b/src/asgard_poisson.cpp
@@ -109,23 +109,71 @@ void poisson_md<P>::solve(std::vector<P> &density, momentset<P> &moms,
 }
 
 template<typename P>
+void poisson_md<P>::remap_(std::vector<P> &x, indexset const& iset_old, indexset const &iset_new) const
+{
+  int64_t const num_old = iset_old.num_indexes();
+  int64_t const num_new = iset_new.num_indexes();
+  int const block_size = fm::ipow(pdof, num_dims);
+
+  // Initialize exactly to 0 to automatically handle newly refined points
+  std::vector<P> x_new(num_new * block_size, P{0});
+
+  int64_t iold = 0;
+  int64_t inew = 0;
+
+  enum class index_relation
+  {
+    asameb,
+    abeforeb,
+    bbeforea
+  };
+
+  auto compare_indexes = [&](int const a[], int const b[]) -> index_relation
+    {
+      for(int const d : iindexof(num_dims)) {
+        if (a[d] < b[d]) return index_relation::abeforeb;
+        if (a[d] > b[d]) return index_relation::bbeforea;
+      }
+      return index_relation::asameb;
+    };
+
+  while (inew < num_new && iold < num_old) {
+    index_relation relation = compare_indexes(iset_new[inew], iset_old[iold]);
+
+    if (relation == index_relation::asameb) {
+      // Point survived adaptation: Transfer data
+      std::copy_n(x.data() + iold * block_size, block_size, x_new.data() + inew * block_size);
+      inew++;
+      iold++;
+    } else if (relation == index_relation::abeforeb) {
+      // New point added (Refinement): Already 0, just advance
+      inew++;
+    } else {
+      // Old point removed (Coarsening): Discard data
+      iold++;
+    }
+  }
+  std::swap(x, x_new);
+}
+
+template<typename P>
 void poisson_md<P>::solve_potential_(std::vector<P> &density, sparse_grid const &position_grid,
                                      connection_patterns const &conn, kronmult::workspace<P> &work, poisson_bc const bc)
 {
-  size_t const n = density.size();
-
-  // Resize efield and set an initial guess of zeros
-  potential.resize(n);
-  std::fill(potential.begin(), potential.end(), P{0.0});
+  // Remap the previous potential to the new grid if needed to use as a warm start
+  if (generation != position_grid.generation()) {
+    remap_(potential, iset_, position_grid.iset());
+    iset_ = position_grid.iset();
+    generation = position_grid.generation();
+  }
 
   // Save the average density to restore it later
   P density0 = density[0];
 
   // For periodic boundaries, the solution is only unique up to a constant.
   // We must force the mean of the RHS to 0 so the iterative solver doesn't blow up.
-  if (bc == poisson_bc::periodic) {
+  if (bc == poisson_bc::periodic)
     density[0] = P{0};
-  }
 
   // Define the Matrix-Vector Product (The LHS)
   auto apply_lhs = [&](P alpha, P const x[], P beta, P y[]) -> void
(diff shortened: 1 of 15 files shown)


src/asgard_poisson.cpp
```diff
@@ -106,24 +108,72 @@ void poisson_md<P>::solve(std::vector<P> &density, momentset<P> &moms,
   }
 }
 
+template<typename P>
+void poisson_md<P>::remap_(std::vector<P> &x, indexset const& iset_old, indexset const &iset_new) const
```
mkstoyanov: The calling convention is "inputs - outputs", so the vector x should be the last variable.

src/asgard_poisson.cpp
```diff
@@ -106,24 +108,72 @@ void poisson_md<P>::solve(std::vector<P> &density, momentset<P> &moms,
...
+  auto compare_indexes = [&](int const a[], int const b[]) -> index_relation
+    {
+      for(int const d : iindexof(num_dims)) {
+        if (a[d] < b[d]) return index_relation::abeforeb;
+        if (a[d] > b[d]) return index_relation::bbeforea;
+      }
+      return index_relation::asameb;
+    };
+
+  while (inew < num_new && iold < num_old) {
```
mkstoyanov, line 140: When `inew == num_new` you need to end the iteration, there's nothing to do about the `iold`

src/asgard_poisson.cpp
```diff
@@ -106,24 +108,72 @@ void poisson_md<P>::solve(std::vector<P> &density, momentset<P> &moms,
...
 {
-  size_t const n = density.size();
-
-  // Resize efield and set an initial guess of zeros
-  potential.resize(n);
-  std::fill(potential.begin(), potential.end(), P{0.0});
+  // Remap the previous potential to the new grid if needed to use as a warm start
+  if (generation != position_grid.generation()) {
+    remap_(potential, iset_, position_grid.iset());
+    iset_ = position_grid.iset();
```
mkstoyanov, line 166: The `indexset` has the vector of ints but also more. We should check if we are copying something other than a vector and couple of ints as meta-data.

src/asgard_poisson.cpp
```diff
@@ -189,15 +239,64 @@ void poisson_md<P>::solve(gpu::vector<P> &density, sparse_grid const &position_g
...
+      }
+      return index_relation::asameb;
+    };
+
+  while (inew < num_new && iold < num_old) {
+    index_relation relation = compare_indexes(iset_new[inew], iset_old[iold]);
+
+    if (relation == index_relation::asameb) {
+      // Point survived adaptation: Transfer data
+      gpu::memcopy_dev2dev(block_size, x.data() + iold * block_size, x_new.data() + inew * block_size);
```
mkstoyanov, line 277: I wonder if it isn't better to make it into a single call to mem-copy.

src/asgard_program_options.cpp
```diff
@@ -161,6 +161,15 @@ Options          Short   Value      Description
                                     for GMRES this is the number of outer iterations.
 -isolve-inner    -isn    int        (GMRES only) The maximum number of inner GMRES iterations,
                                     this is ignored by BiCGSTAB.
+-poisson-precon  -ppc    string     accepts: none/jacobi
+                                    applies to domains with 2 or 3 spatial dimensions and an electric field moment
+                                    specifies the preconditioner for the Poisson solver
+                                    none - is not advisable as it takes too long
+                                    jacobi - preconditioner that applies basic rescaling
+-poisson-tol     -pt     double     Poisson solver tolerance,
```
mkstoyanov: `pt` is too generic, use `-ptol`

src/asgard_program_options.cpp
```diff
@@ -161,6 +161,15 @@ Options          Short   Value      Description
...
+                                    applies to domains with 2 or 3 spatial dimensions and an electric field moment
+-poisson-iter    -pi     int        Poisson solver maximum number of iterations,
```
mkstoyanov: Use `-piter` here.

src/asgard_program_options.cpp
```diff
@@ -198,6 +207,9 @@ void prog_opts::process_inputs(std::vector<std::string_view> const &argv, handle
       {"-isolve-iter", optentry::isol_iterations}, {"-isi", optentry::isol_iterations},
       {"-isolve-inner", optentry::isol_inner_iterations},
       {"-isn", optentry::isol_inner_iterations},
+      {"-poisson-precon", optentry::poisson_precond}, {"-ppc", optentry::poisson_precond},
```
mkstoyanov, line 210: This is one of the few places in ASGarD that we are doing unit testing, we want to make sure we are getting the correct behavior if the user makes a typo.

See the `program_options_tests.cpp` file and add checks, set, not set, get exception.

mkstoyanov (changes requested):

src/asgard_poisson.cpp
TTBurnett: If iold reaches the end then that means every value in x_new should be a newly refined value which can stay zero so you don't need to do any more iterations. This should be a possible state to reach if you have a large amount of refinement and many of the values are near the end of the vector, you shouldn't need to keep looping once all of the old values have been transferred, right? Let me know if I'm missing something

src/asgard_poisson.cpp
TTBurnett: Do you mean flagging all of the data that needs to be copied during the loop and then copying once at the end? Does mem-copy let you do this since there would be several separate blocks of data which need to be transferred at the end?

src/asgard_program_options.cpp
TTBurnett: I meant to do this earlier but forgot, thanks. There are some tests now

src/asgard_poisson.cpp
TTBurnett: Here is all of the data on the index set which should be copied:
```
int num_dimensions_;
int64_t num_indexes_;
std::vector<int> indexes_;
```

src/asgard_poisson.cpp
mkstoyanov: Just thinking of a way to improve this, probably good enouhg.

src/asgard_poisson.cpp
mkstoyanov: It's OK to copy that.

src/asgard_poisson.cpp
mkstoyanov: If you flag the data, you will have to make your own kernel that does the copy.

As you saw in your tests, data movement can be very costly and kill scaling from using multiple GPUs and such.

It's OK to merge now but add a comment `TODO: see if can be turned into one kernel`

mkstoyanov:
Just add that one comment to the GPU mem-copy and I'll merge later today.

src/asgard_poisson.cpp
TTBurnett: I think I'll actually go ahead and write the kernel

mkstoyanov: I prefer if you don't so we can merge it now. If you write the CUDA kernel, then I have to include and test it on ROCm and that's more work.

We can resolve this now and you can write the kernel is another PR that has just that change, so I can do just one thing on Monday.

TTBurnett: Okay, I removed the changes for that kernel and put them on a separate branch for later. The PR is ready to merge now.

mkstoyanov (approved):

PR #851, opened 2026-07-24
Fix for set_adapt_weight() with Poisson solver.

TTBurnett:
Adapt weight was crashing when used with an electric moment because the poisson solver was not initialized before set_initial_condition() was called. I fixed this so set_adapt_weight() works with the Poisson solver. I also added a single kernel for the remap_() function on GPU in asgard_poisson.cpp to reduce the number of kernel launches.

### DIFF
--- a/src/asgard_gpu_algorithms_cuda.cpp
+++ b/src/asgard_gpu_algorithms_cuda.cpp
@@ -173,6 +173,26 @@ __global__ void kernel_sum5(int64_t num, P const x[], P a1, P const x1[],
     i += num_threads * gridDim.x;
   }
 }
+template<typename P, int num_threads>
+__global__ void kernel_flagged_memcopy(int block_size, int64_t num_transfers,
+                                       int64_t const transfers[], P const src[], P dest[])
+{
+  int i = threadIdx.x + blockIdx.x * num_threads;
+  int64_t total_elements = block_size * num_transfers;
+  while (i < total_elements) {
+    int64_t transfer_idx = i / block_size;
+    int64_t element_offset = i % block_size;
+
+    // 3. Look up the old and new block indices from the transfer map
+    // transfers is packed as: [src_idx_0, dest_idx_0, src_idx_1, dest_idx_1, ...]
+    int64_t src_block_idx = transfers[transfer_idx * 2];
+    int64_t dest_block_idx = transfers[transfer_idx * 2 + 1];
+
+    // 4. Perform the exact element copy
+    dest[dest_block_idx * block_size + element_offset] = src[src_block_idx * block_size + element_offset];
+    i += num_threads * gridDim.x;
+  }
+}
 
 template<typename P>
 void sum2(gpu::vector<P> const &x, no_deduce<P> a1, gpu::vector<P> const &x1, gpu::vector<P> &y) {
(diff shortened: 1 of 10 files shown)


src/pde/two_stream_md.cpp
```diff
@@ -82,24 +102,34 @@ __global__ void interp_negative_kernel(int64_t num, P const* field, P const* mom
...
-  // the domain has one position and one velocity dimension: 1x1v
-  asgard::pde_domain<P> domain(asgard::position_dims{2}, asgard::velocity_dims{2},
-                               {{-2 * PI, 2 * PI}, {-2 * PI, 2 * PI},
-                                {-2 * PI, 2 * PI}, {-2 * PI, 2 * PI}});
+  options.title = std::to_string(pos_dims) + "X" + std::to_string(pos_dims) + "V Two Stream Instability";
+
+  asgard::pde_domain<P> domain;
+  if (pos_dims == 2) {
+    domain = asgard::pde_domain<P>(asgard::position_dims{2}, asgard::velocity_dims{2},
+                                 {{-2 * PI, 2 * PI}, {-2 * PI, 2 * PI},
```
mkstoyanov: If everything is using the same domain range, you can use:
```
std::vector<asgard::domain_range>(4, {-2 * PI, 2 * PI});
```
Same for the 3D case.

src/pde/two_stream_md.cpp
```diff
@@ -234,85 +350,296 @@ asgard::pde_scheme<P> make_two_stream(asgard::prog_opts options) {
...
+        std::vector<P> const &e_y = moments[melectric_y];
+  #pragma omp parallel for
+        for (size_t i = 0; i < vals.size(); i++)
+          vals[i] = field[i] * (e_x[i] * e_x[i] + e_y[i] * e_y[i]);
+      };
+  } else {
+    pde += asgard::term_md<P>{
+        asgard::term_md(asgard::term_interp<P>(md_positive_x, {melectric_x, })),
+        asgard::term_md<P>{
+          asgard::term_identity{},
```
mkstoyanov: Having these kind of long patterns can be good for clarity, but at some point they become just clutter and it becomes very difficult to read.

Think how you can do this programatically with for loops, as opposed to manually unrolling such long statement.

mkstoyanov (changes requested):
Severe jetlag here, saw these will have other comments later.

src/pde/two_stream_md.cpp
TTBurnett: I cleaned this up so it is much more concise

src/asgard_gpu_algorithms_cuda.cpp
```diff
@@ -173,6 +173,26 @@ __global__ void kernel_sum5(int64_t num, P const x[], P a1, P const x1[],
...
   }
 }
+template<typename P, int num_threads>
+__global__ void kernel_flagged_memcopy(int block_size, int64_t num_transfers,
+                                       int64_t const transfers[], P const src[], P dest[])
+{
+  int i = threadIdx.x + blockIdx.x * num_threads;
+  int64_t total_elements = block_size * num_transfers;
+  while (i < total_elements) {
+    int64_t transfer_idx = i / block_size;
```
mkstoyanov, line 183: Next time we talk, remind me to talk about launching a kernel on a 2D thread grid and doing cycles.

Nothing to change right now, this is fine.

mkstoyanov (approved):

src/asgard_gpu_algorithms_cuda.cpp
TTBurnett: Will do

PR #853, opened 2026-08-06
Adapt weight for 1D Poisson solver

TTBurnett:
Adapt weight was updated to work for 1D moments such as the electric field in two_stream.cpp.

### DIFF
--- a/src/asgard_moment_manager.cpp
+++ b/src/asgard_moment_manager.cpp
@@ -641,7 +641,7 @@ void moment_manager<P>::complete_level(hierarchy_manipulator<P> const &hier,
 template<typename P>
 void moment_manager<P>::solve_poisson(sparse_grid const &grid, connection_patterns const &conn,
                                       hierarchy_manipulator<P> const &hier, interpolation_manager<P> const &interp,
-                                      kronmult::workspace<P> &work, bool result_to_cpu) const
+                                      kronmult::workspace<P> &work) const
 {
   std::visit([&](auto &p) {
       if constexpr (std::is_same_v<std::decay_t<decltype(p)>, poisson_1d<P>>) {
@@ -649,35 +649,19 @@ void moment_manager<P>::solve_poisson(sparse_grid const &grid, connection_patter
       } else if constexpr (std::is_same_v<std::decay_t<decltype(p)>, poisson_md<P>>) {
         #ifdef ASGARD_USE_GPU
         compute->set_device(gpu::device{0});
+
+        // transfer density to GPU
         int const num_entries = pos_grid.num_dof();
         std::array<gpu::vector<P>, max_num_gpus> &work1 = interp.gpu_it1;
         assert(work1[0].size() >= num_entries);
         work1[0].copy_from_host(num_entries, raw_vals.get(p.moment0()).data());
-        for (moment_id const &mid : p.moments_electric()) {
-          if (mid == moment_id::unset()) continue;
-          gpu::vector<P> &res = gpu_interps[0][mid];
-          res.resize(full_block * grid.num_indexes());
-        }
+
+        // solve
+        constexpr bool result_to_cpu = true;
         solve_poisson_gpu(grid, conn, hier, interp, work, result_to_cpu);
-        auto find_mom = [this](moment_id mid) -> const mom_on_gpu &
-        {
-          for (auto const &mom : this->gpu_moments[0]) {
-            if (mom.mid == mid)
-              return mom;
-          }
-          throw std::runtime_error("could not find the moment");
-        };
-        for (moment_id const &mid : p.moments_electric()) {
-          if (mid == moment_id::unset()) continue;
-          mom_on_gpu const &mom = find_mom(mid);
-          if (mom.interp_on_cpu()) continue; // solve_poisson_gpu will already copy if interp_on_cpu() is true
-          gpu::vector<P> &res = gpu_interps[0][mid];
-          res.copy_to_host(interps.get(mid));
-        }
         #else
         std::ignore = interp;
         std::ignore = grid;
-        std::ignore = result_to_cpu;
         update_position_grid_dsort();
         p.solve_periodic(raw_vals.get(p.moment0()), raw_vals, pos_grid, conn, work);
         #endif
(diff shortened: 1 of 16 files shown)


src/asgard_transformations.cpp
```diff
@@ -9,6 +9,56 @@
...
+  size_t const old_pdof = old_degree + 1;
+  size_t const new_pdof = new_degree + 1;
+  assert(old_pdof > 0);
+  assert(new_pdof > old_pdof);
+  assert(not vec.empty());
+  assert(vec.size() % old_pdof == 0);
+  size_t const nelems = vec.size() / old_pdof;
+  size_t const new_size = nelems * new_pdof;
+  vec.resize(new_size, P{0});
+  for (size_t i = nelems - 1; i > 0; i--)
```
mkstoyanov: You cannot cycle in reverse with a for loop and unsigned counter. You need to use `int64_t`, otherwise the loop will never terminate or (as in this case) you will be missing the i = 0 index.

src/asgard_transformations.hpp
```diff
@@ -5,6 +5,50 @@
...
+ * \p src is expected to come from a degree \p src_degree representation
+ * and \p dest will be degree \p dest_degree where the higher degree coefficients
+ * are set to zero.
+ *
+ * \p dest_degree must be greater than \p src_degree
+ * 
+ * \endinternal
+ */
+template<typename P>
+void pad_degree(int src_degree, int dest_degree, std::vector<P> const &src, std::vector<P> &dest);
```
mkstoyanov: I don't think you need all variants of `pad_degree`. I think you need only one.

src/asgard_transformations.cpp
```diff
@@ -9,6 +9,56 @@
...
+  size_t const src_pdof = src_degree + 1;
+  size_t const dest_pdof = dest_degree + 1;
+  assert(src_pdof > 0);
+  assert(dest_pdof > src_pdof);
+  assert(not src.empty());
+  assert(src.size() % src_pdof == 0);
+  size_t const nelems = src.size() / src_pdof;
+  size_t const dest_size = nelems * dest_pdof;
+  dest.resize(dest_size);
+  std::fill(dest.begin(), dest.end(), P{0});
```
mkstoyanov: For a large vector, you want to merge this with the for-loop below. Otherwise you will do a full read of the data, then another full read below and won't be able to use caching.

src/asgard_transformations.cpp
TTBurnett: The i = 0 case is not needed since it would just copy data to where it already is

src/asgard_transformations.hpp
TTBurnett: That is true for now, I figured it might be nice to make more variants in case another was needed in the future but I can remove them.

src/asgard_transformations.cpp
mkstoyanov: The non-zero data is there, but the padding zeros also have to be put in the padded block.

TTBurnett: Oh right, good catch.  I deleted this variation of the pad_degree function anyways

src/pde/two_stream.cpp
```diff
@@ -152,6 +152,19 @@ asgard::pde_scheme<P> make_two_stream(asgard::prog_opts options) {
 
   pde.add_initial(asgard::separable_func<P>({ic_x, ic_v}));
 
+  asgard::moment_id melectric = pde.register_electric_moment(asgard::dimension_id(0), 1);
```
mkstoyanov: What's the second number here?

src/asgard_moment_manager.cpp
```diff
@@ -641,99 +643,96 @@ void moment_manager<P>::complete_level(hierarchy_manipulator<P> const &hier,
...
-    }, poisson_solver);
+    if constexpr (std::is_same_v<std::decay_t<decltype(p)>, poisson_1d<P>>) {
+      p.solve_periodic(get_cached_level(p.moment0(), hier), full_level.get(p.moment_electric()));
+    } else if constexpr (std::is_same_v<std::decay_t<decltype(p)>, poisson_md<P>>) {
+      #ifdef ASGARD_USE_GPU
+      compute->set_device(gpu::device{0});
+
+      // transfer density to GPU
+      int const num_entries = pos_grid.num_dof();
+      std::array<gpu::vector<P>, max_num_gpus> &work1 = interp.gpu_it1;
```
mkstoyanov: Just get the reference to the `gpu::vector`, otherwise you're dereferencing interp to get the array address and then jumping again.

The compiler will optimize this but if you're only ever using the 0-th vector, then just take a reference to the 0-th vector.

src/asgard_moment_manager.cpp
```diff
@@ -641,99 +643,96 @@ void moment_manager<P>::complete_level(hierarchy_manipulator<P> const &hier,
...
+      assert(work1[0].size() >= num_entries);
+      work1[0].copy_from_host(num_entries, raw_vals.get(p.moment0()).data());
```
mkstoyanov: raw-vals returns a reference to vector or vector2D?

The get method is there so it makes the getting process cleaner when dealing with raw-pointers and there is something about the `[]` operator. This should probably use the `[]` operator and then assignment can be done without the assertion above.

src/asgard_transformations.cpp
```diff
@@ -9,6 +9,23 @@
 namespace asgard
 {
 
+template<typename P>
+std::vector<P> pad_degree(int src_degree, int dest_degree, std::vector<P> const &src)
+{
+  size_t const src_pdof = src_degree + 1;
+  size_t const dest_pdof = dest_degree + 1;
+  assert(src_pdof > 0);
+  assert(dest_pdof > src_pdof);
```
mkstoyanov, line 18: What happens if I run the Poisson solver on a problem with degree 0?

mkstoyanov (changes requested):

src/asgard_moment_manager.hpp
```diff
@@ -109,19 +117,20 @@ class moment_manager {
...
-  //! solves the poisson equation and caches them as electric moments
-  void solve_poisson(sparse_grid const &grid, connection_patterns const &conn, hierarchy_manipulator<P> const &hier,
-                     interpolation_manager<P> const &interp, kronmult::workspace<P> &work) const {
-    constexpr bool raw_on_cpu = false;
-    solve_poisson(grid, conn, hier, interp, work, raw_on_cpu);
+  //! return the cached raw moment defined on the position grid with moment id mid, overload for 1d electric moments
+  std::vector<P> const &get_cached_raw(moment_id mid, hierarchy_manipulator<P> const &hier) const {
+    if (num_pos_ == 1 and needs_poisson(mid))
+      cache_raw_from_level(mid, hier);
+    return raw_vals.get(mid);
```
mkstoyanov: Use `[]`

mkstoyanov (approved):

PR #855, opened 2026-08-13
Separable BCs

TTBurnett:
Separable boundary conditions are now supported on both CPU and GPU.

### DIFF
--- a/src/asgard_compute.hpp
+++ b/src/asgard_compute.hpp
@@ -234,7 +234,7 @@ public:
   template<typename P>
   void getrs(int M, gpu::vector<P> const &A, gpu::vector<gpu::direct_int> const &ipiv,
              std::vector<P> &b) const {
-    gpu::vector<P> gpu_b(b);
+    gpu::vector<P> gpu_b{b};
     getrs(M, A, ipiv, gpu_b);
     gpu_b.copy_to_host(b);
   }
--- a/src/asgard_compute.hpp
+++ b/src/asgard_compute.hpp
@@ -36,7 +36,7 @@ public:
       gpu::memfree(data_);
   }
   //! \brief Construct a vector with given size.
-  vector(int64_t size)
+  explicit vector(int64_t size)
   {
     this->resize(size);
   }
@@ -66,7 +66,7 @@ public:
     return *this;
   }
   //! \brief Constructor that copies from an existing std::vector
-  vector(std::vector<T> const &other) : vector()
+  explicit vector(std::vector<T> const &other) : vector()
   {
     *this = other;
   }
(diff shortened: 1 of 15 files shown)


src/asgard_compute.hpp
```diff
@@ -234,7 +234,7 @@ class __signleton_compute_resources {
   template<typename P>
   void getrs(int M, gpu::vector<P> const &A, gpu::vector<gpu::direct_int> const &ipiv,
              std::vector<P> &b) const {
-    gpu::vector<P> gpu_b = b;
+    gpu::vector<P> gpu_b(b);
```
mkstoyanov: There is a nasty C++ problem, look at this code:
```
gpu::vector<P> gpu_b();
```
Is this an empty GPU vector or the declaration of a function that takes no parameters and returns a `gpu::vector`. The standard says that his is a function, which is unexpected for many people.

A better practice is to use `{}` for constructors, e.g.,
```
gpu::vector<P> gpu_b{};
```
which is an empty vector and it comes with a few more checks.

So, I prefer to write:
```
gpu::vector<P> gpu_b{b};
```
I just have to make sure this isn't ambiguous with respect to an initializer list.

src/asgard_gpu_tensors_cuda.cpp
```diff
@@ -727,6 +727,219 @@ void remap_state(int block_size, gpu::vector<int> const &map, gpu::vector<P> &st
...
+  while (i < num_indexes)
+  {
+    int const *idx = indexes + i * num_dims;
+    int const isub = map[i];
+
+    P const *block1d = con1d + pdof * idx[ib_dim];
+    P const *subblock = bnd + subgrid_block_size * isub;
+
+    // Each thread handles four coefficients.
+    for (int c = 0; c < pdof; c++)
```
mkstoyanov: Cycle 4 is different from pdof 4, they may happen to coincide here but in general the number are different. Depending on the dimension, you can have a wide range of possible pdof and cycles.

Set this to 4, or `int constexpr num_cycles = 4;`

src/asgard_gpu_tensors_rocm.cpp
```diff
@@ -730,6 +730,219 @@ void remap_state(int block_size, gpu::vector<int> const &map, gpu::vector<P> &st
...
+  while (i < num_indexes)
+  {
+    int const *idx = indexes + i * num_dims;
+    int const isub = map[i];
+
+    P const *block1d = con1d + pdof * idx[ib_dim];
+    P const *subblock = bnd + subgrid_block_size * isub;
+
+    // Each thread handles four coefficients.
+    for (int c = 0; c < pdof; c++)
```
mkstoyanov: Same as the CUDA case.

mkstoyanov (changes requested):

src/asgard_compute.hpp
TTBurnett: Ah yes that makes sense, I'll switch that to {}

mkstoyanov (approved):

